PR #54, opened 2017-08-22
feat(graphile-build-pg): export case-altering functions

0ff:
This allows for custom inflections to use the same case-altering functions (`upperFirst, camelCase, constantCase`) as the provided inflections.


benjie:
For this to be useful we ought to export these things from `src/index.js`. I don't want to do `export * from './utils'` though - some of these things should be private, so if you want to add that only export a whitelist of properties from utils (specifically the 3 case functions - i.e. `export { upperFirst, camelCase, constantCase } from "./utils";`)

PR #59, opened 2017-08-24
fix(mutations): nullable payloads, rollback individual mutations on error

### LINKED ISSUE #75: If an SQL error occurs in a mutation, previous data may still be rendered
For example if you perform createUser1, mutationThatProducesError, createUser2; the response may contain `{ createUser1: { id: 1, name: ... }, mutationThatProducesError: null, createUser2: null}`; however the details of user 1 will have been rolled back by the mutation and thus should not be displayed to the user.

I suspect the solution is to use transaction checkpoints, because GraphQL will resolve each mutation field to completion before executing the next mutation field we don't need to worry about errors thrown in later mutations causing issues for resolve values of earlier mutation fields.

benjie:
Wraps each mutation in a transaction, automatically rolling back so that a) results of previous mutations are not invalidated (see #75) and b) future mutations are not prevented.

Fixes postgraphql/postgraphql#563
Fixes #75

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
@@ -195,13 +195,24 @@ export default (function PgMutationCreatePlugin(
                       ) values(${sql.join(sqlValues, ", ")})`
                     : sql.fragment`default values`} returning *`;
 
-                  const { rows: [row] } = await viaTemporaryTable(
-                    pgClient,
-                    sql.identifier(table.namespace.name, table.name),
-                    mutationQuery,
-                    insertedRowAlias,
-                    query
-                  );
+                  let row;
+                  try {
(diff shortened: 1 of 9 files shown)


benjie:
From https://github.com/postgraphql/postgraphql/issues/563#issuecomment-325673345:

The error only happens with missing grants:

```sql
CREATE OR REPLACE FUNCTION public.test() RETURNS boolean LANGUAGE 'sql' AS $func$ SELECT true $func$;
REVOKE EXECUTE ON FUNCTION public.test() FROM PUBLIC;
mutation regression {
  test(input: {}) {
    clientMutationId
    boolean
  }
}
{
  "data": null, // <---
  "errors": [
    {
      "message": "permission denied for function test",
      "locations": [
        {
          "line": 29,
          "column": 3
        }
      ],
      "path": [
        "test"
      ]
    }
  ]
}
```

PR #64, opened 2017-09-11
feat(inflection): export inflection engine

benjie:


EyMaddis:
Can you document the changes?

I tried
```js 
import { inflections } from 'graphile-build-pg';
const { newInflector } = inflections

const customInflections = newInflector({
   singleRelationByKeys() { console.log('Called me') }
})
....

postgraphql(pgConfig, schemas, {
        ....
        graphqlBuildOptions: {
            pgInflection: customInflections,
        }
    })
```

But the code is never called. 


I previously used the following:

```js
import { inflections } from 'graphile-build-pg';

const defaultInflections = inflections.postGraphQLInflection;
const customInflections = Object.assign({}, defaultInflections, {
  singleRelationByKeys(detailedKeys, table, schema) {
     // doStuff
  }
});
```

leading to 

```
TypeError: inflection.rangeType is not a function
    at enforceGqlTypeByPgType (node_modules/graphile-build-pg/src/plugins/PgTypesPlugin.js:388:46)
    at Array.forEach (native)
    at node_modules/graphile-build-pg/src/plugins/PgTypesPlugin.js:547:37
    at SchemaBuilder.applyHooks (node_modules/graphile-build/src/SchemaBuilder.js:288:20)
    at SchemaBuilder.createBuild (node_modules/graphile-build/src/SchemaBuilder.js:324:31)
    at SchemaBuilder.buildSchema (node_modules/graphile-build/src/SchemaBuilder.js:339:26)
    at SchemaBuilder._callee$ (node_modules/graphile-build/src/SchemaBuilder.js:371:32)
    at tryCatch (node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js:65:40)
    at Generator.invoke [as _invoke] (node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js:299:22)
    at Generator.prototype.(anonymous function) [as next] (node_modules/babel-runtime/node_modules/regenerator-runtime/runtime.js:117:21)
```

EyMaddis:
Ok, I got the code working.
There was another problem and `inflections.postGraphQLInflection` was `undefined`.
I now get them via `import { postGraphQLInflection } from 'postgraphile-core';`. Is there a better way, @benjie ?

benjie:
That's currently the best way; I felt the PostGraphQL inflections belonged in `postgraphile-core` rather than `graphile-build-pg` - sorry for the breakage!

https://github.com/postgraphql/postgraphql/pull/506#issuecomment-328831456

You can just use `inflector` now rather than hacking via `graphqlBuildOptions`.

EyMaddis:
I wonder why you did not use classes for this purpose.
Right now, we cannot extend from the PostGraphQl inflections without `Object.assign`, which could also be `class MyInflector extends PostGraphQLInflector`.

benjie:
Because classes are unnecessarily heavy weight; you can just us object splats (supported natively in Node v8.4+):

```
const newInflector = {
   ...postGraphQLInflector,
   myOverride() { return 'foo'; }
};
```

disarticulate:
my thought would be to define a function in the database schema that would return an alias to be used instead of the default. My personal goal is to not have to do anything to graphile backend beyond setup the various flags. I'd prefer to, as it seems like this project wants me to, put everything into the database schema and focus on the frontend->database relationship as opposed to the frontend->backend->database or the other contortions people go through to make ORMs.

benjie:
I'm planning to use the database comments for this instead, using the smart comments feature introduced in #128, just haven't got around to it yet.

disarticulate:
@benjie 

my naive comment would be that a schema function would allow for a more powerful database tooling than a DSL in comments. I can see the point of a DSL which allows documentation & definition, but I think it detracts from development tooling.

I've got a postgres dev environment experiment up running with this package here:
https://github.com/disarticulate/osm2pgsqlauto

Everything fit really well together, including graphile and now I'm looking at the client end of it. Within the above, I side step some of the query/interspection stuff with simple functions like
```
CREATE FUNCTION monitoring.list_all_tables (schema text)
  RETURNS json
AS $$
  from json import loads, dumps
  rv = plpy.execute("""
    SELECT * 
    FROM information_schema.tables 
    WHERE table_schema = '{}'
  """.format(schema))
  feature = { "tables": [i for i in rv] }
  return dumps(feature)
$$ LANGUAGE plpython3u stable;
```
Where I'm just dumping json into the stream, which I'm vaguely aware is not supposed to be the miracle of GraphQL. But functions like the above could be used any json (or a DSL)  from a database to define the graphql custom queries.

benjie:
Could you expand on your request a bit more - both in terms of motivation (what's the aim) and execution (how do you see it being acheived)? At the moment your example seems to do indirect introspection to return a list of tables; I can't see how this is helpful over us introspecting the database directly. It doesn't seem to handle renaming the tables that are returned, which I think is the need you were attempting to address?

disarticulate:
I'm concerned that the API isn't recognizably connected to the SQL backend, which is why the above ugly snippet exists.

in SQL I would have 
```SELECT * from test_table```

But graphql would present the query as
```{ allTestTables ... }```

Which means my client has to be hard coded to understand that if I want all the test tables, I can't just look at the SQL Schema and say, ok, this is the endpoint.

What I want to do is point graphile at a arbitrary Schema, and have:
```SELECT * from test_table``` be equal to ```{ select_test_table... }```

This allows my limited scope to go back and forth, and to parse them basically the same. But I'm aware that graphile is looking at a different set of characteristics & standards, which means I'd like to redefine 

```allTestTables -> select_test_table``` _without having to modify graphile or maintain a DSL that's seperate from either_

As I said above, this is naive, as I have not fully reviewed the code, and it doesn't help that flow script throws my spider senses. What I'd like to do is have graphile run a startup query like:
```
findEndPoint(allTestTables) -> " inflector_definition(default_endpoint)"
```
the equivalent in postgres as a function:
```
CREATE FUNCTION inflector_definition (default_endpoint text)
  RETURNS setof inflectors
AS $$
  SELECT inflector_map FROM api_endpoints
  WHERE endpoint == $default_endpoint
$$ LANGUAGE sql stable;
```
The problem I see with managing a definition file in comments is that it abruptly removes the full introspection powers. As I was looking from the docs, I wanted a generic search function, and this is what I build it as, since conditions are too specific for my liking. I'm still learning what graphile is capable of but I was suprised how well this seemed to work:
```
DROP FUNCTION if exists monitoring.create_search_functions;
CREATE FUNCTION monitoring.create_search_functions (schema text)
  RETURNS boolean
AS $$
  rv = plpy.execute("""
    SELECT * 
    FROM information_schema.tables 
    WHERE table_schema = '{}'
  """.format(schema))
  for table in rv:
    table_name = table["table_name"]
    plpy.info(table["table_name"])
    script_tokens = "$"*2
    query = """
      DROP FUNCTION if exists {schema}.search_{table};
      CREATE FUNCTION {schema}.search_{table} (
          col text[], 
          search text[])
        RETURNS setof {schema}.{table}
      AS {script_tokens}
        query = [
          'SELECT * FROM "',
          '{schema}',
          '"."',
          '{table}',
          '" ',
          'WHERE '
        ]
        for c, s in zip(col,search):
          query.append('"')
          query.append(c)
          query.append('"::text')
          query.append(' LIKE ')
          query.append("'%")
          query.append(s)
          query.append("%'")
          query.append(' AND ')
        query = query[:-1]
        rv = plpy.execute(''.join(query))
        plpy.info(''.join(query))
        return rv
      {script_tokens} LANGUAGE plpython3u stable;
    """.format(
      script_tokens=script_tokens,
      schema=schema,
      table=table_name
    )
    plpy.execute(query)
    plpy.info(query)
  return True
$$ LANGUAGE plpython3u volatile;

select monitoring.create_search_functions('monitoring');
```
It's ugly, but it dynamically pulls in a search function. And I'd like to have another function that informs graphile that I'd like to use this as the allTestTables function with the added search functionality.

This lets you guys develop whatever core functionality you need, then lets me develop any missing pieces that I need within the database. 

So I think what I'd like to see is graphile query a pre-set database function that remaps inflectors inside the schema.

benjie:
Yeah, definitely sounds like you want something quite custom - you’re probably better off using graphile-build-pg directly and writing a few of your own plugins.

An alternative is to just write your own inflector (you don’t need to modify postgraphile to do so, just pass it into the library function as one of the configuration parameters) which shouldn’t require much in the way of maintenance. If you think other people would want this too you could release the inflector to npm so that others could use it too.

disarticulate:
No worries. Once I resolved the above trick, graphile works fine with what I need. I just don't think comments are where one should define functional aspects of code. Much better to just create conventional functions in the schema which return the data requested.

PR #71, opened 2017-09-14
Add breaking tests for creating objects with explicit "null" values

EyMaddis:
It is currently impossible to create objects and pass a value as `null`, so I added a breaking test, see https://github.com/postgraphql/postgraphql/issues/584

``` graphql
createDefaultValue(input: {
    defaultValue: {
      id: 2000
      nullValue: null
    }
  }) {
    defaultValue {
      id
      nullValue // => is "defaultValue!" instead of "null"
    }
  }
```

### DIFF
--- a/packages/postgraphile-core/__tests__/fixtures/mutations/mutation-create.graphql
+++ b/packages/postgraphile-core/__tests__/fixtures/mutations/mutation-create.graphql
@@ -201,6 +201,17 @@ mutation {
       issue27UserExists: exists(email: "graphile-build.issue.27@example.com")
     }
   }
+  i: createDefaultValue(input: {
+    defaultValue: {
+      id: 2000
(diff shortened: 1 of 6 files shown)


chadfurman:
Would it make sense to have a test case for when it is nullable?

EyMaddis:
@chadfurman can you be more specific? I am not sure what you mean.

chadfurman:
@EyMaddis I misunderstood and thought this PRdid something differently, and actually quite opposite, than it is.  The case you're testing for is "null" values on a nullable field (which is what my comment was referencing, thus the confusion):

https://github.com/graphile/graphile-build/pull/71/files#diff-c8bb689a58197b1592c6e704ad935d16R225

benjie:
Closing in favour of #74

PR #73, opened 2017-09-14
add an optional pgColumnFilter for the columns plugin

### LINKED ISSUE #65: override keys with PgForwardRelationPlugin
Hi,

I’m trying to override a key with it’s single relation, for that I’ve, overridden the inflector like this:

```js
defaultInflection.singleRelationByKeys = ([{column}]) => column;
```

but I have this error:

> Error: Overwriting key '…' is not allowed!

I’d like to override to avoid having, for example, both `owner` and `accountByOwner` and instead just have `owner` resolve to the `Account` type. 
So my question is, is `extend` complaining because it can’t override a key? or because it thinks it’s probably an error?

If the error is just to prevent accidental overriding, I could duplicate and adapt `PgForwardRelationPlugin` but it’ll make it harder to maintain it in the future

maybe I can avoid overriding those fields by not adding them or removing them, but it will probably mean at duplicating some logic form `PgForwardRelationPlugin` into a new plugin (if I can add a plugin before `PgForwardRelationPlugin` that would remove those fields) or duplicating the plugin that add the fields and make it more complex and hard to maintain too

do you have suggestion on how to approach that ? would you be open to a PR that adds an option to `PgForwardRelationPlugin` so that it overrides the keys instead of calling the inflector ?

mathroc:
it works very nicely for the use case I described in #65 :)

closes #65


benjie:
Looks like you need to run "eslint --fix" (I do this automatically on save in my editor)

packages/graphile-build-pg/src/plugins/PgColumnsPlugin.js
```diff
@@ -151,6 +150,7 @@ export default (function PgColumnsPlugin(
         fields,
         introspectionResultsByKind.attribute
           .filter(attr => attr.classId === table.id)
+          .filter(attr => pgColumnFilter(attr, build, Object.assign({}, context, {type: "GraphQLObjectType"})))
```
benjie: I can't expand on mobile but is this one meant to be GraphQLInputObjectType?

mathroc: correct, it’s fixed

mathroc:
I’ve fixed most of the linting issue but I don’t know how to fix the flow lint errors :/ do you have hints on this ?

> Awesome! Fix linting, check the type, and add a test or two (probably in the form of an acceptance test on postgraphile-core) and we should be good to go.

ah yes I didn’t find tests in this package and didn’t look on the other packages. do you have somewhere in mind for the new tests ? as in, inside an existing test file (I didn’t find one related to building schemas) or just another file next to packages/postgraphile-core/__tests__/integration/queries.test.js ?

benjie:
> do you have somewhere in mind for the new tests ?

You're probably going to want a schema snapshot test to ensure that the fields are indeed missing:

https://github.com/graphile/graphile-build/blob/master/packages/postgraphile-core/__tests__/integration/schema.test.js

It'd also be worth adding some query tests to ensure that it doesn't break anything; for that you could add an alternative createPostGraphQLSchema instance that omits some fields in here, assuming you can omit fields that won't break the queries:

https://github.com/graphile/graphile-build/blob/23ecc372c7c81b2d322d7a31fac5d1090c302bab/packages/postgraphile-core/__tests__/integration/queries.test.js#L31-L41

Failing that make a separate query test file like I do for `queries-jwt.test.js`

> I don’t know how to fix the flow lint errors

I can't expand on this right now, sorry!

mathroc:
ok, I’ve fixed the travis build, can you take a look at the diff again ? also, I had expose the option to `createPostGraphQLSchema`, is that what you intented ? I don’t use postgraphql-core so I don’t if it’s relevant

benjie:
From a glance over this all looks great 👍 I'm probably not going to be able to merge it until Tuesday. If it's not merged by Friday please give me a nudge ☺️

mathroc:
ping ☺

packages/graphile-build-pg/src/plugins/PgColumnsPlugin.js
```diff
@@ -6,41 +6,46 @@ import type { Plugin } from "graphile-build";
...
-          .filter(attr => attr.classId === table.id)
-          .reduce((memo, attr) => {
-            /*
+      introspectionResultsByKind.attribute
+        .filter(attr => attr.classId === table.id)
+        .filter(attr =>
+          pgColumnFilter(
+            attr,
+            build,
+            Object.assign({}, context, { type: "GraphQLObjectType" })
```
benjie: I want to make this `Object.assign` unnecessary by moving it to core.

mathroc:
I’ll let you do the first point ( I probably won’t have time to work much on this until next month )
but if you didn’t take care of he second point then, I can work on it. (this means other value will be available on context for columns, orderBys, etc, right ?)

I also have use cases to filter orderBys (eg: I want to remove order by on uuid v4)

and as we’re talking about filtering, I also would like to remove some relations added by the PgBackwardRelationPlugin

benjie:
Hopefully that first part is handled by #79 👍  (Still need to remove Object.assign from this PR though - it shouldn't be necessary any more)

> (this means other type value will be available on context for columns, orderBys, etc, right ?)

Type will be, but it'll just be the object type (GraphQLInputObjectType for orderBy, conditions, etc). But because we pass the whole of context through hopefully there will be other useful things to filter it by.

> and as we’re talking about filtering, I also would like to remove some relations added by the PgBackwardRelationPlugin

I think this would probably be a different function (separate PR).

mathroc:
Hi,
ck is 
I’ve made some progress here, tell me what you think. I’m not sure about the callback function name anymore

packages/graphile-build-pg/src/plugins/PgConnectionArgCondition.js
```diff
@@ -1,67 +1,66 @@
 // @flow
 import type { Plugin } from "graphile-build";
+
+const defaultPgColumnFilter = (_table, _build, _context) => true;
```
benjie: I think that's meant to be `_attr` rather than `_table`?

benjie:
Please don't rebase / squash / `--amend` again now - just add commits. We'll be doing a squash and merge when it goes out but I'd rather not have to read through the entire commit again when reviewing!

packages/graphile-build-pg/src/plugins/PgConnectionArgCondition.js
mathroc: I wasn’t sure about this one, the argument, at call time, is `table` in this plugin (and in `PgConnectionArgOrderBy.js` too) should I call them `_attr` in `defaultPgColumnFilter` regardless ?

packages/graphile-build-pg/src/plugins/PgConnectionArgCondition.js
```diff
@@ -1,67 +1,66 @@
...
+    const {
+      newWithHooks,
+      pgIntrospectionResultsByKind: introspectionResultsByKind,
+      pgGqlInputTypeByTypeId: gqlTypeByTypeId,
+      graphql: { GraphQLInputObjectType, GraphQLString },
+    } = build;
+    introspectionResultsByKind.class
+      .filter(table => table.isSelectable)
+      .filter(table => !!table.namespace)
+      .filter(table => pgColumnFilter(table, build, context))
```
benjie: This is filtering in the wrong place, we should be filtering the condition fields in the section below 👇

packages/graphile-build-pg/src/plugins/PgConnectionArgCondition.js
```diff
@@ -1,67 +1,66 @@
...
+        newWithHooks(
+          GraphQLInputObjectType,
+          {
+            description: `A condition to be used against \`${tableTypeName}\` object types. All fields are tested for equality and combined with a logical ‘and.’`,
+            name: inflection.conditionType(
+              inflection.tableType(table.name, table.namespace.name)
+            ),
+            fields: ({ fieldWithHooks }) =>
+              introspectionResultsByKind.attribute
+                .filter(attr => attr.classId === table.id)
```
benjie: This is where we want the filter from above 👆

packages/graphile-build-pg/src/plugins/PgConnectionArgCondition.js
benjie: Ah! I clearly wasn't reading carefully enough - that explains your concerns calling it `pgColumnFilter`!

We only want to filter the columns in this PR; so every time we call `pgColumnFilter` we should be passing an `attr`. The orderBy columns come from https://github.com/graphile/graphile-build/blob/master/packages/graphile-build-pg/src/plugins/PgOrderAllColumnsPlugin.js

mathroc: ah yes, I didn’t take enough time figuring out how this plugin worked, maybe you can double check the 2 calls I made to pgColumnFilter in PgConnectionArgOrderBy as I’m not confident I did the right thing™ there either

benjie:
```
    -  # Checks for equality with the object’s `headline` field.
    -  headline: String
    -
```

^ The snapshots are failing because `headline` was correctly removed! 🎉

packages/graphile-build-pg/src/plugins/PgConnectionArgOrderBy.js
```diff
@@ -2,68 +2,65 @@
...
+  builder.hook("init", (_, build, context) => {
+    const {
+      newWithHooks,
+      pgIntrospectionResultsByKind: introspectionResultsByKind,
+      graphql: { GraphQLEnumType },
+    } = build;
+    introspectionResultsByKind.class
+      .filter(table => table.isSelectable)
+      .filter(table => !!table.namespace)
+      .filter(table => pgColumnFilter(table, build, context))
```
benjie: This one is filtering a table too, so is misplaced? 👆

packages/graphile-build-pg/src/plugins/PgConnectionArgOrderBy.js
```diff
@@ -2,68 +2,65 @@
...
-    ) => {
+      } = context;
       if (
         !isPgConnectionField ||
         !table ||
         table.kind !== "class" ||
         !table.namespace ||
-        !table.isSelectable
+        !table.isSelectable ||
+        !pgColumnFilter(table, build, context)
```
benjie: Also here

benjie:
Sterling work! Keep it up 👍

mathroc:
I’ve fixed the tests, removed the filter call from `PgConnectionArgOrderBy.js` and add a few `// todo` where I think the column filter could be called, please tell me if they’re appropriate places and if I missed some. maybe in `PgMutationUpdateDeletePlugin.js` & `PgMutationCreatePlugin.js` too ? if you know where it’ll save me some time

benjie:
I've started looking through this, I'm making a few small adjustments. I ran out of time now, but I'll take a further look tomorrow.

benjie:
@mathroc Sorry for the delay on this, life has been really hectic!

I've just pushed a commit that uses pgColumnFilter in more contexts and fixes a few of the contexts it was being passed before. It's quite a big change but so long as we treat this as not a supported feature just yet I'm happy to release it - we might change it slightly in future to make sure it works consistently. Before it gets official support it's going to need a lot more tests doing things in a lot more interesting ways (e.g. preventing ordering by a column but nothing else; preventing a column appearing in the input type but nothing else; etc)

benjie:
GitHub seems to be having some issues right now - my commit is not showing up everywhere yet.

benjie:
Great; it's working now - tests passed 👍 

I'd love some more comments on this PR before it gets merged - especially from people willing to give it a try (or even write more tests for it).

benjie:
(Also @mathroc if you can confirm the latest code still works for your use case that'd be great - since you raised the PR from your master branch I had no choice but the push the commits there - I hope that hasn't caused you any issues?)

mathroc:
don’t worry, I made the branch editable just for that. I’ll give this new version a try next week

mathroc:
I just updated with this PR and it’s working nicely ☺

benjie:
Awesome 👍 

I'd love to hear from anyone else interested in using this functionality if it works for them or not.

packages/graphile-build-pg/src/plugins/PgJWTPlugin.js
```diff
@@ -61,6 +61,7 @@ export default (function PgJWTPlugin(
         throw new Error("JWT type has already been overridden?");
       }
       const attributes = introspectionResultsByKind.attribute
+        // TODO: consider adding to pgColumnFilter?
```
benjie: Remove?

packages/graphile-build-pg/src/plugins/PgOrderByPrimaryKeyPlugin.js
```diff
@@ -15,6 +15,7 @@ export default (function PgOrderByPrimaryKeyPlugin(builder) {
       const attributes = introspectionResultsByKind.attribute.filter(
         attr => attr.classId === table.id
       );
+      // todo
```
benjie: TODO

mathroc:
I don't mind an `_unstable` prefix and completely agree that you should break the compatibility as needed. I won't have time to work on this or remaining todos right away though

benjie:
I think we'll just say that it's undocumented and therefore experimental for now. Might rename it if I don't get a chance to test it fully before the v4 release.

All good, as soon as the tests pass I'm going to merge it 👍

PR #80, opened 2017-09-23
add foreign table to manyRelationByKeys call

### LINKED ISSUE #77: needs foreign table info in inflector.manyRelationByKeys
I’d like to know the foreign table name when building the manyRelationByKeys name

use case, column site on table discussion targets a row in table site

on the type `Site` I’d like to call the field `discussions` instead of `discussionBySite` because there is only one column on the key and that this column is named as the foreign key target table

mathroc:
closes #77

(not tested)


mathroc:
done :)

benjie:
🙏

PR #86, opened 2017-10-01
[WIP] feat(mutations): add option to enable default mutations only for the specified tables

knqyf263:
Close https://github.com/postgraphql/postgraphql/issues/576

I added the filter, but test failed.
How should I fix the test?
```
$ scripts/test --updateSnapshot
...
  ● mutation-delete-error.graphql

    GraphQLError: Schema is not configured for mutations

      at getOperationRootType (../../node_modules/graphql/execution/execute.js:261:15)
      at executeOperation (../../node_modules/graphql/execution/execute.js:225:14)
      at executeImpl (../../node_modules/graphql/execution/execute.js:133:26)
      at execute (../../node_modules/graphql/execution/execute.js:110:229)
      at ../../node_modules/graphql/graphql.js:75:34
          at Promise (<anonymous>)
      at graphqlImpl (../../node_modules/graphql/graphql.js:59:10)
      at graphql (../../node_modules/graphql/graphql.js:48:227)
      at withPgClient (__tests__/integration/mutations.test.js:57:28)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:188:7)
```

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
@@ -8,7 +8,7 @@ const debug = debugFactory("graphile-build-pg");
 
 export default (function PgMutationCreatePlugin(
   builder,
-  { pgInflection: inflection, pgDisableDefaultMutations }
+  { pgInflection: inflection, pgDisableDefaultMutations, pgEnableDefaultMutationTables }
 ) {
   if (pgDisableDefaultMutations) {
     return;
@@ -44,6 +44,11 @@ export default (function PgMutationCreatePlugin(
           .filter(table => !!table.namespace)
           .filter(table => table.isSelectable)
           .filter(table => table.isInsertable)
+          .filter(
+            table =>
+              (pgEnableDefaultMutationTables.length == 0) ||
+              (pgEnableDefaultMutationTables.indexOf(table.name) >= 0)
+          )
           .reduce((memo, table) => {
             const Table = getTypeByName(
               inflection.tableType(table.name, table.namespace.name)
--- a/packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
@@ -17,6 +17,9 @@ export default (function PgMutationCreatePlugin(
   if (pgDisableDefaultMutations) {
     return;
   }
+  if (!Array.isArray(pgEnableDefaultMutationTables)) {
+    pgEnableDefaultMutationTables = [];
+  }
   builder.hook(
     "GraphQLObjectType:fields",
     (
@@ -51,7 +54,9 @@ export default (function PgMutationCreatePlugin(
           .filter(
             table =>
               pgEnableDefaultMutationTables.length == 0 ||
-              pgEnableDefaultMutationTables.indexOf(table.name) >= 0
+              pgEnableDefaultMutationTables.indexOf(
+                `${table.namespace.name}.${table.name}`
+              ) >= 0
           )
           .reduce((memo, table) => {
             const Table = getTypeByName(
--- a/packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
@@ -3,6 +3,7 @@ import type { Plugin } from "graphile-build";
 import queryFromResolveData from "../queryFromResolveData";
 import debugFactory from "debug";
 import viaTemporaryTable from "./viaTemporaryTable";
+import { parseTypeIdentifier } from "./PgJWTPlugin";
 
 const debug = debugFactory("graphile-build-pg");
 
@@ -20,6 +21,11 @@ export default (function PgMutationCreatePlugin(
   if (!Array.isArray(pgEnableDefaultMutationTables)) {
     pgEnableDefaultMutationTables = [];
   }
+  pgEnableDefaultMutationTables = pgEnableDefaultMutationTables.map(function(
+    element
+  ) {
+    return parseTypeIdentifier(String(element));
+  });
   builder.hook(
     "GraphQLObjectType:fields",
     (
@@ -54,9 +60,12 @@ export default (function PgMutationCreatePlugin(
           .filter(
             table =>
               pgEnableDefaultMutationTables.length == 0 ||
-              pgEnableDefaultMutationTables.indexOf(
-                `${table.namespace.name}.${table.name}`
-              ) >= 0
+              pgEnableDefaultMutationTables.some(function(element) {
+                return (
+                  element.namespaceName == table.namespace.name &&
+                  element.typeName == table.name
+                );
+              })
           )
           .reduce((memo, table) => {
             const Table = getTypeByName(
--- a/packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
@@ -8,7 +8,11 @@ const debug = debugFactory("graphile-build-pg");
 
 export default (function PgMutationCreatePlugin(
   builder,
-  { pgInflection: inflection, pgDisableDefaultMutations, pgEnableDefaultMutationTables }
+  {
+    pgInflection: inflection,
+    pgDisableDefaultMutations,
+    pgEnableDefaultMutationTables,
+  }
 ) {
   if (pgDisableDefaultMutations) {
     return;
@@ -46,8 +50,8 @@ export default (function PgMutationCreatePlugin(
           .filter(table => table.isInsertable)
           .filter(
             table =>
-              (pgEnableDefaultMutationTables.length == 0) ||
-              (pgEnableDefaultMutationTables.indexOf(table.name) >= 0)
+              pgEnableDefaultMutationTables.length == 0 ||
+              pgEnableDefaultMutationTables.indexOf(table.name) >= 0
           )
           .reduce((memo, table) => {
             const Table = getTypeByName(
--- a/packages/graphile-build-pg/src/plugins/PgMutationUpdateDeletePlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgMutationUpdateDeletePlugin.js
@@ -20,6 +20,9 @@ export default (async function PgMutationUpdateDeletePlugin(
   if (pgDisableDefaultMutations) {
     return;
   }
+  if (!Array.isArray(pgEnableDefaultMutationTables)) {
+    pgEnableDefaultMutationTables = [];
+  }
   builder.hook(
     "GraphQLObjectType:fields",
     (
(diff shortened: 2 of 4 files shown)


packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
```diff
@@ -44,6 +44,11 @@ export default (function PgMutationCreatePlugin(
           .filter(table => !!table.namespace)
           .filter(table => table.isSelectable)
           .filter(table => table.isInsertable)
+          .filter(
+            table =>
+              (pgEnableDefaultMutationTables.length == 0) ||
```
benjie: Since the option above has no default, this line will throw an error if the caller doesn't specify the option.

packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
```diff
@@ -44,6 +44,11 @@ export default (function PgMutationCreatePlugin(
...
+              (pgEnableDefaultMutationTables.indexOf(table.name) >= 0)
```
benjie: `table.name` is not sufficient because tables with the same name might be sourced from multiple schemas; you'll need a bit of a more complex check here that takes the schema into consideration also (maybe a tuple?)

packages/graphile-build-pg/src/plugins/PgMutationUpdateDeletePlugin.js
```diff
@@ -56,6 +56,11 @@ export default (async function PgMutationUpdateDeletePlugin(
                   (mode === "update" && table.isUpdatable) ||
                   (mode === "delete" && table.isDeletable)
               )
+              .filter(
+                table =>
+                  (pgEnableDefaultMutationTables.length == 0) ||
+                  (pgEnableDefaultMutationTables.indexOf(table.name) >= 0)
```
benjie: Same comments as above

benjie:
CI doesn't get that far due to lint errors (these should be auto-fixable via `npm run lint -- --fix` in the relevant directory) so I can't see the exact issue, but I think it is probably because `pgEnableDefaultMutationTables` is not set; and thus `pgEnableDefaultMutationTables.length` is throwing; and when an error is thrown whilst building mutations we skip the mutations currently (this needs work, but the idea is to detect if no mutation was set).

packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
knqyf263: Oh, I misunderstood the option was set default value here.
https://github.com/postgraphql/postgraphql/pull/595/commits/0cefb2bf6809c6adcb345b6e69368c5d442b7444#diff-6c38b9313e10fd19f30ddbdd09c18c73R87

Where should I set the default value? Around here?
https://github.com/graphile/graphile-build/pull/86/commits/b4fbf693c1639003ef8110e029aeaae0d9085c0d#diff-886ba16c8a211054c79f05dbd24bac7fR14

knqyf263:
At first, I fixed lint errors.

packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
knqyf263: How about `schema.table` ? (e.g. public.post)
I think that it is easier to pass from cli (--enable-default-mutation-tables public.post).
If tuple, how should I receive the value of this option?

knqyf263:
@benjie I reflected your review. All tests have passed.
Please review again!

packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
```diff
@@ -44,6 +51,13 @@ export default (function PgMutationCreatePlugin(
           .filter(table => !!table.namespace)
           .filter(table => table.isSelectable)
           .filter(table => table.isInsertable)
+          .filter(
+            table =>
+              pgEnableDefaultMutationTables.length == 0 ||
+              pgEnableDefaultMutationTables.indexOf(
+                `${table.namespace.name}.${table.name}`
+              ) >= 0
```
benjie: I'm afraid this might be slightly too simple; e.g. it's possible to define tables that contain a `.` in their name, and if a table contains capital letters then people should wrap the table name in quotes otherwise the case-insensitivity of PostgreSQL will drop the capitals. 

I recommend you use the `parseTypeIdentifier` method from PGJWTPlugin.js:

https://github.com/graphile/graphile-build/blob/d2da24040a6a352679b5985b4829634f32396e7a/packages/graphile-build-pg/src/plugins/PgJWTPlugin.js#L120-L134

to extract the namespace and table names (which you should do just before the `builder.hook` call so it only has to be evaluated once at server start time) then check these during the filter.

packages/graphile-build-pg/src/plugins/PgMutationUpdateDeletePlugin.js
```diff
@@ -56,6 +63,13 @@ export default (async function PgMutationUpdateDeletePlugin(
                   (mode === "update" && table.isUpdatable) ||
                   (mode === "delete" && table.isDeletable)
               )
+              .filter(
+                table =>
+                  pgEnableDefaultMutationTables.length == 0 ||
+                  pgEnableDefaultMutationTables.indexOf(
+                    `${table.namespace.name}.${table.name}`
+                  ) >= 0
+              )
```
benjie: Likewise

knqyf263:
@benjie I fixed again!
Since I'm a JavaScript beginner, please tell me something strange.
e.g. Don't export `parseTypeIdentifier` ,etc.

knqyf263:
@benjie Is there any progress? 
If I need to fix more, I will do soon!

benjie:
Hey, sorry for the delay on this, my TODO list just keeps growing!

I think the code itself is fine (have not reviewed it in depth); I was going to implement some tests for it but I haven't had time. Can you add schema tests for me? Start here:

https://github.com/graphile/graphile-build/blob/master/packages/postgraphile-core/__tests__/integration/schema.test.js

We want to ensure that the relevant mutations are omitted when the setting is set - e.g. if you set it to just `[]` then there should be no default mutations, then try passing just a couple tables and ensuring that only those tables have mutations created.

👍  Thanks! 🙏

packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
```diff
@@ -44,6 +57,16 @@ export default (function PgMutationCreatePlugin(
           .filter(table => !!table.namespace)
           .filter(table => table.isSelectable)
           .filter(table => table.isInsertable)
+          .filter(
+            table =>
+              pgEnableDefaultMutationTables.length == 0 ||
+              pgEnableDefaultMutationTables.some(function(element) {
```
benjie: I think `pgEnableDefaultMutationTables  = []` should be equivalent to `pgDisableDefaultMutations = true` - it's a whitelist with no entries.

As such, this check should be more like:

packages/graphile-build-pg/src/plugins/PgMutationCreatePlugin.js
```diff
@@ -44,6 +57,16 @@ export default (function PgMutationCreatePlugin(
...
+                return (
+                  element.namespaceName == table.namespace.name &&
+                  element.typeName == table.name
```
benjie: We should change `parseTypeIdentifier` into simply `parseIndentifier` and move it into a common library file. We should also change `typeName` to `name` along with this change.

packages/graphile-build-pg/src/plugins/PgMutationUpdateDeletePlugin.js
```diff
@@ -5,17 +5,30 @@ import debugFactory from "debug";
...
+    pgInflection: inflection,
+    pgDisableDefaultMutations,
+    pgEnableDefaultMutationTables,
+  }
 ) {
   if (pgDisableDefaultMutations) {
     return;
   }
+  if (!Array.isArray(pgEnableDefaultMutationTables)) {
+    pgEnableDefaultMutationTables = [];
```
benjie: As above, prefer

packages/graphile-build-pg/src/plugins/PgMutationUpdateDeletePlugin.js
```diff
@@ -5,17 +5,30 @@ import debugFactory from "debug";
...
+  }
+  pgEnableDefaultMutationTables = pgEnableDefaultMutationTables.map(function(
```
benjie: pgEnableDefaultMutationTables = pgEnableDefaultMutationTables ? pgEnableDefaultMutationTables.map(function(...)) : null

packages/graphile-build-pg/src/plugins/PgMutationUpdateDeletePlugin.js
```diff
@@ -56,6 +69,16 @@ export default (async function PgMutationUpdateDeletePlugin(
                   (mode === "update" && table.isUpdatable) ||
                   (mode === "delete" && table.isDeletable)
               )
+              .filter(
+                table =>
+                  pgEnableDefaultMutationTables.length == 0 ||
```
benjie: `...Tables == null ||`

benjie:
Hey @knqyf263, are you still interested in getting this PR merged? It's pretty close!

benjie:
I still want this functionality, but I'm wondering if smart comments might be a better way of achieving it - more subtlety. You can then disable default mutations and enable them one mutation at a time via comments on the relevant tables.

```sql
create table tag (
  id serial primary key,
  title text
);

comment on table tag is '@allowInsert\n@allowDelete\nA tag to associate with items';
```

Anyone have any opinions on this?

mattbretl:
As an alternative (more readable?) syntax, what if we recognized `@allow` and `@prevent` followed by a comma-separated list of `select`/`insert`/`update`/`delete`?

With default mutations enabled, you could selectively disable them like this:
``` sql
create table tag (
  ...
);
comment on table tag is '@prevent update,delete\nA tag to associate with items';
```

With default mutations disabled, you could selectively enable them like this:
``` sql
create table tag (
  ...
);
comment on table tag is '@allow insert\nA tag to associate with items';
```

(TBH, I think the better way to filter the exposed mutations is to use `GRANT`s instead of smart comments, and then introspect the permissions of the role used by PostGraphile.  I have a way-too-early-to-publish v4 plugin that does this.  That said, I can still see valid use cases for this smart comment approach.)

benjie:
Yeah; introspecting the permissions is probably a better way. It's a challenge because you have to check not just the role itself but all roles that were granted to that role including NOINHERIT. I've heard it might be a slow (performance-wise) thing to do also, but caching should solve that. Seems `GRANT` and `REVOKE` fire the event triggers too, so we can still listen for them in watch mode 👍 

https://www.postgresql.org/docs/9.6/static/event-trigger-matrix.html

benjie:
This PR is being replaced by https://github.com/graphile/graphile-build/pull/199

Thanks so much for the PR - it really got me thinking 🙏

PR #93, opened 2017-10-16
add docker compose config for dev workflow

mathroc:
I don’t know if you’re interested in having this in the repository, that’s the config I’m using to work on graphile in docker

### DIFF
--- a/config/docker/Dockerfile
+++ b/config/docker/Dockerfile
@@ -1,0 +1,5 @@
+FROM node:8
+
+RUN apt-get update && apt-get install -y postgresql-client
+
+RUN npm install -g lerna yarm flow-bin
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,28 @@
+version: "3"
+services:
+  js: &js
+    build: config/docker/
+    volumes:
+      - .:/gb:rw
+    working_dir: /gb
+    depends_on: [db]
+    environment:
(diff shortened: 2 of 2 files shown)


config/docker/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM node:8
+
+RUN apt-get update && apt-get install -y postgresql-client
+
+RUN npm install -g lerna yarm flow-bin
```
benjie: Should that be yarn?

benjie:
Thanks for the PR; I'd prefer this as a wiki article for now (trying to keep everything as lean as possible), but we can revisit at a later date. Thanks!

config/docker/Dockerfile
mathroc: ahh in fact it shouldn’t be there at all, you just made me realize I’ve installed another package by mistake ^^

mathroc:
hmm I was going to take a look at [the wiki](https://github.com/graphile/graphile-build/wiki) but it redirect to the repository home 😕 or did you meant https://github.com/graphile/graphile.github.io ?

benjie:
Try now, I've opened the permissions up I think 👍

mathroc:
yes that worked : https://github.com/graphile/graphile-build/wiki/Development-with-docker-compose

thank you

benjie:
Awesome! Really appreciate that, thanks @mathroc 👍 

If you want to patch the README to link to the wiki page that would be welcome too.

Issue #88, opened 2017-10-10
Order by column on related table [plugin available]


mathroc:
I’ve tried to make a plugin to order by column on foreign table of a collection, I took inspiration from `PgOrderAllColumnsPlugin.js` however I don’t know how to go further, this is what I have:

```js

    const attrs = introspectionResultsByKind.attribute.filter(attr => attr.classId === table.id);

    const res = introspectionResultsByKind.constraint
      .filter(con => con.type === "f")
      .filter(con => con.classId === table.id)
      .map(con => ({con, foreignTable: introspectionResultsByKind.classById[con.foreignClassId]}))
      .map(({con, foreignTable}) => ({
        con,
        foreignTable,
        foreignAttrs: introspectionResultsByKind.attribute
          .filter(attr => attr.classId === foreignTable.id)
          .filter(attr => attr.type.type === "b")
          .filter(attr => attr.type.name !== "uuid")
      })).filter(({foreignAttrs}) => foreignAttrs.length > 0);

    return extend(values, res.reduce((memo, r) => {
      return r.foreignAttrs.reduce((memo, ra) => {
        [true, false].reduce((memo, dir) => {
          const ascFieldName = inflection.orderByEnum(
            [...r.con.keyAttributeNums.map(i => attrs.filter(a => a.num === i)[0].name), ra.name].join("-"),
            dir,
            table.name,
            table.namespaceName
          );
          console.log(table.name, ascFieldName);
          memo[ascFieldName] = {
            value: {
              alias: ascFieldName.toLowerCase(),
              specs: [["???", dir]]
            }
          };
          return memo;
        }, memo)
        return memo;
      }, memo)
    }, {}));
```

when I inspect the schema, I can see it adds the order bys I want. but I don’t know what to put instead of `"???"`. 

for starters I’d be happy if it works only when another part of the graphql query adds the "join" on the related table, but here I can’t yet know the subquery for that related table alias. I can see [here](https://github.com/graphile/graphile-build/blob/master/packages/graphile-build-pg/src/plugins/PgConnectionArgOrderBy.js#L88) that I could put something else than a string but I can’t find example of that and can’t think of any pg-sql expr would be useful in my case

is this possible without patching `PgConnectionArgOrderBy.js` ?

(related: https://github.com/postgraphql/postgraphql/issues/119#issuecomment-324275857)

benjie:
[...]

PR #94, opened 2017-10-16
give the QueryBuilder to callback function in callIfNecessary

mathroc:
see #88 

I have not found a way to get access to the QueryBuilder yet from a orderby callback spec. maybe some sleep will help, meanwhile. I could complete my task with this patch to the QueryBuilder

### DIFF
--- a/packages/graphile-build-pg/src/QueryBuilder.js
+++ b/packages/graphile-build-pg/src/QueryBuilder.js
@@ -10,10 +10,7 @@ type GenContext = {
 };
 type Gen<T> = (context: GenContext) => T;
 
-function callIfNecessary<T>(
-  o: Gen<T> | T,
-  context: GenContext
-): T {
+function callIfNecessary<T>(o: Gen<T> | T, context: GenContext): T {
   if (typeof o === "function") {
     return o(context);
   } else {
--- a/packages/graphile-build-pg/src/QueryBuilder.js
+++ b/packages/graphile-build-pg/src/QueryBuilder.js
@@ -5,7 +5,10 @@ import isSafeInteger from "lodash/isSafeInteger";
 
 const isDev = ["test", "development"].indexOf(process.env.NODE_ENV) >= 0;
 
-type Gen<T> = () => T;
+type GenContext = {
+  queryBuilder: QueryBuilder,
+};
+type Gen<T> = (context: GenContext) => T;
 
 function callIfNecessary<T>(
   o: Gen<T> | T,
@@ -23,7 +26,7 @@ function callIfNecessaryArray<T>(
   context: { queryBuilder: QueryBuilder }
 ): Array<T> {
   if (Array.isArray(o)) {
-    return o.map(callIfNecessary, context);
+    return o.map(v => callIfNecessary(v, context));
   } else {
     return o;
   }
@@ -489,6 +492,9 @@ class QueryBuilder {
   }
   lock(type: string) {
     if (this.locks[type]) return;
+    const getContext = () => ({
+      queryBuilder: this,
+    });
     const beforeLocks = this.data.beforeLock[type];
     this.data.beforeLock[type] = [];
     for (const fn of beforeLocks || []) {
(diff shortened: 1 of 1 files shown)


packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -7,17 +7,23 @@ const isDev = ["test", "development"].indexOf(process.env.NODE_ENV) >= 0;
...
 }
 
-function callIfNecessaryArray<T>(o: Array<Gen<T> | T>): Array<T> {
+function callIfNecessaryArray<T>(
+  o: Array<Gen<T> | T>,
+  context: { queryBuilder: QueryBuilder }
+): Array<T> {
   if (Array.isArray(o)) {
-    return o.map(callIfNecessary);
+    return o.map(callIfNecessary, context);
```
benjie: Rather: `return o.map(v => callIfNecessary(v, context));`

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -495,38 +501,51 @@ class QueryBuilder {
...
     } else if (type === "limit") {
-      this.compiledData[type] = callIfNecessary(this.data[type]);
+      this.compiledData[type] = callIfNecessary(this.data[type], {
+        queryBuilder: this,
+      });
     } else if (type === "offset") {
-      this.compiledData[type] = callIfNecessary(this.data[type]);
+      this.compiledData[type] = callIfNecessary(this.data[type], {
+        queryBuilder: this,
+      });
```
benjie: Let's DRY this up by putting something like:

```js
const getContext = () => ({
  queryBuilder: this,
})
```
at the top of the `lock` method and then calling it at each place instead:

```js
this.compiledData[type] = callIfNecessary(this.data[type], getContext());
```

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -7,17 +7,23 @@ const isDev = ["test", "development"].indexOf(process.env.NODE_ENV) >= 0;
...
 type Gen<T> = () => T;
 
-function callIfNecessary<T>(o: Gen<T> | T): T {
+function callIfNecessary<T>(
+  o: Gen<T> | T,
+  context: { queryBuilder: QueryBuilder }
+): T {
   if (typeof o === "function") {
-    return o();
+    return o(context);
```
benjie: I think we need to modify the type of `Gen<T>` to enable this. Probably just

```js
type GenContext = {
  queryBuilder: QueryBuilder,
};
type Gen<T> = (context: GenContext) => T;

function callIfNecessary<T>(
  o: Gen<T> | T,
  context: GenContext
): T {
...

function callIfNecessaryArray<T>(
  o: Array<Gen<T> | T>,
  context: GenContext
): Array<T> {
```

mathroc:
after some more research I’ve not found another solution for #88. and I don’t think there is one because the query builder has probably not been created before the call to the callback function. 
so I’ve updated the PR

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -5,19 +5,25 @@ import isSafeInteger from "lodash/isSafeInteger";
 
 const isDev = ["test", "development"].indexOf(process.env.NODE_ENV) >= 0;
 
-type Gen<T> = () => T;
+type Gen<T> = (context: GenContext) => T;
```
benjie: You omitted the definition of `GenContext`?

mathroc:
tests are green !

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -5,19 +5,28 @@ import isSafeInteger from "lodash/isSafeInteger";
...
-type Gen<T> = () => T;
+type GenContext = {
+  queryBuilder: QueryBuilder,
+};
+type Gen<T> = (context: GenContext) => T;
 
-function callIfNecessary<T>(o: Gen<T> | T): T {
+function callIfNecessary<T>(
+  o: Gen<T> | T,
+  context: { queryBuilder: QueryBuilder }
```
benjie: This should be context: GenContext I think?

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -5,19 +5,28 @@ import isSafeInteger from "lodash/isSafeInteger";
...
+    return o(context);
   } else {
     return o;
   }
 }
 
-function callIfNecessaryArray<T>(o: Array<Gen<T> | T>): Array<T> {
+function callIfNecessaryArray<T>(
+  o: Array<Gen<T> | T>,
+  context: { queryBuilder: QueryBuilder }
```
benjie: Ditto

benjie:
Hah; I wish prettier would run when editing via GitHub!

/me checks out the branch

benjie:
Merged; thanks! 🙏 

I've got a few other bugs to fix and I should be rolling it into a release late next week most likely.

PR #101, opened 2017-10-25
Improve pg view cursor support

d3k4r:
## Reason
Postgres views do not support constraints like the primary key constraint. graphile uses the primary key constraint to set the default orderBy which is used to create the unique cursors. The view cursors end up working only when the graphql orderBy parameter is always unique.

## Implementation (Method 1)
* Add pg class relkind to introspection query
* Add check for pg class relkind view and uniqueId attribute.
* In the case of no primary key and no configured view unique key then default to row number. ref #104

NOTE: ~~When creating the view in postgres view one of the columns needs to be named specifically **uniqueId**. e.g. `create view testview as select id as "uniqueId", name from device;`~~ If viewUniqueKey option is set to a column then it will be used.

## Verification
Added two test queries w/ a configured viewUniqueKey and updated the snapshot. The previous view queries have updated snapshot data because of the new default.

The existing tests continue to work. Manually tested queries against the view and the cursors now contain
```
echo WyJ1bmlxdWVfaWRfYXNjIixbImYyNjkzMjNjLThkZjctNTQ3MS05NTdjLTFmMjAwN2M3OTJhZiJdXQ== | base64 -d
["unique_id_asc",["f269323c-8df7-5471-957c-1f2007c792af"]]%
```
 and when there's an orderBy it looks like this
```
echo WyJoZWFydGJlYXRfZGVzYyIsWyIyMDE3LTA5LTEyVDA5OjA1OjQ2LjU1MyswMzowMCIsImYyNjkzMjNjLThkZjctNTQ3MS05NTdjLTFmMjAwN2M3OTJhZiJdXQ== | base64 -d
["heartbeat_desc",["2017-09-12T09:05:46.553+03:00","f269323c-8df7-5471-957c-1f2007c792af"]]%
```

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgAllRows.js
+++ b/packages/graphile-build-pg/src/plugins/PgAllRows.js
@@ -122,6 +122,11 @@ export default (async function PgAllRows(
                               );
                               builder.setOrderIsUnique();
                             });
+                          } else {
+                            builder.orderBy(
+                              sql.fragment`(row_number() over (partition by 1))`,
+                              true
+                            );
                           }
                         }
                       );
--- a/packages/graphile-build-pg/src/plugins/PgAllRows.js
+++ b/packages/graphile-build-pg/src/plugins/PgAllRows.js
@@ -58,7 +58,9 @@ export default (async function PgAllRows(
                 num => attributes.filter(attr => attr.num === num)[0]
               );
             const isView = t => t.classKind === "v";
-            const hasUniqueId = a => !!a.find(elem => elem.name === "uniqueId");
+            const uniqueIdAttribute = attributes.filter(
+              attr => attr.name === "uniqueId" || attr.name === "unique_id"
+            )[0];
             if (!ConnectionType) {
               throw new Error(
                 `Could not find GraphQL connection type for table '${table.name}'`
@@ -107,12 +109,12 @@ export default (async function PgAllRows(
                                 builder.setOrderIsUnique();
                               }
                             });
-                          } else if (isView(table) && hasUniqueId(attributes)) {
+                          } else if (isView(table) && !!uniqueIdAttribute) {
                             builder.beforeLock("orderBy", () => {
                               builder.data.cursorPrefix = ["unique_id_asc"];
                               builder.orderBy(
                                 sql.fragment`${builder.getTableAlias()}.${sql.identifier(
-                                  "uniqueId"
+                                  uniqueIdAttribute.name
                                 )}`,
                                 true
                               );
--- a/packages/graphile-build-pg/src/plugins/PgAllRows.js
+++ b/packages/graphile-build-pg/src/plugins/PgAllRows.js
@@ -57,6 +57,8 @@ export default (async function PgAllRows(
               primaryKeyConstraint.keyAttributeNums.map(
                 num => attributes.filter(attr => attr.num === num)[0]
               );
+            const isView = t => t.classKind === 'v';
+            const hasUniqueId = a => !!a.find(elem => elem.name === 'uniqueId');
             if (!ConnectionType) {
               throw new Error(
                 `Could not find GraphQL connection type for table '${table.name}'`
@@ -104,6 +106,15 @@ export default (async function PgAllRows(
                                 });
                                 builder.setOrderIsUnique();
                               }
+                            });
+                          } else if (isView(table) && hasUniqueId(attributes)) {
+                            builder.beforeLock("orderBy", () => {
+                              builder.data.cursorPrefix = ["unique_id_asc"];
+                              builder.orderBy(
+                                sql.fragment`${builder.getTableAlias()}.${sql.identifier('uniqueId')}`,
+                                true
+                              );
+                              builder.setOrderIsUnique();
                             });
                           }
                         }
--- a/packages/graphile-build-pg/src/plugins/PgAllRows.js
+++ b/packages/graphile-build-pg/src/plugins/PgAllRows.js
@@ -93,7 +93,7 @@ export default (async function PgAllRows(
                         builder => {
                           if (primaryKeys) {
                             builder.beforeLock("orderBy", () => {
-                              if (!builder.isOrderUnique()) {
+                              if (!builder.isOrderUnique(false)) {
                                 // Order by PK if no order specified
                                 builder.data.cursorPrefix = ["primary_key_asc"];
                                 primaryKeys.forEach(key => {
@@ -120,11 +120,6 @@ export default (async function PgAllRows(
                               );
                               builder.setOrderIsUnique();
                             });
-                          } else {
-                            builder.orderBy(
-                              sql.fragment`(row_number() over (partition by 1))`,
-                              true
-                            );
                           }
                         }
                       );
--- a/packages/graphile-build-pg/src/plugins/PgAllRows.js
+++ b/packages/graphile-build-pg/src/plugins/PgAllRows.js
@@ -58,9 +58,9 @@ export default (async function PgAllRows(
                 num => attributes.filter(attr => attr.num === num)[0]
               );
             const isView = t => t.classKind === "v";
-            const uniqueIdAttribute = attributes.filter(
+            const uniqueIdAttribute = attributes.find(
               attr => attr.name === "uniqueId" || attr.name === "unique_id"
-            )[0];
+            );
             if (!ConnectionType) {
               throw new Error(
                 `Could not find GraphQL connection type for table '${table.name}'`
--- a/packages/graphile-build-pg/src/plugins/PgAllRows.js
+++ b/packages/graphile-build-pg/src/plugins/PgAllRows.js
@@ -9,7 +9,7 @@ const debugSql = debugFactory("graphile-build-pg:sql");
 
 export default (async function PgAllRows(
   builder,
-  { pgInflection: inflection }
+  { pgInflection: inflection, pgViewUniqueKey: viewUniqueKey }
 ) {
   builder.hook(
     "GraphQLObjectType:fields",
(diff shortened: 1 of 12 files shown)


packages/graphile-build-pg/src/plugins/PgAllRows.js
```diff
@@ -57,6 +57,8 @@ export default (async function PgAllRows(
               primaryKeyConstraint.keyAttributeNums.map(
                 num => attributes.filter(attr => attr.num === num)[0]
               );
+            const isView = t => t.classKind === "v";
+            const hasUniqueId = a => !!a.find(elem => elem.name === "uniqueId");
```
benjie: We should allow the snake-case version also. If both are present we should use the one with the smaller `num` value

benjie:
This approach seems reasonable. I'm not sure if `uniqueId` is sufficiently explicit and non-breaking. That said if we can squeeze this in before v4 goes live then we can list it in breaking changes.

packages/graphile-build-pg/src/plugins/PgAllRows.js
d3k4r: Good point, I'll add that.

packages/graphile-build-pg/src/plugins/PgAllRows.js
```diff
@@ -58,7 +58,9 @@ export default (async function PgAllRows(
                 num => attributes.filter(attr => attr.num === num)[0]
               );
             const isView = t => t.classKind === "v";
-            const hasUniqueId = a => !!a.find(elem => elem.name === "uniqueId");
+            const uniqueIdAttribute = attributes.filter(
+              attr => attr.name === "uniqueId" || attr.name === "unique_id"
+            )[0];
```
benjie: Prefer `.find(...)` over `.filter(...)[0]` as it does less work and consumes less memory.

d3k4r:
FYI, when both snake_case (unique_id) and camelCase (uniqueId) are provided then they both end up with the same converted graphql name uniqueId. Then only one is able to be queried. e.g. `create view testsnakebeforecamel as select id as "unique_id", name as "uniqueId" from device;` correctly uses unique_id for the cursor but only the name is accessible in the node as uniqueId.

packages/graphile-build-pg/src/plugins/PgAllRows.js
d3k4r: okay

benjie:
Thinking about it, I'd expect that example to throw actually. If it doesn't that means there's a bug in the column plugin. Good spot!

benjie:
I'm still concerned about the uniqueId field name; for example:

```sql
create table foo (
  unique_id serial primary key,
  name text,
  foo_thing_1 text,
  ...
);

create table bar (
  unique_id serial primary key,
  name text,
  bar_thing_1 text,
  ...
);

create view foo_and_bar as
  select unique_id, name from foo
  union all
  select unique_id, name from bar;
```

It doesn't seems sufficiently special/unique for this purpose which risks it going wrong for someone.

How about this: we make it an option. By default it is something super unweildy such as `postgraphqlViewUID` / `postgraphqlViewUid` / `postgraphql_view_uid` but you can override it with a command-line flag. Thoughts?

Also before I can merge this PR it's going to need some tests:

- create a query file in here https://github.com/graphile/graphile-build/tree/master/packages/postgraphile-core/__tests__/fixtures/queries
- tweak the schema file here https://github.com/graphile/graphile-build/blob/master/packages/postgraphile-core/__tests__/kitchen-sink-schema.sql

Great work so far! 👍

d3k4r:
yea, totally agree that it's not special/unique enough. I like the idea of making it configurable. That seems like the ideal way to do it. I'm not sure how to do it as I'm not yet super familiar with this code but I'll look into it.

How about if we default it to use the first attribute/column but if an option/flag is configured then it uses that attribute name? It seems like it would potentially make views in general more usable for cursor pagination.

I'll add some test cases.

benjie:
First column isn't going to work for a good proportion of views I've seen. Also thinking about it we might need to allow compound uniqueness (e.g. it might be unique on `type` and `id` together). I think this is a good first solution though.

To add it as an option:

1/ Destructure the option from the options argument

https://github.com/graphile/graphile-build/blob/db30a95d80b5eca17b78cdedda63b778b9463c37/packages/graphile-build-pg/src/plugins/PgAllRows.js#L12

2/ Pass the option through from PostGraphQL (you ought to rename it to follow the PostGraphQL rather than Graphile conventions)

https://github.com/graphile/graphile-build/blob/db30a95d80b5eca17b78cdedda63b778b9463c37/packages/postgraphile-core/src/index.js#L25-L37

https://github.com/graphile/graphile-build/blob/db30a95d80b5eca17b78cdedda63b778b9463c37/packages/postgraphile-core/src/index.js#L70-L79

https://github.com/graphile/graphile-build/blob/db30a95d80b5eca17b78cdedda63b778b9463c37/packages/postgraphile-core/src/index.js#L103-L116

Then once this is merged we'll expose it as a CLI option in PostGraphQL itself via:

https://github.com/postgraphql/postgraphql/blob/benjie/graphql-build/src/postgraphql/cli.ts

d3k4r:
> First column isn't going to work for a good proportion of views I've seen. Also thinking about it we might need to allow compound uniqueness (e.g. it might be unique on type and id together). I think this is a good first solution though.

yeah, first column could have some problems of it's own. Main thing I was thinking was that the current default gives a unique cursor when the data is not ordered, e.g. 
```
echo WyJuYXR1cmFsIiwxXQ== | base64 -d
["natural",1]%
```
But then when an orderBy parameter is added it doesn't include that natural uniqueness, e.g.
```
echo WyJoZWFydGJlYXRfZGVzYyIsWyIyMDE3LTA5LTIwVDE2OjMzOjMyLjQ0NiswMzowMCJdXQ== | base64 -d
["heartbeat_desc",["2017-09-20T16:33:32.446+03:00"]]%
```

benjie:
The current cursors when there is no unique order by use the row number so they're absolutely terrible and should not be relied on. It's just a hack so we at least *have* cursor, but the cursor value for the exact same entry will change if a row is inserted ahead of it so it's dreadful. I've not figured out how to do it better yet. Basically it just exposes standard limit/offset but via a cursors interface. 😖 

The cursors should only switch to using the orderby if it's unique. In the case of views where order by is not unique we should maybe append row_number over the order by grouping; this at least will be more reliable because it'll only change if another row with the exact same orderby value is inserted.

Maybe we should make orderBy on views required.

TL;DR: Views are a pain because it's hard for the computer to know their intent

d3k4r:
> TL;DR: Views are a pain because it's hard for the computer to know their intent

😄 true that

I think once this configurability is in there and maybe this appending of the row_number then we are already in a much better situation. Still room for improvement but these views are tough! They are also super handy though. :)

d3k4r:
Closing until it's ready for review again. Still needs work for the appending of the row number or making orderBy on views required and missing tests.

benjie:
I'm afraid the SQL for this is quite complex due to the `row_number()` stuff due to non-uniqueness. Are you sure the issue above is newly introduced and not an existing issue?

benjie:
The non-unique-cursor issue, yeah.

d3k4r:
yea, it's an existing issue afaik.

benjie:
Raised as https://github.com/graphile/graphile-build/issues/104

d3k4r:
@benjie I'm not 100 percent sure about the default orderBy code addition. Although, it works how I expect it to work.

benjie:
Just merged the latest master into this (it's changed a bit)

packages/graphile-build-pg/src/plugins/PgAllRows.js
```diff
@@ -103,6 +107,24 @@ export default (async function PgAllRows(
...
+                                )}`,
+                                true
+                              );
+                              builder.setOrderIsUnique();
+                            });
+                          } else {
+                            builder.orderBy(
+                              sql.fragment`(row_number() over (partition by 1))`,
+                              true
+                            );
```
benjie: I'm going to remove this `else` clause so we don't change the existing cursor format.

benjie: No I'm not... But I'm going to tweak it slightly.

benjie:
I forgot from when I wrestled with this a couple months ago... We can't do this.

Reason is that `row_number()` cannot be used in `WHERE` clauses, and when using `before`/`after` we're passing the cursor generator which will include that `row_number()`

We need a different way of detecting uniqueness.

Options I've evaluated:

- using `oids` - discounted because they may be turned off and they don't apply to views
- using `ctid` - discounted because it doesn't apply to views and it changes
- using every value of the row - discounted because it could leak columns that should be hidden by `pgColumnFilter` or similar
- using a hashed version of every value of the row - discounted because it's too expensive

I think our best bet is to, in the case of views or tables without a primary key, just have the cursor be the offset and only use offset in all cases - that way it's basically perfectly equivalent to `limit/offset` based pagination.

benjie:
I've just pushed some code that should hopefully fix cursors on tables lacking primary keys also (in future we should look for unique constraints if PKs aren't available, but I don't have time to do that right now) which cascades into views (and probably set-returning function calls too, I'm guessing).

packages/graphile-build-pg/src/plugins/PgAllRows.js
```diff
@@ -103,6 +107,21 @@ export default (async function PgAllRows(
...
+                                ];
+                                builder.orderBy(
+                                  sql.fragment`${builder.getTableAlias()}.${sql.identifier(
+                                    uniqueIdAttribute.name
+                                  )}`,
+                                  true
+                                );
+                                builder.setOrderIsUnique();
+                              }
+                            });
```
benjie: I tweaked this so that other plugins can add unique orders to views first if they so desire - this way we only add the order if it's not already unique.

packages/postgraphile-core/__tests__/integration/__snapshots__/queries.test.js.snap
```diff
@@ -374,47 +374,47 @@ Object {
     "i": Object {
       "edges": Array [
         Object {
-          "cursor": "WyJjb25zdGFudF9hc2MiLFsyXV0=",
+          "cursor": "WyJjb25zdGFudF9hc2MiLDFd",
```
benjie: This is a view without a unique ordering so its cursor has changed. This one went from `["constant_asc",[2]]` to `["constant_asc",1]`, `1` being the `row_number()`

PR #102, opened 2017-10-25
Add pg view unique id cursor plugin (Method 2)

d3k4r:
## Reason
Postgres views do not support constraints like the primary key constraint. graphile uses the primary key constraint to set the default orderBy which is used to create the unique cursors. The view cursors end up working only when the graphql orderBy parameter is always unique.

## Implementation (Method 2)
* Add plugin for replacing the resolver in the views with uniqueId
* Add PgViewPlugin someplace after PgAllRows.
* Add spread operator to babel

NOTE: When creating the view in postgres view one of the columns needs to be named specifically **uniqueId**. e.g. `create view testview as select id as "uniqueId", name from device;`

## Verification
The existing tests continue to work. Manually tested queries against the view and the cursors now contain
```
echo WyJ1bmlxdWVfaWRfYXNjIixbImYyNjkzMjNjLThkZjctNTQ3MS05NTdjLTFmMjAwN2M3OTJhZiJdXQ== | base64 -d
["unique_id_asc",["f269323c-8df7-5471-957c-1f2007c792af"]]%
```
 and when there's an orderBy it looks like this
```
echo WyJoZWFydGJlYXRfZGVzYyIsWyIyMDE3LTA5LTEyVDA5OjA1OjQ2LjU1MyswMzowMCIsImYyNjkzMjNjLThkZjctNTQ3MS05NTdjLTFmMjAwN2M3OTJhZiJdXQ== | base64 -d
["heartbeat_desc",["2017-09-12T09:05:46.553+03:00","f269323c-8df7-5471-957c-1f2007c792af"]]%
```

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgViewPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgViewPlugin.js
@@ -1,0 +1,91 @@
+"use strict";
+
+Object.defineProperty(exports, "__esModule", {
+  value: true
+});
+
+var _queryFromResolveData = require("../queryFromResolveData");
+
+var _queryFromResolveData2 = _interopRequireDefault(_queryFromResolveData);
+
+var _debug = require("debug");
+
+var _debug2 = _interopRequireDefault(_debug);
+
+var _addStartEndCursor = require("./addStartEndCursor");
+
+var _addStartEndCursor2 = _interopRequireDefault(_addStartEndCursor);
+
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
+
+const debugSql = (0, _debug2.default)("graphile-build-pg:sql");
+
+exports.default = async function PgViewPlugin(builder, { pgInflection: inflection }) {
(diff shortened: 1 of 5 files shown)


packages/graphile-build-pg/src/plugins/PgViewPlugin.js
```diff
@@ -0,0 +1,132 @@
...
+
+var _debug2 = _interopRequireDefault(_debug);
+
+var _addStartEndCursor = require("./addStartEndCursor");
+
+var _addStartEndCursor2 = _interopRequireDefault(_addStartEndCursor);
+
+function _interopRequireDefault(obj) {
+  return obj && obj.__esModule ? obj : { default: obj };
+}
```
benjie: You seem to have added the compiled version rather than the raw source?

d3k4r: oh yea whoops. I'll fix that. I had been editing the raw source when developing it then just copied it over.

d3k4r:
Closing this in favor the other method.

PR #109, opened 2017-10-30
Issues with custom types / malformed record literal

ganlub:
I added some tests for the apparent issue with `viaTemporaryTable` . While doing that I found another  `malformed record literal` issue related with the `comptypes`. I added a test for that on `mutation-create`.

As per our discussion we found this other [issue](https://github.com/postgraphql/postgraphql/blob/607f6629986735139e2b77d5a9b1143846c4f691/src/postgraphql/withPostGraphQLContext.ts#L218-L230) where the queries are logged before execution, so if the second query never gets logged then it's never ran and thus the error does not exist there.

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgColumnsPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgColumnsPlugin.js
@@ -69,46 +69,54 @@ export default (function PgColumnsPlugin(
                 pgGetGqlTypeByTypeId(attr.typeId) || GraphQLString;
               addDataGenerator(parsedResolveInfoFragment => {
                 const { alias } = parsedResolveInfoFragment;
-                if (attr.type.type === "c") {
-                  return {
-                    pgQuery: queryBuilder => {
-                      // json_build_object
-                      /*
-                      queryBuilder.select(
-                        sql.identifier(queryBuilder.getTableAlias(), attr.name),
-                        alias
-                      );
-                      */
-                      const resolveData = getDataFromParsedResolveInfoFragment(
-                        parsedResolveInfoFragment,
-                        ReturnType
-                      );
-                      const jsonBuildObject = queryFromResolveData(
-                        sql.identifier(Symbol()), // Ignore!
+                return {
+                  pgQuery: queryBuilder => {
+                    const blahForType = (sqlFullName, type) => {
+                      if (type.arrayItemType) {
+                        const ident = sql.identifier(Symbol());
+                        return sql.fragment`
+                          (
+                            select json_agg(${blahForType(
+                              ident,
+                              type.arrayItemType
+                            )})
+                            from unnest(${sqlFullName}) as ${ident}
+                          )
+                        `;
+                      } else if (type.type === "c") {
+                        // json_build_object
+                        /*
+                          queryBuilder.select(
+                            sql.identifier(queryBuilder.getTableAlias(), name),
+                            alias
+                          );
+                          */
+                        const resolveData = getDataFromParsedResolveInfoFragment(
+                          parsedResolveInfoFragment,
+                          ReturnType
+                        );
+                        const jsonBuildObject = queryFromResolveData(
+                          sql.identifier(Symbol()), // Ignore!
+                          sqlFullName,
+                          resolveData,
+                          { onlyJsonField: true }
+                        );
+                        return jsonBuildObject;
+                      } else {
+                        return pgTweakFragmentForType(sqlFullName, type);
+                      }
+                    };
+                    queryBuilder.select(
+                      blahForType(
                         sql.fragment`(${queryBuilder.getTableAlias()}.${sql.identifier(
                           attr.name
                         )})`, // The brackets are necessary to stop the parser getting confused, ref: https://www.postgresql.org/docs/9.6/static/rowtypes.html#ROWTYPES-ACCESSING
-                        resolveData,
-                        { onlyJsonField: true }
-                      );
-                      queryBuilder.select(jsonBuildObject, alias);
-                    },
-                  };
-                } else {
-                  return {
-                    pgQuery: queryBuilder => {
-                      queryBuilder.select(
-                        pgTweakFragmentForType(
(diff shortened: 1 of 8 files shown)


benjie:
Thanks for reproducing :+1:

`{ error: malformed record literal: "{"headline":"headline","body":"body","authorId":9000,"enums":["awaiting"],"comptypes":[{"schedule":"2009-10-24 10:23:54+02","isOptimised":true}]}"`

Seems to be that we're using JSON where we oughtn't be.

farant:
Hey guys, I think I'm running into the same issue. The only (possible?) difference being that I'm passing an array of a custom type instead of an array of table rows.

So instead of 
```
create function a.post_many(posts a.post[]) ...
```

It would be
```
create function a.comptype_test(items a.comptype[]) ...
```

I noticed that `a.post` has the column `comptypes a.comptype[]` so it might be causing the issue with `a.post_many` as well. 

Thanks for all the work you've been putting into this @benjie! Super excited to start using v4 :)

benjie:
Custom types and row types are treated basically the same in this regard 👍

chadfurman:
@benjie this just came up for us, also.  Any thoughts on what's causing it?

chadfurman:
I see, so this PR adds tests that cause the error, and the tests need to be made to pass before this PR can be merged.

What graphile-build files should we look at?

benjie:
I'm not sure; but this is my highest priority issue right now (sadly I can't work on it today).

I thought it was related to `viaTemporaryTable`:

https://github.com/graphile/graphile-build/blob/99e845a6100f6e1697dfaff8f50f0caa91daef61/packages/graphile-build-pg/src/plugins/viaTemporaryTable.js#L76-L85

but @ganlub says the relevant query for that was not logged which implies the issue occurs before that. So it's probably related to mapping the value from the JSON received from GraphQL to the `sql.value(...)` being executed; as such it's most likely an issue with `gql2pg`:

https://github.com/graphile/graphile-build/blob/99e845a6100f6e1697dfaff8f50f0caa91daef61/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js#L135-L141

I wonder if it's falling through to the `} else {` statement due to the `pg2GqlMapper` not being registered early enough?

My first port of call would be to make sure that only this one failing test was being executed; and then I'd turn on `DEBUG="graphile*"` env-var and then I'd probably just sprinkle `console.log` and `debugger` statements everywhere. Note `debugger` works with the latest Node versions even from within Jest now.

If you plan to take this on please log that you're doing so (and when you stop) on this issue so we can benefit from each other's work.

benjie:
So to set up the tests as I would:

1/ delete everything but `j: ...` from `mutation-create.graphql`
2/ do the same to the snapshot file: https://github.com/ganlub/graphile-build/blob/cc21b1fb0e982d0a7b993ce8c65ab4d083dda561/packages/postgraphile-core/__tests__/integration/__snapshots__/mutations.test.js.snap#L6-L353
3/ change this line:

https://github.com/ganlub/graphile-build/blob/cc21b1fb0e982d0a7b993ce8c65ab4d083dda561/packages/postgraphile-core/__tests__/integration/mutations.test.js#L25

to read

```js
const mutationFileNames = ['mutation-create.graphql'] || readdirSync(mutationsDir);
```
(The `||` isn't needed, it's just a more interesting way of commenting ;) )

4/ cd into `postgraphile-core` and run `npm run test -- __tests__/integration/mutations.test.js`

chadfurman:
One thing I will note is that I'm only experiencing this problem now that I'm no longer able to pass through a string for my custom type but instead must pass through an object.

benjie:
Were you passing a raw PG string before, like "(1,2,3)" before?

chadfurman:
yes

chadfurman:
actually, this even happened when I reverted to alpha2.19 -- I'm guessing because of the version locking bug you fixed in 2.21

benjie:
Probably. Sorry 😞

chadfurman:
@benjie I've not had a chance to take this yet.  Any chance you'll get to it today?

benjie:
I've just started looking at this and can confirm that the pg2GqlMapper is *NOT* being called within gql2pg for this custom type.

Will continue digging.

benjie:
Ah yes, we're registering the pg2GqlMapper for the base type but the array type is not handled.

https://github.com/graphile/graphile-build/blob/cab2d3bc51963c043ad08e3bbd608c40e225d8b1/packages/graphile-build-pg/src/plugins/PgTablesPlugin.js#L161-L165

I think I can fix this in gql2pg and thus everything will be fixed at once.

benjie:
So I have that working, but it has surfaced a different issue:

```
    +    "j": Object {
    +      "post": Object {
    +        "comptypes": Array [
    +          Object {
    +            "isOptimised": null,
    +            "schedule": "2009-10-24T04:23:54-04:00",
    +          },
    +        ],
    +        "headline": "super headline",
    +        "id": 12,
    +      },
    +    },
```

the `isOptimised` is being pulled back out as `null` instead of `true`

benjie:
Ah; that seems to be an issue with a missing field alias.

benjie:
So the inputting of arrays of custom types is now fixed in a fairly clean way.

However actually performing the mutation on these, which goes via `viaTemporaryTable`, is going to be a lot more work to solve.

benjie:
Plan is to dynamically construct an array mapping over `gql2pg(pg2gql(value, ...))`, then `unnest` this array again using the table syntax `with foo as (select * from unnest(ARRAY[...stuff from above...]) as table (...columns from above...)) select ...normal stuff here...`

But I gotta dash out so can't write it now. Might not get around to it today, but at least have a plan!

benjie:
Even better; have managed to solve it by having Postgres turn it into a text representation and back again - that way we know it should definitely be able to process it!

😅 

Tidying up.

benjie:
I've concluded one thing: I'm going to have to upgrade prettier because reading these ternaries is a nightmare in the way old prettier has formatted them!

PR #113, opened 2017-11-01
correctly detect non nullable domain

### LINKED ISSUE #111: not null domain column shown as nullable in graphql schema
I have not yet took a deep look at the issue, this is the sql migration that introduced this kind of domains:

```sql
-- Deploy beesquad:graphql to pg

BEGIN;

CREATE DOMAIN graphql.url AS CHARACTER VARYING(2048);
CREATE DOMAIN graphql.not_null_url AS graphql.url NOT NULL;
CREATE DOMAIN graphql.keywords AS CHARACTER VARYING[] NOT NULL;

CREATE TYPE graphql.page AS (
	url graphql.not_null_url,
	canonical_url graphql.url,
	keywords graphql.keywords
);

COMMENT ON COLUMN graphql.page.url IS 'The page url';
COMMENT ON COLUMN graphql.page.canonical_url IS 'The page canonical url';
COMMENT ON COLUMN graphql.page.keywords IS 'The page keywords';

ALTER TABLE graphql.discussion ADD COLUMN origin graphql.page NOT NULL;

COMMENT ON COLUMN graphql.discussion.origin IS 'The page where the discussion have been initiated';

COMMIT;
```

In the schema both the fields origin.url & origin.keywords are shown as nullable

mathroc:
closes #111

### DIFF
--- a/packages/postgraphile-core/__tests__/integration/__snapshots__/schema.test.js.snap
+++ b/packages/postgraphile-core/__tests__/integration/__snapshots__/schema.test.js.snap
@@ -689,6 +689,8 @@ interface Node {
   id: ID!
 }
 
+scalar NotNullUrl
+
 # Information about pagination in a connection.
 type PageInfo {
   # When paginating forwards, the cursor to continue.
@@ -743,6 +745,8 @@ enum PeopleOrderBy {
   NATURAL
   PRIMARY_KEY_ASC
   PRIMARY_KEY_DESC
+  SITE_ASC
+  SITE_DESC
 }
 
 # Person test comment
@@ -831,6 +835,7 @@ type Person implements Node {
   # The person’s name
   name: String!
   rowId: Int!
+  site: WrappedUrl
 }
 
 # A condition to be used against \`Person\` object types. All fields are tested for equality and combined with a logical ‘and.’
@@ -849,6 +854,9 @@ input PersonCondition {
 
   # Checks for equality with the object’s \`rowId\` field.
   rowId: Int
+
+  # Checks for equality with the object’s \`site\` field.
+  site: WrappedUrlInput
 }
 
 # An input for mutations affecting \`Person\`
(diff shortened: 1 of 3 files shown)


mathroc:
wip: needs tests

packages/postgraphile-core/__tests__/integration/__snapshots__/schema.test.js.snap
```diff
@@ -1228,6 +1238,15 @@ type UpdatePersonPayload {
...
 # A universally unique identifier as defined by [RFC 4122](https://tools.ietf.org/html/rfc4122).
 scalar UUID
+
+type WrappedUrl {
+  url: NotNullUrl!
+}
+
+# An input for mutations affecting \`WrappedUrl\`
+input WrappedUrlInput {
+  url: NotNullUrl
```
mathroc: shouldn’t this one be not null too?

benjie: Yes if it doesn’t have a default 👍

benjie: Ah you gave it a default, but set the default to null. I don’t really see this as a bug; but please remove “default null” and see if that fixes it. Otherwise there’s probably another nullableIf we need to track down.

mathroc: I gave a `default null` to the `site b.wrapped_url` not the url inside. so `site` inside the `Person` & `PersonInput` is correctly nullable. but the `url` inside `WrappedUrlInput` should not be. and indeed if I replace `site b.wrapped_url default null` with `site b.wrapped_url not null` it does not affect the `WrappedUrlInput` type

benjie: Ah, my bad 👍

mathroc:
everything seems fine now !

benjie:
Thanks! 👍

PR #117, opened 2017-11-16
Fix example in README.md

mattbretl:
With these changes, the approach in this example can be executed (assuming the proper schema is already in place).

FYI, there is a very similar example in graphile.github.io/src/pages/graphile-build-pg/security.md, so if/when this is merged, I can clean up that file as well.

That said.. is there any reason you're not using the forum example schema here?  In general, I think the examples would be a lot more useful if they could be run against a known schema.

### DIFF
--- a/packages/postgraphile-core/README.md
+++ b/packages/postgraphile-core/README.md
@@ -37,11 +37,14 @@ const schema = await createPostGraphQLSchema(
 Full example:
 
 ```js
-const { createPostGraphQLSchema } = require('graphile-build-pg');
-const pg = require('pg');
+const { createPostGraphQLSchema } = require("postgraphile-core");
+const { graphql } = require("graphql");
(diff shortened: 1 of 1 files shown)


packages/postgraphile-core/README.md
```diff
@@ -98,7 +101,7 @@ runQuery(
   "query MyQuery { allPosts { nodes { id, title, author: userByAuthorId { username } } } }"
 ).then(result => {
   console.dir(result);
-  pgPool.release();
+  pgPool.end();
```
benjie: This always catches me out!

benjie:
I scratched all this stuff out as fast as I possibly could; virtually none of it has been checked. Really appreciate the help fixing it 👍 

Using it with a known schema is probably a good idea; I can't think of a good reason not to.

PR #126, opened 2017-12-03
chore(test): Assert that partial indexes don't get exported

benjie:
They don't (because no indexes get exported, only constraints - and constraints can't be partial), but this will enforce that's always the case.


benjie:
[Unique constraint with WHERE clause should not be added to schema](https://trello.com/c/FT5xvDOH/20-unique-constraint-with-where-clause-should-not-be-added-to-schema)

PR #128, opened 2017-12-05
Deprecate fields from column comments

mattbretl:
See https://github.com/postgraphql/postgraphql/issues/473 for background.

I still need to add a test to postgraphile-core.  Also, do you like PgColumnDescriptionsPlugin like I've written it here, or should we split `description` and `deprecationReason` into separate plugins?


packages/graphile-build-pg/src/plugins/PgColumnDescriptionsPlugin.js
```diff
@@ -0,0 +1,34 @@
...
+    if (
+      !(isPgRowType || isPgCompoundType) ||
+      !table ||
+      table.kind !== "class" ||
+      !pgFieldIntrospection ||
+      typeof pgFieldIntrospection.description !== "string"
+    ) {
+      return field;
+    }
+    const parsed = parseTags(pgFieldIntrospection.description);
```
benjie: So I've been thinking about this, and I think we should move the `parseTags` call into `PgIntrospectionQuery` - then other plugins can make use of the other tags also without having to re-parse the tags. We can also implement this for all the other types (functions, tables, etc) so that they don't have tags included in their `description` fields.

benjie: What do you think?

packages/graphile-build-pg/src/plugins/PgColumnDescriptionsPlugin.js
```diff
@@ -0,0 +1,34 @@
...
+    return parsed.tags.deprecated
+      ? Object.assign({}, field, {
+          description: parsed.text,
+          deprecationReason: parsed.tags.deprecated,
+        })
+      : Object.assign({}, field, {
+          description: parsed.text,
+        });
```
benjie: This would become `pgFieldIntrospection.tags` rather than `parsed.tags` is we move the tag parsing to the introspection plugin.

packages/graphile-build-pg/src/utils.js
```diff
@@ -44,3 +44,24 @@ export const camelCase = formatInsideUnderscores(camelCaseAll);
...
+export const parseTags = (str: string) => {
+  return str.split(`\n`).reduce((prev, curr) => {
+    const match = curr.match(/^@[a-z]+ ?/);
+    return match &&
+    prev.text === "" &&
+    curr.split(" ")[0] === match[0].split(" ")[0]
+      ? Object.assign({}, prev, {
+          tags: Object.assign({}, prev.tags, {
+            [match[0].substr(1).trim()]:
+              match[0] === curr ? true : curr.replace(match[0], ""),
```
benjie: Let's turn this into an array if there's multiple matches. i.e.

benjie: Let's turn this into an array if there's multiple matches. i.e.

```
@jsonField date timestamp
@jsonField name text
@jsonField episode enum ONE=1 TWO=2
```

benjie: Should produce `{jsonField: ["date timestamp", "name text", "episode enum ONE=1 TWO=2"]}`

benjie: (This is an enhancement to the previous spec in order to enable https://github.com/postgraphql/postgraphql/issues/642#issuecomment-348731162 )

mattbretl: Makes sense.  Should we allow mixed case tag names like `jsonField`?  The original spec was `/@[a-z]+/`.

benjie: Definitely; that was a bug in my spec, sorry! Probably underscores too. Maybe even numbers... enforce it starts with [a-zA-Z] though.

packages/graphile-build-pg/src/plugins/PgColumnDescriptionsPlugin.js
benjie: \*PgIntrospectionPlugin

mattbretl: I agree.  The only downside is tighter coupling - we won't have an easy way to drop the tagging feature from the plugin stack.  But if you're willing to support this as a first-class feature in the core, I think the pros (avoiding re-parsing, including them on other types, and simpler development of tag-based plugins) outweigh that con.

benjie: Good point; let's add "pgEnableTags" as an option and have it default true in `postgraphile-core` - that way people who don't want it can turn it off (and we'd just set `tags` to be `[]` on the relevant introspection fields).

benjie: (I do want tags in core, they're going to be super useful for things that can't be done automatically 👍 )

mattbretl:
Ready for another review.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -57,6 +58,15 @@ export default (async function PgIntrospectionPlugin(
...
         (memo, { object }) => {
+          if (object.description) {
+            if (pgEnableTags) {
+              const parsed = parseTags(object.description);
+              object.tags = parsed.tags;
+              object.description = parsed.text;
+            } else {
+              object.tags = {};
+            }
+          }
```
benjie: This is a bit of a jammy place to add this code. It's not really the "proper" place; but I'll let it slide since it gets the job done - we can refactor later.

packages/graphile-build-pg/src/utils.js
```diff
@@ -44,3 +44,33 @@ export const camelCase = formatInsideUnderscores(camelCaseAll);
...
+        text: curr,
+      });
+    }
+    const key = match[0].substr(1).trim();
+    const value = match[0] === curr ? true : curr.replace(match[0], "");
+    return Object.assign({}, prev, {
+      tags: Object.assign({}, prev.tags, {
+        [key]: !prev.tags.hasOwnProperty(key)
+          ? value
+          : prev.tags[key].constructor === Array
```
benjie: This should just be `Array.isArray(prev.tags[key])`; this has been supported since Node 0.x; e.g.

benjie:
I'm not gunna merge this right now because it's late and I'm tired; I'll merge it in the morning. Thanks!! 🙏 

(If I haven't merged it within 24 hours then give me a nudge please!)

packages/graphile-build-pg/src/plugins/PgColumnDeprecationPlugin.js
```diff
@@ -0,0 +1,29 @@
...
+        pgIntrospection: table,
+        pgFieldIntrospection,
+      },
+    } = context;
+    if (
+      !(isPgRowType || isPgCompoundType) ||
+      !table ||
+      table.kind !== "class" ||
+      !pgFieldIntrospection ||
+      typeof pgFieldIntrospection.tags !== "object"
```
benjie: I'll edit this in a minute because `typeof null === 'object'` and then line 23 would crash. I'll just replace with a truthy check on `pgFieldIntrospection.tags`

packages/graphile-build-pg/src/plugins/PgColumnDeprecationPlugin.js
```diff
@@ -0,0 +1,29 @@
...
+      !table ||
+      table.kind !== "class" ||
+      !pgFieldIntrospection ||
+      !pgFieldIntrospection.tags
+    ) {
+      return field;
+    }
+    return pgFieldIntrospection.tags.deprecated
+      ? Object.assign({}, field, {
+          deprecationReason: pgFieldIntrospection.tags.deprecated,
```
benjie: Feel free to follow up with a PR that changes this to:

benjie:
```
          deprecationReason: Array.isArray(pgFieldIntrospection.tags.deprecated) ? pgFieldIntrospection.tags.deprecated.join("\n") : pgFieldIntrospection.tags.deprecated,
```

(I can't do via GitHub because of prettier and don't have time to check it out.)

benjie:
Thanks! That's another task checked off the v4 release list; only 2 more to go!

PR #135, opened 2017-12-10
fix(getTypeByName): register default types like JSON earlier

### LINKED ISSUE #134: `getTypeByName("JSON")` doesn't always work in plugins
Because it's only defined as-needed. This works as a substitute:

```js
const JSONType = pgGetGqlTypeByTypeId("114");
```

e.g. the following code sample is broken by this:

https://www.graphile.org/postgraphile/extending/#adding-root-querymutation-fields

Potential solution I just hacked out - change PgTypesPlugin to define and register common types up front:

```js

    // pgExtendedTypes might change what types we use for things
    const JSONType = pgExtendedTypes ? GraphQLJSON : SimpleJSON;
    const UUIDType = SimpleUUID; // GraphQLUUID
    const DateType = SimpleDate; // GraphQLDate
    const DateTimeType = SimpleDatetime; // GraphQLDateTime
    const TimeType = SimpleTime; // GraphQLTime

    // Other plugins might want to use JSON
    addType(JSONType);
    addType(UUIDType);
    addType(DateType);
    addType(DateTimeType);
    addType(TimeType);

    const oidLookup = {
      "20": stringType(
        "BigInt",
        "A signed eight-byte integer. The upper big integer values are greater then the max value for a JavaScript number. Therefore all big integers will be output as strings and not numbers."
      ), // bitint - even though this is int8, it's too big for JS int, so cast to string.
      "21": GraphQLInt, // int2
      "23": GraphQLInt, // int4
      "700": GraphQLFloat, // float4
      "701": GraphQLFloat, // float8
      "1700": GraphQLString, // numeric
      "790": GraphQLFloat, // money

      "1186": GQLInterval, // interval
      "1082": DateType, // date
      "1114": DateTimeType, // timestamp
      "1184": DateTimeType, // timestamptz
      "1083": TimeType, // time
      "1266": TimeType, // timetz

      "114": JSONType, // json
      "3802": JSONType, // jsonb
      "2950": UUIDType, // uuid

      "1560": GraphQLString, // bit
      "1562": GraphQLString, // varbit
    };
```

Potentially better solution is to register default types with `getTypeByName` function, allow overriding these with plugins.

Hat tip @syndesis

benjie:
Fixes #134 

Not sure if this is the best solution, gunna think on it for a bit.

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
@@ -311,46 +311,47 @@ export default (function PgTypesPlugin(
       "UUID",
       "A universally unique identifier as defined by [RFC 4122](https://tools.ietf.org/html/rfc4122)."
     );
-    const oidLookup = Object.assign(
-      {
-        "20": stringType(
-          "BigInt",
-          "A signed eight-byte integer. The upper big integer values are greater then the max value for a JavaScript number. Therefore all big integers will be output as strings and not numbers."
-        ), // bitint - even though this is int8, it's too big for JS int, so cast to string.
-        "21": GraphQLInt, // int2
-        "23": GraphQLInt, // int4
-        "700": GraphQLFloat, // float4
-        "701": GraphQLFloat, // float8
-        "1700": GraphQLString, // numeric
-        "790": GraphQLFloat, // money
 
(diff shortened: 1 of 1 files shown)


benjie:
Yeah, I'm okay with this. It gives us a few more reserved words but I think the trade-off is worth it.

PR #139, opened 2017-12-12
feat(postgraphile-core): Add removePlugins option

mattbretl:


benjie:
Sorry it's taken a while to get back to you, I've been unwell.

I think rather than "removePlugins" we should call it "skipPlugins", and I think it should apply to *all* plugins - not just the default ones (otherwise we should call it `skipDefaultPlugins`). So something more like:

`(replaceAllPlugins ? [...] : [...]).filter(p => skipPlugins.indexOf(p) === -1)`

What do you think?

mattbretl:
Originally, I couldn't come up with a valid scenario for adding plugins (via appendPlugins/prependPlugins) and then subsequently stripping them out with removePlugins.  That's why I only filtered the defaults (though you're right, the option name should have been `removeDefaultPlugins`).

But thinking about this a bit more now.. I can foresee someone adding a plugin "pack" composed of many individual plugins, and then wanting to selectively exclude some of the individual plugins.  So I agree with changing the structure to:
```
(replaceAllPlugins ? [...] : [...]).filter(p => skipPlugins.indexOf(p) === -1)
```

I'm fine with the "remove" to "skip" change as well.

benjie:
I was also thinking it might be useful for debugging by selectively disabling plugins that you may have added elsewhere in your sourcecode 👍

mattbretl:
Also, I'd originally prevented the joint usage replaceAllPlugins and removePlugins, but just in case there's a valid use case, I'll remove that restriction too.

PR #172, opened 2018-02-21
fix(deps): Fix dependency patterns, graphql compatibility

benjie:
Updates to newer GraphQL library - now uses `"""` for comments rather than `#`.

Also fixes an issue where the system doesn't correctly skip the Mutation type if there's no mutations.


benjie:
The only reason this is passing is because Flow isn't running across the packages. There's hundreds of errors once running across packages is enabled.

benjie:
Fixing Flow is going to take too long and delay v4 too much. I'm going to merge without and we can go through and add flow to everything when we refactor later.

PR #177, opened 2018-02-22
feat: one-to-one relationships

benjie:
This'll add one-to-one relationships - i.e. where a remote table references the current table using a *unique* column.

~~This will be a breaking change (the last planned one before the release of v4), but~~ I'll be adding a flag to allow you to go enable the old behaviour instead. (Actually this should probably be a dual flag so that you can export the old interface as deprecated along with the new interface which isn't?)

This won't be a breaking change because the default for the flag will be to deprecate the old relations (but still support them).

Fixes postgraphql/postgraphql#432
Fixes postgraphql/postgraphql#283
Fixes postgraphql/postgraphql#123

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.js
@@ -284,7 +284,8 @@ export default (function PgBackwardRelationPlugin(
                     return addStartEndCursor(data[alias]);
                   },
                   deprecationReason: isDeprecated
-                    ? `Please use ${singleRelationFieldName} instead`
+                    ? // $FlowFixMe
+                      `Please use ${singleRelationFieldName} instead`
                     : undefined,
                 };
               },
--- a/packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.js
@@ -7,10 +7,19 @@ import type { Plugin } from "graphile-build";
 
 const debug = debugFactory("graphile-build-pg");
 
+const OMIT = 0;
+const DEPRECATED = 1;
+const ONLY = 2;
+
 export default (function PgBackwardRelationPlugin(
   builder,
-  { pgInflection: inflection }
+  { pgInflection: inflection, pgLegacyRelations }
 ) {
+  const legacyRelationMode =
+    {
+      only: ONLY,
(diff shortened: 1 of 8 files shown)


chadfurman:
I feel like this PR should include an update to a readme or some reference to documentation changes or a ticket for documentation changes reflecting what the ramifications are of this change.

You mention it is a breaking change: what is the old behavior that will be breaking, how will the breakage manifest in our use-cases, and what is the recommended path to fixing?  :)

benjie:
I've not added any code yet - just tests! These things will be coming. But basically:

Say you have:

```sql
create table foo (
  id serial primary key
);

create table bar (
  foo_id int not null primary key references foo,
  name text
);
```

currently you get something like:

```graphql
type Foo {
  id: Int!

  # This next field is annoying - you'll only ever get 0 or 1 records
  # giving a connection makes accessing it too indirect.
  barsByFooId: BarsConnection!
}

type Bar {
  fooId: Int!
  name: String
  fooByFooId: Foo
}
```

which you have to query like:

```graphql
{
  fooById(id: 1) {
    barsByFooId {
      edges {
        node {
          name
        }
      }
    }
  }
}
```

with this change you'd get instead:

```graphql
type Foo {
  id: Int!
  barByFooId: Bar
}

type Bar {
  fooId: Int!
  name: String
  fooByFooId: Foo
}
```


which you have to query like:

```graphql
{
  fooById(id: 1) {
    barByFooId {
      name
    }
  }
}
```

No more unnecessary indirection.

This *will* be the default; however I plan to add a CLI flag that you can use to either have the old schema back, or to have both behaviours supported (your choice) - the both behaviours schema would look something like:

```graphql
type Foo {
  id: Int!
  barsByFooId: BarsConnection! @deprecated(reason: "Please use barByFooId instead")
  barByFooId: Bar
}

type Bar {
  fooId: Int!
  name: String
  fooByFooId: Foo
}
```

benjie:
Relevant issues:

https://github.com/postgraphql/postgraphql/issues/432
https://github.com/postgraphql/postgraphql/issues/283
https://github.com/postgraphql/postgraphql/issues/123

benjie:
Changed my mind - the legacy relations will be deprecated in v4 and removed in v5 - I'll still add an option to push your schema one way or the other though.

benjie:
👍  Implemented without breaking changes; people are recommended to tidy up their schemas with `legacyRelations: 'omit'` though (which will be the default in the next version - v5).

PR #206, opened 2018-04-07
chore(deps): Remove babel-runtime

### LINKED ISSUE #205: Do we need babel-runtime any more?
I think since we've stated Node 8.6+ we don't need it and should remove it.

skvale:
I came into these files to see if `babel-runtime` was necessary and noticed that there were a few superfluous devDependencies in the yarn workspace packages.

> I think since we've stated Node 8.6+ we don't need it and should remove it.

(Fixes #205)


benjie:
Thanks! I’ll build this and give it a test before merging as it could be hiding a dependency on babel-runtime currently.

benjie:
As I suspected the compiled code still references babel-runtime; e.g.

```js
node8plus/plugins/PgConnectionArgOrderByDefaultValue.js:var _assign = require("babel-runtime/core-js/object/assign");
```

😞

benjie:
I think that's all nicely cleaned up now 👍

PR #212, opened 2018-04-19
fix: gracefully handle stale pg_catalog.pg_description

bradleyayers:
After upgrading a database using `pg_upgrade`, it's possible for stale entries in `pg_description` to exist. These refer to an old `pg_class` system catalog. These need to be filtered out by only including descriptions related to the current `pg_class` system catalog.

---

This is a fix I'm using after upgrading my Postgres 9.6 to Postgres 10.3 RDS instance via AWS (which uses pg_upgrade). It seems that pg_upgrade left behind a stale record in pg_description, which caused this error:

```
Error: Type 'NotesOrderBy' has already been registered! 
    at Object.newWithHooks (/app/node_modules/graphile-build/src/makeNewBuild.js:650:17) 
    at introspectionResultsByKind.class.filter.filter.forEach.table (/app/node_modules/graphile-build-pg/src/plugins/PgConnectionArgOrderBy.js:28:11) 
    at Array.forEach (<anonymous>) 
    at PgConnectionArgOrderBy.builder.hook (/app/node_modules/graphile-build-pg/src/plugins/PgConnectionArgOrderBy.js:22:10) 
    at SchemaBuilder.applyHooks (/app/node_modules/graphile-build/src/SchemaBuilder.js:292:20) 
    at SchemaBuilder.createBuild (/app/node_modules/graphile-build/src/SchemaBuilder.js:337:10) 
    at SchemaBuilder.buildSchema (/app/node_modules/graphile-build/src/SchemaBuilder.js:343:26) 
    at Object.exports.createPostGraphQLSchema (/app/node_modules/postgraphile-core/src/index.js:235:26) 
    at <anonymous> 
    at process._tickCallback (internal/process/next_tick.js:182:7) 
```

I have a `note` table, and what was happening is that the introspection query was returning two tuples for that table (because there were two items in `pg_catalog.pg_description` associated with that table). This patch narrows the join between `pg_catalog.pg_class` and `pg_catalog.pg_description` to honor the current `pg_class` oid.

This was a fairly serious bug for me, as it meant by node server would not start after upgrading to Postgres 10.3. Fortunately I did this on a staging environment first.

### DIFF
--- a/packages/graphile-build-pg/res/introspection-query.sql
+++ b/packages/graphile-build-pg/res/introspection-query.sql
@@ -102,7 +102,7 @@ with
       (pg_catalog.pg_relation_is_updatable(rel.oid, true)::bit(8) & B'00000100') = B'00000100' as "isDeletable"
     from
       pg_catalog.pg_class as rel
-      left join pg_catalog.pg_description as dsc on dsc.objoid = rel.oid and dsc.objsubid = 0
+      left join pg_catalog.pg_description as dsc on dsc.objoid = rel.oid and dsc.objsubid = 0 and dsc.classoid = (select oid from pg_catalog.pg_class where relname = 'pg_class')
       left join pg_catalog.pg_namespace as nsp on nsp.oid = rel.relnamespace
     where
       rel.relpersistence in ('p') and


benjie:
Awesome; thanks. I'll review next week 👍

(Btw do you have a link to PG release notes on this issue that I can link from the changelog?)

bradleyayers:
I didn't find any release note information about this. I came up with this solution from lots of poking around and the [docs for `pg_catalog.pg_description`](https://www.postgresql.org/docs/9.5/static/catalog-pg-description.html).

packages/graphile-build-pg/res/introspection-query.sql
```diff
@@ -102,7 +102,7 @@ with
       (pg_catalog.pg_relation_is_updatable(rel.oid, true)::bit(8) & B'00000100') = B'00000100' as "isDeletable"
     from
       pg_catalog.pg_class as rel
-      left join pg_catalog.pg_description as dsc on dsc.objoid = rel.oid and dsc.objsubid = 0
+      left join pg_catalog.pg_description as dsc on dsc.objoid = rel.oid and dsc.objsubid = 0 and dsc.classoid = (select oid from pg_catalog.pg_class where relname = 'pg_class')
```
benjie: I don't think this will work universally, for example:

benjie: Unfortunately this discards comments on things that aren't tables/columns, such as functions.

```
[store] # select oid, relname from pg_class where oid in (select distinct classoid from pg_description);
 oid  |    relname
------+----------------
 1247 | pg_type
 1255 | pg_proc
 1259 | pg_class
 2617 | pg_operator
 2601 | pg_am
 2612 | pg_language
 2615 | pg_namespace
 2607 | pg_conversion
 3602 | pg_ts_config
 3600 | pg_ts_dict
 3601 | pg_ts_parser
 3764 | pg_ts_template
 3079 | pg_extension
 3456 | pg_collation
(14 rows)

Comment (benjie) [inline packages/graphile-build-pg/res/introspection-query.sql]: Here's the comments/sources I get in one of my databases:

```
# select classoid, count(*), relname from pg_description inner join pg_class on (classoid = pg_class.oid) group by classoid, relname order by count desc;
 classoid | count |    relname
----------+-------+----------------
     1255 |  2878 | pg_proc
     2617 |   787 | pg_operator
     2607 |   132 | pg_conversion
     1247 |    73 | pg_type
     3602 |    16 | pg_ts_config
     3600 |    16 | pg_ts_dict
     2601 |     6 | pg_am
     3764 |     5 | pg_ts_template
     2612 |     4 | pg_language
     3079 |     4 | pg_extension
     2615 |     3 | pg_namespace
     3456 |     3 | pg_collation
     1259 |     2 | pg_class
     3601 |     1 | pg_ts_parser
(14 rows)
```

Seems legit to me that when we want comments on a class we'd only want ones from `pg_class`. I'd definitely be interested to see where your stray comments came from though?

benjie: Feels to me we should replicate this check on columns (also `pg_class`), procs (`pg_proc` instead of `pg_class`), etc.

benjie: Filed as issue: https://github.com/graphile/graphile-build/issues/216

bradleyayers: I have no idea actually. They came into existence after upgrading with `pg_upgrade`, so perhaps there's either a bug or a quirk in how that works. It doesn't technically seem "wrong" to have multiple rows though, the schema indeed supports it, but unexpected yes.

PR #218, opened 2018-05-05
fix(procs): handle returning arrays of custom types

benjie:
Fixes https://github.com/graphile/postgraphile/issues/752

Aside: no idea why this change was made: https://github.com/graphile/graphile-build/commit/abfd40d98c0aaf66f0230c0f991fb6d576e432c8


benjie:
I've also extended the tests to cover types that need pg2gql to be called (e.g. intervals)

PR #231, opened 2018-06-02
Update parseMoney to handle negative values

farant:
This is to fix a bug where the negative sign is lost for money types.


benjie:
CI isn’t happy about the backslash; remove it and we should be good to merge 👍

PR #233, opened 2018-06-16
feat(pg): introspection of subtypes, exclude extension resources

benjie:
We now auto-filter functions/tables from extensions; this is technically a breaking change but I don't think many people were using this deliberately. You can go back to the old functionality with the CLI flag `--include-extension-resources`

---

[pg_attribute](https://www.postgresql.org/docs/10/static/catalog-pg-attribute.html) contains a field `atttypmod` which is a modifier for the type of a column, for example your column could be a `decimal`; but it could be `decimal(3,2)`. This `atttypmod` is where this distinction is stored (or `-1` if there is no subtype).

So we really need this in our modelling. Annoyingly this means that I've had to rebuild all the code that assumes we can do things based on type alone.

I've managed to do all these changes whilst maintaining backwards compatibility, *except* if you use the (undocumented) build properties `pgGqlTypeByTypeId` or `pgGqlInputTypeByTypeId` in a custom plugin you're using. If this is the case then an error will be thrown which will make it very clear you need to fix it - so if you upgrade and there's no error then you're safe. I don't expect anyone at all to hit this issue, but in case someone does, the fix is simple - the object has become two dimensional so rename the property to `pgGqlTypeByTypeIdAndModifier` and insert a [-1] as the modifier; e.g. `pgGqlTypeByTypeId[1234] = Type` would become `pgGqlTypeByTypeIdAndModifier[1234] = {"-1": Type}`. **CHANGES TO UNDOCUMENTED PROPERTIES ARE NOT DEEMED TO BE BREAKING CHANGES; IF YOU WANT TO RELY ON AN UNDOCUMENTED PROPERTY THEN PLEASE SUBMIT A DOCUMENTATION PR!**

A few methods that you might be using are now deprecated (but will continue to work, albeit not supporting advanced types): `pgGetGqlTypeByTypeId`, `pgGetGqlInputTypeByTypeId` and `pgTweakFragmentForType`; please switch to using the `AndModifier` variants - if you're using it to find a table type you can just pass `null` as the modifier.

Here's an example of `atttypmod`:

```sql
drop table decimals;
create table decimals (
  decimal_1_1 decimal(1,1),
  decimal_2_1 decimal(2,1),
  decimal_3_2 decimal(3,2),
  decimal_4_3 decimal(4,3),
  decimal_4_4 decimal(4,4),
  decimal_10_4 decimal(10,4),
  decimal_60_45 decimal(60, 45),
  decimal_97_79 decimal(97, 79)
);
select
  attname,
  atttypid,
  atttypmod,
  atttypmod >> 16,
  (atttypmod & ((1 << 16) - 1)) - 4
from pg_attribute
where attname like 'decimal_%';
```

this generates the following table:

```
    attname    | atttypid | atttypmod | ?column? | ?column?
---------------+----------+-----------+----------+----------
 decimal_1_1   |     1700 |     65541 |        1 |        1
 decimal_2_1   |     1700 |    131077 |        2 |        1
 decimal_3_2   |     1700 |    196614 |        3 |        2
 decimal_4_3   |     1700 |    262151 |        4 |        3
 decimal_4_4   |     1700 |    262152 |        4 |        4
 decimal_10_4  |     1700 |    655368 |       10 |        4
 decimal_60_45 |     1700 |   3932209 |       60 |       45
 decimal_97_79 |     1700 |   6357075 |       97 |       79
(8 rows)
```

- [x] atttypmod
- [x] typtypmod
- [x] pg_extension (exclude tables from extension config)
- [x] exclude tables/functions from extensions automatically
- [ ] tests

### DIFF
--- a/packages/graphile-build-pg/res/introspection-query.sql
+++ b/packages/graphile-build-pg/res/introspection-query.sql
@@ -170,6 +170,7 @@ with
         (typ.typelem <> 0 and typ.typlen = -1) as "isPgArray",
         nullif(typ.typrelid, 0) as "classId",
         nullif(typ.typbasetype, 0) as "domainBaseTypeId",
+        nullif(typ.typtypmod, -1) as "domainTypeModifier",
         -- If this type is an enum type, let’s select all of its enum variants.
         --
         -- @see https://www.postgresql.org/docs/9.5/static/catalog-pg-enum.html
--- a/packages/graphile-build-pg/res/introspection-query.sql
+++ b/packages/graphile-build-pg/res/introspection-query.sql
@@ -242,6 +242,23 @@ with
       con.contype in ('f', 'p', 'u')
     order by
       con.conrelid, con.conkey, con.confrelid, con.confkey, con.conname
+  ),
+  -- @see https://www.postgresql.org/docs/9.5/static/catalog-pg-extension.html
+  "extension" as (
+    select
+      'extension' as "kind",
+      ext.oid as "id",
+      ext.extname as "name",
+      ext.extnamespace as "namespaceId",
+      ext.extrelocatable as "relocatable",
+      ext.extversion as "version",
+      ext.extconfig as "configurationClassIds",
+      dsc.description as "description"
+    from
+      pg_catalog.pg_extension as ext
+      left join pg_catalog.pg_description as dsc on dsc.objoid = ext.oid
+    order by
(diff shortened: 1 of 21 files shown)


benjie:
Maybe the option, rather than being boolean, should be a list of the extension names that you want to include the tables/methods from.

PR #247, opened 2018-07-11
Removed unnecessary freezing and cloning

jclangst:
IN PROGESS

- [x] Remove the unecessary freezing and cloning on the initial introspection
- [x] Separate the internal functions into isolated areas of concern (also bring them to the top level)
- [ ] Reevaluate the lookup by ID methodology
- [x] Add more verbose comments
...

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
@@ -69,10 +69,10 @@ export default (async function PgIntrospectionPlugin(
   );
 
   //rehydrate the objects by linking them by their ID fields
-  toRelate.forEach(({ kind, IdAttr, newAttr, missingOk }: relation) => {
+  toRelate.forEach(({ kind, idAttr, newAttr, missingOk }: relation) => {
     relate(
       introspectionResultsByKind[kind],
-      IdAttr,
+      idAttr,
       newAttr,
       introspectionResultsById,
       missingOk
@@ -298,7 +298,7 @@ function parseTagsFromStructures(pgStructuresByKind, smartComments) {
  * relate([types], 'classId', 'class', structsByID) will attach the corresponding class struct to each type.
  *
  * @param arrayOfStructs - An array of structs from the initial PG introspection
- * @param IdAttr - The ID attribute on each struct in 'arrayOfStructs' to  use in looking up another struct with
+ * @param idAttr - The ID attribute on each struct in 'arrayOfStructs' to  use in looking up another struct with
  * @param newAttr - The property name by which you want to attach the looked up struct to each struct in arrayOfStructs
  * @param lookupTable - An ES6 Map that can be used to look up each struct by its ID (see createStructuresByID)
  * @param missingOk - Should throw error if lookupTable does not contain an ID
@@ -306,7 +306,7 @@ function parseTagsFromStructures(pgStructuresByKind, smartComments) {
 
 function relate(
   arrayOfStructs,
-  IdAttr,
+  idAttr,
   newAttr,
   lookupTable,
   missingOk?: boolean = false
@@ -318,7 +318,7 @@ function relate(
       if (missingOk) return;
       else
         throw new Error(
-          `Could not look up '${newAttr}' by '${IdAttr}' on '${JSON.stringify(
+          `Could not look up '${newAttr}' by '${idAttr}' on '${JSON.stringify(
             struct
           )}'`
         );
@@ -328,7 +328,7 @@ function relate(
 
   //apply the lookup logic to each struct in the array (can handle both an array of ids and a single id)
   arrayOfStructs.forEach(struct => {
-    const id = struct[IdAttr];
+    const id = struct[idAttr];
     if (id)
       struct[newAttr] = Array.isArray(id)
         ? id.map(id => getRelatedStruct(id, struct))
@@ -343,7 +343,7 @@ function relate(
  */
 type relation = {
   kind: string,
-  IdAttr: string,
+  idAttr: string,
   newAttr: string,
   missingOk?: boolean,
 };
@@ -351,57 +351,57 @@ type relation = {
 const toRelate: Array<relation> = [
   {
     kind: "class",
-    IdAttr: "namespaceId",
+    idAttr: "namespaceId",
     newAttr: "namespace",
     missingOk: true,
   },
   {
     kind: "class",
-    IdAttr: "typeId",
+    idAttr: "typeId",
     newAttr: "type",
   },
   {
     kind: "attribute",
-    IdAttr: "classId",
+    idAttr: "classId",
     newAttr: "class",
   },
   {
     kind: "attribute",
-    IdAttr: "typeId",
+    idAttr: "typeId",
     newAttr: "type",
   },
   {
     kind: "procedure",
-    IdAttr: "namespaceId",
+    idAttr: "namespaceId",
     newAttr: "namespace",
   },
   {
     kind: "type",
-    IdAttr: "classId",
+    idAttr: "classId",
     newAttr: "class",
     missingOk: true,
   },
   {
     kind: "type",
-    IdAttr: "domainBaseTypeId",
+    idAttr: "domainBaseTypeId",
     newAttr: "domainBaseType",
     missingOk: true,
   },
   {
     kind: "type",
-    IdAttr: "arrayItemTypeId",
+    idAttr: "arrayItemTypeId",
     newAttr: "arrayItemType",
     missingOk: true,
   },
   {
     kind: "extension",
-    IdAttr: "namespaceId",
+    idAttr: "namespaceId",
     newAttr: "namespace",
     missingOk: true,
   },
   {
     kind: "extension",
-    IdAttr: "configurationClassesId",
+    idAttr: "configurationClassesId",
     newAttr: "configurationClasses",
     missingOk: true,
   },
@@ -446,7 +446,7 @@ function generateWatcher(pgConfig, schemasToWatch) {
   const watcher = async triggerRebuild => {
     withPgClient(pgConfig, async pgClient => {
       //activate the pg_notify channel with the schema watch query
-      await installPgWatch(pgClient);
+      const uninstall = await installPgWatch(pgClient);
       //because of the promise, client will remain active until resolve() is called
       return new Promise(resolve => {
         const listener = generateNotificationHandler(
@@ -460,9 +460,7 @@ function generateWatcher(pgConfig, schemasToWatch) {
           schemasToWatch
         );
         unwatcher = () => {
-          pgClient.query(`unlisten ${WATCH_CHANNEL}`).catch(e => {
-            debug(`Error occurred trying to unlisten watch: ${e}`);
-          });
+          uninstall();
           pgClient.removeListener("notification", listener);
           resolve();
         };
@@ -476,7 +474,8 @@ function generateWatcher(pgConfig, schemasToWatch) {
 }
 
 /**
- * Creates a pg_notify alert on the WATCH_CHANNEL to enable listening for schema changes
+ * Creates a pg_notify alert on the WATCH_CHANNEL to enable listening for schema changes; returns
+ * the thunk that will unlisten to the channel
  *
  * @param pgClient
  */
@@ -510,7 +509,10 @@ async function installPgWatch(pgClient) {
     await pgClient.query("rollback");
   }
   await pgClient.query(`listen ${WATCH_CHANNEL}`);
-  return pgClient;
+  return () =>
+    pgClient.query(`unlisten ${WATCH_CHANNEL}`).catch(e => {
+      debug(`Error occurred trying to unlisten watch: ${e}`);
+    });
 }
 
 /**
--- a/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
@@ -24,95 +24,185 @@ const WATCH_CHANNEL = "postgraphile_watch";
  * (1) Queries the designated PostgreSQL database for some relevant metadata
  * (2) Rehyradates the metadata by connecting related structures by their id fields (most of the metadata
  * links to each other (e.g., classes and attributes, classes and namespaces, etc.)
- * (3) Provides two access layers to the metadata on the build object:
- *  - pgIntrospectionResultsByKind: All the definitions organized by their kind
- *  - pgIntrospectionResultsById: All thee definitions indexed by their unique ids
+ * (3) Provides access to the introspection metadata on the build object via `pgIntrospectionResultsByKind`:
+ *
+ *  This object contains lists of the following PostgreSQL metadata structures
+ *
+ *    -namespace
+ *    -class
+ *    -procedure
+ *    -type
+ *    -attribute
+ *    -extension
+ *    -constraint
+ *
+ * Additionally, it has a fex indexed by PostgreSQL oid (a string):
+ *
+ *    -classById
+ *    -namespaceById
+ *    -typeById,
+ *    -extensionById
+ *
+ * Finally it has the attributes indexed by class and then identifying number on attributeByClassIdAndNum.
+ *
+ * See the flow types below for more information about what's contained on `pgIntrospectionResultsByKind`.
+ *
+ *
  * (4) Registers watchers with the builder in case you want to watch the PostgreSQL database for schema changes;
  * if a watcher see a DB change, it will trigger a build of the schema
  ****************************************************************************************/
 
-export default (async function PgIntrospectionPlugin(
-  builder,
-  options: PgIntrospectionPluginOptions
-) {
-  const { pgSchemas, pgConfig } = options;
+/****************************************************************************************
+ * Flow Types
+ ****************************************************************************************/
 
-  if (!Array.isArray(pgSchemas)) {
-    throw new Error("Argument 'schemas' (array) is required");
-  }
+export type PgNamespace = {
+  kind: "namespace",
+  id: string,
+  name: string,
+  description: ?string,
+  tags: { [string]: string },
+};
 
-  let introspectionResultsByKind: PgIntrospectionResultsByKind;
+export type PgProc = {
+  kind: "procedure",
+  name: string,
+  description: ?string,
+  namespaceId: string,
+  isStrict: boolean,
+  returnsSet: boolean,
+  isStable: boolean,
+  returnTypeId: string,
+  argTypeIds: Array<string>,
+  argNames: Array<string>,
+  argDefaultsNum: number,
+  namespace: PgNamespace,
+  tags: { [string]: string },
+  aclExecutable: boolean,
+};
 
-  //provide the logic for updating the introspection results asynchronously; needed
-  //for the watcher, but also called once during the initial setup
-  const updateIntrospectionResults = async () => {
-    introspectionResultsByKind = await getIntrospectionResultsByKind(options);
-  };
-  await updateIntrospectionResults();
+export type PgClass = {
+  kind: "class",
+  id: string,
+  name: string,
+  description: ?string,
+  classKind: string,
+  namespaceId: string,
+  namespaceName: string,
+  typeId: string,
+  isSelectable: boolean,
+  isInsertable: boolean,
+  isUpdatable: boolean,
+  isDeletable: boolean,
+  isExtensionConfigurationTable: boolean,
+  namespace: PgNamespace,
+  type: PgType,
+  tags: { [string]: string },
+  attributes: [PgAttribute],
+  aclSelectable: boolean,
+  aclInsertable: boolean,
+  aclUpdatable: boolean,
+  aclDeletable: boolean,
+};
 
-  //add the introspection results to the build object
-  builder.hook("build", build => {
-    return build.extend(build, {
-      pgIntrospectionResultsByKind: introspectionResultsByKind,
+export type PgType = {
+  kind: "type",
+  id: string,
+  name: string,
+  description: ?string,
+  namespaceId: string,
+  namespaceName: string,
+  type: string,
+  category: string,
+  domainIsNotNull: boolean,
+  arrayItemTypeId: ?string,
+  typeLength: ?number,
+  isPgArray: boolean,
+  classId: ?string,
+  domainBaseTypeId: ?string,
+  domainTypeModifier: ?number,
+  tags: { [string]: string },
+};
+
+export type PgAttribute = {
+  kind: "attribute",
+  classId: string,
+  num: number,
+  name: string,
+  description: ?string,
+  typeId: string,
+  typeModifier: number,
+  isNotNull: boolean,
+  hasDefault: boolean,
+  class: PgClass,
+  type: PgType,
+  namespace: PgNamespace,
+  tags: { [string]: string },
+  aclSelectable: boolean,
+  aclInsertable: boolean,
+  aclUpdatable: boolean,
+};
+
+export type PgConstraint = {
+  kind: "constraint",
+  name: string,
+  type: string,
+  classId: string,
+  foreignClassId: ?string,
+  description: ?string,
+  keyAttributeNums: Array<number>,
+  foreignKeyAttributeNums: Array<number>,
+  namespace: PgNamespace,
+  tags: { [string]: string },
+};
+
+export type PgExtension = {
+  kind: "extension",
+  id: string,
+  name: string,
+  namespaceId: string,
+  relocatable: boolean,
+  version: string,
+  configurationClassIds?: Array<string>,
+  description: ?string,
+  tags: { [string]: string },
+};
+
+export type PgIntrospectionResultsByKind = {
+  namespace: Array<PgNamespace>,
+  class: Array<PgClass>,
+  attribute: Array<PgAttribute>,
+  type: Array<PgType>,
+  constraint: Array<PgConstraint>,
+  procedure: Array<PgProc>,
+  extension: Array<PgExtension>,
+  namespaceById: { [string]: PgNamespace },
+  classById: { [string]: PgClass },
+  typeById: { [string]: PgType },
+  extensionById: { [string]: PgExtension },
+  attributeByClassIdAndNum: { [string]: { [number]: PgAttribute } },
+};
+
+export type PgIntrospectionPluginOptions = {
+  pgConfig: pg.Client | pg.Pool | string,
+  pgSchemas: Array<string>,
+  pgEnableTags: boolean,
+  persistentMemoizeWithKey?: <T>(string, () => T) => T,
+  pgThrowOnMissingSchema?: boolean,
+  pgIncludeExtensionResources?: boolean,
+};
+
+/****************************************************************************************
+ * Utilities Functions
+ ****************************************************************************************/
+
+function readFile(filename, encoding) {
+  return new Promise((resolve, reject) => {
+    rawReadFile(filename, encoding, (err, res) => {
+      if (err) reject(err);
+      else resolve(res);
     });
   });
-
-  //add the watchers to trigger rebuild if necessary
-  builder.registerWatcher(
-    ...generateWatcher(pgConfig, pgSchemas, updateIntrospectionResults)
-  );
-}: Plugin);
-
-/* The main workhorse of the plugin; accepts all of the plugin options - see
- * above for details
- */
-async function getIntrospectionResultsByKind({
-  pgConfig,
-  pgSchemas,
-  pgEnableTags,
-  persistentMemoizeWithKey = (key, fn) => fn(),
-  pgThrowOnMissingSchema = false,
-  pgIncludeExtensionResources = false,
-}: PgIntrospectionPluginOptions): Promise<PgIntrospectionResultsByKind> {
-  //query the PostgresQL database for the relevant meta data
-  const introspectionResultsByKind = await getRawPGStructuresByKind(
-    pgConfig,
-    pgSchemas,
-    pgIncludeExtensionResources,
-    persistentMemoizeWithKey
-  );
-
-  //verify that all of the schemas are present
-  verifySchemas(
-    introspectionResultsByKind.namespace,
-    pgSchemas,
-    pgThrowOnMissingSchema
-  );
-
-  // parse the smart comments
-  addTagsToStructures(introspectionResultsByKind, pgEnableTags);
-
-  //index the structures by ID
-  addlookupTableById(introspectionResultsByKind);
-
-  //rehydrate the objects by linking them by their ID fields
-  toRelate.forEach(
-    ({ kind, key, lookupTable, newAttr, missingOk }: relation) => {
-      relate(
-        introspectionResultsByKind[kind],
-        key,
-        newAttr,
-        introspectionResultsByKind[lookupTable],
-        missingOk
-      );
-    }
-  );
-  attachClassAttributes(
-    introspectionResultsByKind.classById,
-    introspectionResultsByKind.attribute
-  );
-
-  return introspectionResultsByKind;
 }
 
 /****************************************************************************************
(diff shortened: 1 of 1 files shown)


benjie:
What’s the motivation for this?

jclangst:
Hi benjie!

I'm really enjoying this library, and I was just looking through some of the source code to get a better understanding of how the plugin system works especially for the pg introspection and subsequent gql schema building.

As I was going through it, I noticed a few internal things that might be slightly refactored or documented better in order to help fresh eyes get a better grasp on what was happening (and maybe even increase performance slightly here and there). While I don't want to change the external api, I'm hoping my contributions may help others who want to contribute get up to speed faster. Would this be OK with you?

Also, this is very much not a final pull request yet. There will be a lot more explanation about why I'm suggesting the changes I am, but I just wanted to give whoever was interested a rough sketch of what I am doing in the interim.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -150,97 +150,102 @@ export default (async function PgIntrospectionPlugin(
...
+
+  //TODO why have all these nested functions
   async function introspect() {
-    // Perform introspection
+
     if (!Array.isArray(schemas)) {
       throw new Error("Argument 'schemas' (array) is required");
     }
+
+    //TODO find out what the cacheing mechanism is doing
```
benjie: Plugins can cache to speed up startup (e.g. no need to re-introspect the DB); useful for serverless. See https://www.graphile.org/postgraphile/v4-new-features/#cache-invalidation-serverless

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -150,97 +150,102 @@ export default (async function PgIntrospectionPlugin(
...
+
+        //load and run the query for the structure of the database
+        const introspectionQuery = await readFile(INTROSPECTION_PATH, "utf8");
+        const { rows } = await pgClient.query(introspectionQuery, [
+          schemas,
+          pgIncludeExtensionResources,
+        ]);
+
+        //sort the structures by kind
+        const pgStructuresByKind = rows.reduce(
```
benjie: 👍 definitely approve of renaming internal variables to be more readable!

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -150,97 +150,102 @@ export default (async function PgIntrospectionPlugin(
...
+            class: [],
+            attribute: [],
+            type: [],
+            constraint: [],
+            procedure: [],
+            extension: [],
+          }
+        );
+
+        // Parse tags from comments (smart commenting system)
```
benjie: Add a link to https://www.graphile.org/postgraphile/smart-comments/ ?

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -150,97 +150,102 @@ export default (async function PgIntrospectionPlugin(
...
     if (missingSchemas.length) {
-      const errorMessage = `You requested to use schema '${schemas.join(
-        "', '"
-      )}'; however we couldn't find some of those! Missing schemas are: '${missingSchemas.join(
-        "', '"
-      )}'`;
+      const errorMessage = `
+        You requested to use schema '${schemas.join("', '")}';
+        however we couldn't find some of those! Missing schemas are: '${missingSchemas.join("', '")}'
+      `;
```
benjie: Please use `eslint --fix`; it formats the code for you (via prettier) to take code layout opinions out of the equation as much as possible. There are edge cases like this, but they are better addressed by extracting template expressions out to variables instead rather than manual formatting.

jclangst: Absolutely. The final commit in the PR will definitely have everything eslinted (and I will start doing it more in the intermediate ones as well). I was just moving some stuff around for my own visualizations.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -252,6 +257,9 @@ export default (async function PgIntrospectionPlugin(
         memo[x[attrKey]][x[attrKey2]] = x;
         return memo;
       }, {});
+
+    //TODO seems like a confusing namespace convention to put id lookups under kind lookups
```
benjie: A "namespace" is postgres' internal name for a schema; it comes from the same introspection query as everything else, just kind === namespace.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -252,6 +257,9 @@ export default (async function PgIntrospectionPlugin(
...
+    //TODO maybe these should be maps instead of POJO
```
benjie: I don't see the advantage of moving to Maps here; they are typically slower and IDs are always strings so POJO works well. Also this would be a breaking change, plugins depend on this structure. (One more thing is that this is effectively a cacheable JSON structure currently (which might come in handy); using maps would complicate.)

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -264,6 +272,7 @@ export default (async function PgIntrospectionPlugin(
       introspectionResultsByKind.type,
       "id"
     );
+    //TODO maybe just add these straight to the class objects
```
benjie: https://github.com/graphile/graphile-build/blob/70f878364abb6e4d42c62026bf3a6fe7219c022b/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js#L390-L394 We do this here, now. We can't remove this though. Also the code in this link should (but doesn't) sort by attr.num.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -274,9 +283,22 @@ export default (async function PgIntrospectionPlugin(
...
+    /**
+     *
+     *
+     * @param array
+     * @param newAttr
+     * @param lookupAttr
+     * @param lookup
+     * @param missingOk
+     */
     const relate = (array, newAttr, lookupAttr, lookup, missingOk = false) => {
```
benjie: This goes through the objects in `array` and adds a new property `newAttr` to each object, where the value is the result of looking for the content(s) of the objects property `lookupAttr` in the `lookup` hash. Note this works for both scalar and array values of `lookupAttr`. Basically it rehydrates the relations.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
benjie: I recommend just having your editor do it for you every time you save the file. Its magic!

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
jclangst: I was actually referencing the fact that all of the "XbyId" indexes are under the pgIntrospectionResultsByKind object. I thought it might be better to have them under a "pgIntrospectionResultsByID" object. I'll add more on my thoughts below.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
jclangst: I agree completely. No maps!

jclangst:
I am wondering if running the introspection process right before triggering a rebuild is the best? From tracing the code, it looks like triggerRebuild() is going to run the introspection process again anyways, rigth?

https://github.com/jclangst/graphile-build/blob/b8e848953bb8aad3351ac5a24676f58c733a6ed0/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js#L594-L607

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -139,287 +138,377 @@ function readFile(filename, encoding) {
...
-      if (pgThrowOnMissingSchema) {
-        throw new Error(errorMessage);
+/**
+ * Takes the result of getPGStructuresByKind and extracts the tags based on the
+ * smart commenting system exposed by the 'parseTags' function from utils
+ *
+ * @param pgStructuresByKind
+ * @param smartComments - True iff smart comments are enabled
+ */
+function parseTagsFromStructures(pgStructuresByKind, smartComments) {
```
benjie: A more explicitly "mutating" name would be good (otherwise you'd expeect it to return); perhaps `addTagsToStructures` or `parseTagsIntoStructures` or `processTagsForStructures`?

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -139,287 +138,377 @@ function readFile(filename, encoding) {
...
+  // ["namespaceById", "classById", "typeById", "extensionById"].forEach(prop => {
+  //     console.log(prop);
+  //    pgStructuresByKind[prop] = new Proxy(pgStructuresByKind[prop], {
+  //        get: function(target, property, rec){
+  //            console.warn(`@DeprecationWarning: Accessing postgres data by ID via pgIntrospectionResultsByKind will
+  //            soon be deprecated. Please use pgIntrospectionResultsByID instead.`);
+  //            return target[property];
+  //        }
+  //    })
+  // });
```
benjie: Lets skip this for now, I'm happy with the current API and would rather not maintain two interfaces.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -139,287 +138,377 @@ function readFile(filename, encoding) {
...
+ * @param arrayOfStructs - An array of structs from the initial PG introspection
+ * @param IdAttr - The ID attribute on each struct in 'arrayOfStructs' to  use in looking up another struct with
+ * @param newAttr - The property name by which you want to attach the looked up struct to each struct in arrayOfStructs
+ * @param lookupTable - An ES6 Map that can be used to look up each struct by its ID (see createStructuresByID)
+ * @param missingOk - Should throw error if lookupTable does not contain an ID
+ */
+
+function relate(
+  arrayOfStructs,
+  IdAttr,
```
benjie: Why capital `I`?

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -139,287 +138,377 @@ function readFile(filename, encoding) {
...
+ *
+ * @param pgStructuresByKind
+ */
+function createStructuresById(pgStructuresByKind) {
+  // put every struct that has an ID field into a single map
+  const pgStructureById = {};
+  Object.keys(pgStructuresByKind).forEach(kind => {
+    pgStructuresByKind[kind].forEach(struct => {
+      if (struct.id) {
+        pgStructureById[struct.id] = struct;
```
benjie: I cannot find any evidence that object IDs are guaranteed to be unique across the system catalog tables.

The docs state: 

> There are also several alias types for oid: regproc, regprocedure, regoper, regoperator, regclass, regtype, regrole, regnamespace, regconfig, and regdictionary.

and

> The oid type is currently implemented as an unsigned four-byte integer. Therefore, it is not large enough to provide database-wide uniqueness in large databases, or even in large individual tables. So, using a user-created table's OID column as a primary key is discouraged. OIDs are best used only for references to system tables.

> https://www.postgresql.org/docs/10/static/datatype-oid.html

Further pg_depend explicitly states the catalog that the object identifiers it references belong to:

https://www.postgresql.org/docs/10/static/catalog-pg-depend.html

Further, there's no pg table or function that I can find that will tell you what the relevant catalog o find an object in given its object ID.

All of this leads me to believe that you cannot _rely_ on object IDs being globally unique across the system tables, and thus you must factor their type into the lookup (as the existing API does).

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -139,287 +138,377 @@ function readFile(filename, encoding) {
...
-
-    relate(
-      introspectionResultsByKind.attribute,
-      "type",
-      "typeId",
-      introspectionResultsByKind.typeById
+      },
+      {}
     );
+  });
```
benjie: I prefer explicit over implicit; if someone searches the codebase for "namespaceById" I want them to find where it's defined. I don't believe the small reduction in lines of code here justifies the reduction in searchability and explicitness.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -448,6 +537,8 @@ export default (async function PgIntrospectionPlugin(
       pgClient.on("error", e => {
         debug("pgClient error occurred: %s", e);
       });
+
+      //TODO why the if(pgClient) checks here and not above
```
benjie: I guess because above they are guaranteed to exist; here runs after a period of time, which might be after this is called: https://github.com/jclangst/graphile-build/blob/9bc063ba1323b8ac1a40276ffac37b5c0098fd8d/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js#L521-L524

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -500,7 +592,8 @@ export default (async function PgIntrospectionPlugin(
     const handleChange = throttle(
       async () => {
         debug(`Schema change detected: re-inspecting schema...`);
-        introspectionResultsByKind = await introspect();
+        //TODO pretty sure that introspection here is unecessary if we are just calling rebuild immediately after
+        //introspectionResultsByKind = await introspect();
```
benjie: But... this function handles changes to the schema. If it doesn't re-introspect then the schema will be rebuilt without any changes!

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -542,12 +635,7 @@ export default (async function PgIntrospectionPlugin(
       }
     };
     pgClient.on("notification", listener);
-    introspectionResultsByKind = await introspect();
+    //TODO not sure why we would be doing introspection here (this won't change what's already attached to build)
+    //introspectionResultsByKind = await introspect();
```
benjie: After you trigger a watch in graphile, all the watch registration functions are called and then a new schema is built. From then onwards the schema will be up to date with the watch fixtures. We must introspect here otherwise we'll miss the following situation:

- [user-code] build initial schema
- [db] edit a table
- [user-code] enable watch mode

^ the "edit a table" results won't be represented.

Note in graphile you can watch/unwatch a schema at any time:  https://www.graphile.org/graphile-build/schema-builder/#watchschemalistener

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
jclangst: That's a great catch, and I completely agree. Though now that gets me wondering, since it mentions 

>  only for references to system tables

then that would mean that uniqueness is more or less only guaranteed for references to tables on this list https://www.postgresql.org/docs/current/static/catalogs.html.

Given that, for large databases, it seems possible that there could be oid collisions both globally and within certain specific kinds like PostgreSQL ``types`` for example. That would mean even organizing by id as it is currently set up might be error prone as well?

However, upon looking at the docs for say ``types`` (https://www.postgresql.org/docs/10/static/catalog-pg-type.html), it looks many of the foreign keys are oids that don't belong to system tables. For example, it links to procedures by oid. The only practical way to do this would be for oid uniqueness to be guaranteed within a particular system catalog, right? Even if that isn't specified in the oid datatype docs?

That would lead me to believe that the old way was probably the correct way. Though a possible enhancement might be creating a unique id that is ```oid + calalog_name```.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
jclangst: Right. I'll rename it to ``idAttr``.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
jclangst: Absolutely. I got a little ahead of myself. I think looking up the introspection results by a globally unique identifier would simplify the api a bit, but given the comments about the oid uniqueness, this should be put on the backburner. I'll remove my changes here from the PR as soon as we conclude on the best way to handle oid logic.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
jclangst: Good call.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -139,287 +138,377 @@ function readFile(filename, encoding) {
...
+ * type - type information; can be scalars and enums, but are also created for each table in the data base; from pg_type table
+ * constraint - information about the various constraints on the table; from pg_constraints
+ * procedure - information about the aggregate functions as well as custom procedures; from pg_proc
+ * extensions - information about the installed extensions; from pg_extension
+ *
+ * Returns the structures in an collections of arrays indexed by the above kinds
+ *
+ * @param pgConfig - The configuration parameters for connecting to the PostgresQL DB
+ * @param schemas - The requested schemas to pull from the database
+ * @param includeExtensionResources - True iff should include extension resources
```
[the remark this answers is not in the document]
jclangst: Done!

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
benjie: The “kind” is effectively the catalog name, yes 👍

jclangst: Right, so even though it doesn't explicitly say it, the operating assumption is that oid is unique by catalog name (a.k.a. "kind")? Just wanted to double check that was a safe assumption.

jclangst:
@benjie OK, I just want to 100% clarify how the watching mechanism is working. Based on what I am seeing here:
```javascript
  async watchSchema(listener: SchemaListener) {
    if (this._watching || this._busy) {
      throw new Error("We're already watching this schema!");
    }
    try {
      this._busy = true;
      this._watching = true;
      this._explicitSchemaListener = listener;
      this.triggerChange = () => {
        this._generatedSchema = null;
        // XXX: optionally debounce
        this.emit("schema", this.buildSchema());
      };
      if (listener) {
        this.on("schema", listener);
      }
      for (const fn of this.watchers) {
        await fn(this.triggerChange);
      }
      this.emit("schema", this.buildSchema());
    } finally {
      this._busy = false;
    }
  }
```

It looks like both the call to ``watchSchema`` and ``triggerChange`` (aka ``triggerRebuild`` in the watchers themselves) both initiate a call to ``buildSchema``. After tracing where that call goes, it appears it ends up doing the following:
1) Completely scrapping the old schema (aka ``_generatedSchema``)
2) Creating an entirely new build object via ``createBuild``. 
3) Step 2 causes all of the inflection, build, and init hooks being run again. This would include the hooks installed by ``PgIntrospectionPlugin``.
4) As a result, at this point, database introspection would occur again.

Based on that, I think that doing ``introspect`` immediately before ``triggerRebuild`` or really in the watcher code at all would be wasteful as the result would be immediately discarded. Am I getting that right?

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
jclangst: I might be missing something, but I'm fairly sure that that case is covered because we have the ``this.emit("schema", this.buildSchema())`` at the very end of ``watchSchema`` of ``SchemaBuilder``. The already guarantees a schema rebuild upon calling ``watchSchema`` (unless an error is thrown in the try block).

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
jclangst: I'm thinking that because it looks like a lot of thought went into getting the ``pgClient`` based on the polymorphism of ``pgConfig`` in the ``withPgClient`` utility, we might want to stick to using that. That way if there ever needs to be changes in how pgConfig is handled, it will remain consistent across all plugins.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
jclangst: I think you are absolutely right. I am going to change it back.

benjie:
Your misunderstanding in the above is that step 2 is *synchronous* (note the lack of `await`). Plugins themselves are asynchronous; but once they return the hooks they have registered *must* run synchronously. As such if you want to update the data that those hooks use you do that asynchronously, and then once you've done that you trigger the schema to rebuild (synchronously, using the new data).

Hooks being synchronous is very important.

A more full example would be:

1. Load all the plugins, one at a time, each one doing optional asynchronous work and then registering synchronous hooks
2. *Synchronously* generate a GraphQL schema from the synchronous hooks
3. <wait a while, if you like>
4. Enable watch mode: a) execute the asynchronous watcher for every plugin that registered one, one at a time in series; b) synchronously generate a new schema
5. <nothing happens for a while>
6. A plugin detects a change; a) first it does its asynchronous work (e.g. re-introspecting the database schema); b) then it triggers a synchronous schema rebuild


I hope this makes a bit more sense!

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
benjie: That would have to be done *very* carefully to ensure that no functionality changed here. PostGraphile is not the only user of graphile-build-pg. `withPgClient` was generally intended for short-lived clients (a few statements and then release) whereas this is a long-lived client that listens for NOTIFY events.

jclangst:
Thank you! That makes complete sense. I see what you were saying now, and I know the changes to make.

jclangst:
I'll definitely take a look!

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -434,20 +426,20 @@ const toRelate: Array<relation> = [
...
  * @param attributeStructs
- * @param lookupTable
  */
 
-function attachClassAttributes(classStructs, attributeStructs, lookupTable) {
+function attachClassAttributes(classStructsById, attributeStructs) {
   //initialize all of the class structs to have no attributes
-  classStructs.forEach(struct => {
+  Object.values(classStructsById).forEach(struct => {
+    // $FlowFixMe
```
jclangst: Flow fought with me pretty hard here. Think it's related to https://github.com/facebook/flow/issues/2221.

jclangst:
@benjie I was trying to add flow types to the plugin options as a final bit of work on this but I was running into some problems with ``persistentMemoizeWithKey``. It's second argument should return a Promise, correct? However, flow throws errors when I try to mark it as such. Not sure why, so I am just going to leave it as is but let me know what you think.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -3,20 +3,581 @@ import type { Plugin } from "graphile-build";
...
 import { readFile as rawReadFile } from "fs";
-import pg from "pg";
 import debugFactory from "debug";
 import chalk from "chalk";
 import throttle from "lodash/throttle";
 import flatMap from "lodash/flatMap";
-import { quacksLikePgPool } from "../withPgClient";
-
 import { version } from "../../package.json";
+import pg from "pg/lib/index";
```
benjie: Why not just `pg`? This seems brittle; it's generally a bad idea to require from the internals of a package unless they've deemed doing so supported.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -3,20 +3,581 @@ import type { Plugin } from "graphile-build";
...
+ * -------------------------------------------------------------------------------------
+ *
+ * Performs the following operations:
+ *
+ * (1) Queries the designated PostgreSQL database for some relevant metadata
+ * (2) Rehyradates the metadata by connecting related structures by their id fields (most of the metadata
+ * links to each other (e.g., classes and attributes, classes and namespaces, etc.)
+ * (3) Provides two access layers to the metadata on the build object:
+ *  - pgIntrospectionResultsByKind: All the definitions organized by their kind
+ *  - pgIntrospectionResultsById: All thee definitions indexed by their unique ids
```
benjie: Remove `pgIntrospectionResultsById` line

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
jclangst: This was a mistaken auto import by my IDE. I'll fix it.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
jclangst: Yep.

jclangst:
I completely understand. I restored the ordering of the pieces of functionality with flow types at the top and the plugin definition at the bottom. I am afraid it will probably still be hard to view the exact diffs on github because a lot of the changes involved moving the various steps of the plugin into singe purpose functions.

I was kind of partial to putting the plugin definition at the top rather than the bottom because that enables users to quickly see the ~20 line pseudocode rather than scrolling down 600 lines. However, I want to follow whatever standards are already in place.

benjie:
I think we can revisit reordering in a separate PR. Should probably have broken this up into separate phases, but I think there’s been a fair bit of discovery along the way! I’ll have another go at reviewing soon, thanks for your work on this 🙏

jclangst:
I agree with the phases. I was planning to go ahead and do similar documentation and internal renaming for the rest of the plugins, so if you have any suggestions for what PR flow would work best for you, just let me know! Fortunately, it looks like this one and PgTypesPlugin are going to be the only major pain points for reviewing on GH as they both contain a lot of functionality.

benjie:
Yeah; reviewing is super time consuming when it's a large diff! Big refactors should be split into separate tasks as small as possible and each should be their own PR (one at a time, wait for it to be merged before doing the next bit of work). E.g. adding comments would be it's own PR; when that's merged renaming some variables could be a PR; when that's merged moving a few functions around could be a PR; etc. I'm generally pretty quick on the issues so if your PR is very simple it can be reviewed without too much latency. Larger reviews require me to dedicate blocks of time which makes the cycle much much slower (days-weeks rather than hours-days).

jclangst:
Moving forward I will definitely do that. I'm afraid for PgTypesPlugin, it is going to be a monolithic diff (in the sense that things are going to be moving about the file) just because it contains so much disparate functionality each with each of the component parts kindof scattered about. To get all of the related parts together in order to add the appropriate documentation, I am afraid it required some changes that are going to drive GH's diff mechanism crazy.

Fortunately, none of these changes are doing anything but improving the documentation for future contributors, so if it takes a while to get to review them, that's no issue with me at least. Additionally, the subsequent plugins seem very reasonably self contained and organized and I anticipate will more easily lend themselves to incremental PRs.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -3,20 +3,58 @@ import type { Plugin } from "graphile-build";
...
+ *
+ *    -namespace
+ *    -class
+ *    -procedure
+ *    -type
+ *    -attribute
+ *    -extension
+ *    -constraint
+ *
+ * Additionally, it has a fex indexed by PostgreSQL oid (a string):
```
benjie: What is this term "fex"?

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -3,20 +3,58 @@ import type { Plugin } from "graphile-build";
...
+ *
+ *  This object contains lists of the following PostgreSQL metadata structures
+ *
+ *    -namespace
+ *    -class
+ *    -procedure
+ *    -type
+ *    -attribute
+ *    -extension
+ *    -constraint
```
benjie: Please add spaces after the hyphens 🙃

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
benjie: This is not yet fixed

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -3,20 +3,58 @@ import type { Plugin } from "graphile-build";
...
+ *    -extensionById
+ *
+ * Finally it has the attributes indexed by class and then identifying number on attributeByClassIdAndNum.
+ *
+ * See the flow types below for more information about what's contained on `pgIntrospectionResultsByKind`.
+ *
+ *
+ * (4) Registers watchers with the builder in case you want to watch the PostgreSQL database for schema changes;
+ * if a watcher see a DB change, it will trigger a build of the schema
+ ****************************************************************************************/
```
benjie: 👌

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -130,6 +168,34 @@ export type PgExtension = {
...
+  pgEnableTags: boolean,
+  persistentMemoizeWithKey?: <T>(string, () => T) => T,
+  pgThrowOnMissingSchema?: boolean,
+  pgIncludeExtensionResources?: boolean,
+};
+
+/****************************************************************************************
+ * Utilities Functions
+ ****************************************************************************************/
+
```
benjie: 👌

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -139,393 +205,550 @@ function readFile(filename, encoding) {
...
-    pgConfig,
-    pgSchemas: schemas,
-    pgEnableTags,
-    persistentMemoizeWithKey = (key, fn) => fn(),
-    pgThrowOnMissingSchema = false,
-    pgIncludeExtensionResources = false,
-  }
+/****************************************************************************************
+ * Build Extensions - Add two ways to query the PostgreSQL introspection results to the
+ * build object; one by 'kind' and the other by 'id'. See below for details
```
benjie: I think this text is out of date; this PR should introduce no API changes since it is only a refactor, and thus it should (and does?) only expose the `pgIntrospectionResultsByKind`

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -139,393 +205,550 @@ function readFile(filename, encoding) {
...
+/**
+ * Query the postgres server for the following types of structures in the database:
+ *
+ * namespace - the schemas; from pg_namespace
+ * class - anything `relation-like` (e.g., tables, indices, etc.); from pg_class
+ * attribute - information about table columns; from pg_attribute
+ * type - type information; can be scalars and enums, but are also created for each table in the data base; from pg_type table
+ * constraint - information about the various constraints on the table; from pg_constraints
+ * procedure - information about the aggregate functions as well as custom procedures; from pg_proc
+ * extensions - information about the installed extensions; from pg_extension
```
benjie: Would be nice to have a link to the relevant section of the PG manual here: https://www.postgresql.org/docs/10/static/catalogs.html

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -139,393 +205,550 @@ function readFile(filename, encoding) {
...
+/****************************************************************************************
+ * Build Extensions - Add two ways to query the PostgreSQL introspection results to the
+ * build object; one by 'kind' and the other by 'id'. See below for details
+ ****************************************************************************************/
+
+/**
+ * Query the postgres server for the following types of structures in the database:
+ *
+ * namespace - the schemas; from pg_namespace
+ * class - anything `relation-like` (e.g., tables, indices, etc.); from pg_class
```
benjie: Might be worth adding "composite types" explicitly to this list.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -139,393 +205,550 @@ function readFile(filename, encoding) {
...
+ * attribute - information about table columns; from pg_attribute
```
benjie: "table and composite type columns" perhaps

benjie:
I really appreciate the work you've put into this PR, but it's just too big and too many changes for me to effectively review. Are you able to split it up into focussed PRs that can each be reviewed individually?

For example the following facets of this PR could each be separate PRs in a series:

- The new flow types for PgIntrospectionResultsByKind (these would be merged very easily because no code changes take place)
- Renaming variables like `pgIncludeExtensionResources` -> `includeExtensionResources`; `result` -> `pgStructuresByKind`; `pgThrowOnMissingSchema` -> `throwIfMissing` (these kind of simple changes could be merged very easily because they don't change the structure of the code)
- The helpful comments (these would be merged very easily because again they don't change any code)
- refactoring repeated code into array iteration (this will require discussion, the result of this should be kept as close to the original code as possible so as to not distract from the intention)
- separating the code into helper functions with minimal code editing inside the functions themselves (this will require discussion)

At the moment this pull request is insurmountable for me; I cannot be certain that no bugs have been introduced and no functionality subtly altered without effectively having to re-implement all these changes again myself and trying to get the diffs to line up, and I simply don't have time to do that right now.

jclangst:
Absolutely no worries. I completely understand the benefits of doing it
incrementally. When I first jumped in, I just couldn't get a handle on what
was going on without moving things around significantly. I'll create an
issue to track the progress and create separate smaller PR's for your
approval.

benjie:
That’s awesome @jclangst, thanks for understanding!

benjie:
Unfortunately this PR is now massively out of date. I'm planning to move the relevant files to TypeScript in the coming months too, so I'm going to close this for now.

[semi-automated message] Hi, there has been no activity in this for over 8 weeks so I'm closing it to keep the issues/pull request manageable. If this is still an issue, please re-open with a comment explaining why.

PR #249, opened 2018-07-13
PgTypesPlugin update

jclangst:
I apologize in advance for the fact that this is not an eye-ballable PR. It's not a complete rewrite, but it is a complete reorganization, meaning that GH's diff engine isn't going to do you any favors. That being said, I think the plugin is much more grok-able now. 

The are three primary motivations in this PR:
1) add documentation and more intuitive naming conventions where possible
2) separate and mark the individual areas of concern / functionality
3) clean up / optimize where possible (mostly just removing duplicate checks, eliminating recaching / reassignment of the same data, unnecessary indirection, getting the stack traces without throwing errors, etc. -- all of the actual logic is maintained)

I have also put some comments at the top and todos throughout to indicate what questions I had. If I wasn't completely sure what the effects of making a particular change would be, I left it alone and simply added a comment.

I am certainly not expecting a quick turnaround on this, but wanted to put it out there for you to look at. I will be making smaller minor edits to the documentation soon, but this is probably 80-90% done.

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
@@ -549,23 +549,58 @@ export default (function PgTypesPlugin(
      *****************************************************************************************/
 
     // TODO: add more support for geometric types
-
-    let depth = 0;
-
     /*
      * Enforce: this is the fallback when we can't find a specific GraphQL type
      * for a specific PG type.  Use the generators from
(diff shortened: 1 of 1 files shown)


benjie:
[semi-automated message] Hi, there has been no activity in this issue for over 8 weeks so I'm closing it to keep the issues/pull request manageable. If this is still an issue, please re-open with a comment explaining why.

PR #269, opened 2018-08-14
feat(makeExtendSchemaPlugin): add support for enum

bradleyayers:
I want support for enum types, so that I can extend the GraphQL schema to include enums. I'm fairly unfamiliar with this code base, so any guidance here would be appreciated.


benjie:
#271 contains fixes and enhancements to this PR, I think it's good to merge. If you pull it into this branch then you'll still get credit 👍

PR #280, opened 2018-08-16
feat(types): beginnings of TypeScriptification

benjie:
This is an undertaking... 😅

- [x] Use pg-sql2@2.2.1 (converted to TypeScript)
- [x] Convert graphql-parse-resolve-info to TypeScript
- [x] Convert postgraphile-core to TypeScript
- [x] Add typings to graphile-build
- [x] Add typings to graphile-build-pg
- [x] Ensure builds still work
- [x] Ensure tests still pass
- [x] Ensure linting is in place
  - [x] pretter
  - [x] tslint

Fixes https://github.com/graphile/graphile-build/issues/277


packages/graphile-build-pg/package.json
```diff
@@ -3,10 +3,11 @@
   "version": "4.0.0-rc.5",
   "description": "Build a GraphQL schema by reflection over a PostgreSQL schema. Easy to customize since it's built with plugins on graphile-build",
   "main": "node8plus/index.js",
+  "types": "node8plus/index.d.ts",
   "scripts": {
     "test": "jest",
-    "prepack": "mkdir -p node8plus && babel -s true --out-dir node8plus src && flow-copy-source src node8plus",
-    "watch": "mkdir -p node8plus && babel -s true --watch --out-dir node8plus src"
+    "prepack": "mkdir -p node8plus && babel --copy-files -s true --out-dir node8plus src && flow-copy-source src node8plus",
+    "watch": "mkdir -p node8plus && babel --copy-files -s true --watch --out-dir node8plus src"
```
benjie: `--copy-files` is needed to copy the `.d.ts` files.

package.json
```diff
@@ -23,9 +23,9 @@
     "flow": "flow",
     "flow:check": "flow check",
     "test": "lerna run test",
-    "prepack:all": "for I in packages/*/; do echo \"cd $I && npm run prepack\" | perl -p -e 's/\\n/\\0/;'; done | xargs -0 node_modules/.bin/concurrently",
+    "prepack:all": "scripts/prepack-all",
```
benjie: Order now matters (because TypeScript checks dependencies), so we're using a build script now.

packages/graphile-build/src/SchemaBuilder.d.ts
```diff
@@ -0,0 +1,221 @@
...
+export interface Inflection {
+  [str: string]: any;
+}
+export interface Scope<Type> {
+  [str: string]: any;
+}
+export interface Context<Type> {
+  scope: Scope<Type>;
+  [str: string]: any;
+}
```
benjie: A number of `any`s here; when we finish moving to TypeScript these will end up being a lot more precise.

I ultimately plan to use [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) so that each plugin can add its own types to these interfaces. This will make the plugin system a LOT easier to use 👍

packages/postgraphile-core/src/index.ts
```diff
@@ -27,31 +38,35 @@ const ensureValidPlugins = (name, arr) => {
 };
 
 type PostGraphileOptions = {
-  dynamicJson?: boolean,
```
benjie: - [ ] This should probably be an interface.

benjie:
Once last thing that needs to be done before this can be merged: set up tslint.

packages/graphile-build-pg/package.json
benjie: Graphile-build-pg is still in JS+Flow, so I’ve just added manual  .d.ts files for now.

packages/graphile-build/src/SchemaBuilder.d.ts
benjie: Perhaps, but I want people to be able to use it with TypeScript now rather than in a couple months when I finish moving it to TS. Mostly it was just copy and paste from Flow and then fix the syntax in a few places with Vim though 👍

packages/graphile-build-pg/src/index.d.ts
```diff
@@ -0,0 +1,79 @@
...
+export {
+  PgNamespace,
+  PgProc,
+  PgClass,
+  PgType,
+  PgAttribute,
+  PgConstraint,
+  PgExtension
+};
+
```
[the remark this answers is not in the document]
benjie: Yeah, also we use 4 of these on line 30 and I didn’t want to have two blocks importing them 👍

packages/graphile-build/src/SchemaBuilder.d.ts
```diff
@@ -0,0 +1,221 @@
...
+
+export type Hook<Type> = (
+  input: Type,
+  build: Build,
+  context: Context<Type>
+) => Type;
+
```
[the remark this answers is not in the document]
benjie: Interesting; I know you can use interfaces for functions but I thought you typically only did that when they also have properties. Does using an interface for a vanilla function have advantages over a type alias? I used an alias because it seems simpler.

cdaringe:
i missed all of this great action!  😎

benjie:
There’s much more to be done!

Issue #175, opened 2018-02-21
Postgraphile doesn't recognize UDFs with "out" keywords instead of "returns"


w9:
Given two equivalent user defined functions (UDF):

```
create function hello(out tax integer) as $$
begin
  tax := 42;
end;
$$
language plpgsql;

create function hello2() returns integer as $$
begin
  return 42;
end;
$$
language plpgsql;
```

Only `hello2` will be recognized and included in the generated schema.

The usage of the `out` keyword is documented [here](https://www.postgresql.org/docs/current/static/plpgsql-declarations.html#PLPGSQL-DECLARATION-PARAMETERS).

benjie:
Yes, we don't have `OUT` support currently - I should add that to the docs.

If you're interested in adding support, I can probably give you some pointers 👍

w9:
@benjie Sure, I would be glad to give it a try.

In the meantime, is there a way to give the returned value an alias? Right now by default the returned value just has the name as the type name.

w9:
@benjie Then I think the `out` keyword is a natural solution to this. You could just have

```
create function hello(out foo integer, out bar text) ...
```

and get a schema

```
hello(input: {}) {
  foo
  bar
}
```

w9:
@benjie And what is smart comments?

benjie:
Yep, OUT keyword is a natural fit definitely. However it's significantly more code to implement - particularly because if you have 2 or more OUT args that's supposed to be equivalent to returning a RECORD(...), so we'd need to declare a new dynamic type for it which we don't have code for currently. (Also this stuff is complex internally in general because of the way PostgreSQL/pg interact over these types.)

However, simply renaming something that already exists in the codebase is much MUCH less code.

Smart comments:

https://github.com/postgraphql/postgraphql/issues/473#issuecomment-314713154

https://github.com/graphile/graphile-build/pull/128

w9:
@benjie Thanks. Can you point to me places I might be interested to look at if I were to start working on the OUT keyword?

benjie:
This is used for all the procedures (mutations, computed fields, everything):

https://github.com/graphile/graphile-build/blob/master/packages/graphile-build-pg/src/plugins/makeProcField.js

Currently we just use the return type:

https://github.com/graphile/graphile-build/blob/552148e238475c95839e0d5fb3147d08868f3fd5/packages/graphile-build-pg/src/plugins/makeProcField.js#L100

[...]

PR #296, opened 2018-09-09
Support IN / OUT / INOUT arguments and RETURNS TABLE

mattbretl:
This lifts all the restrictions described at https://www.graphile.org/postgraphile/function-restrictions/ with the exception of variadic functions.

Fixes https://github.com/graphile/graphile-engine/issues/175


packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
```diff
@@ -461,6 +461,54 @@ create function c.return_table_without_grants() returns c.compound_key as $$
...
+  select 43 as o;
+$$ language sql stable;
+
+create function c.func_out_out(out first_out int, out second_out text) as $$
+  select 42 as first_out, 'out'::text as second_out;
+$$ language sql stable;
+
+create function c.func_out_out_unnamed(out int, out text) as $$
+  select 42, 'out'::text;
+$$ language sql stable;
```
benjie: Good catch! I was just testing this myself:

```sql
begin;
create function foo(out int, out text, out int) returns record as $$ select 1, 'foo', 3; $$ language sql immutable;
select * from foo();
rollback;
```

benjie: I'd also like to see a test for a mixed named/unnamed, e.g.

```sql
begin; create function foo(out int, out a text, out int) returns record as $$ select 1, 'foo', 3; $$ language sql immutable; select * from foo(); rollback;

 column1 |  a  | column3
---------+-----+---------
       1 | foo |       3
```

packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
```diff
@@ -461,6 +461,54 @@ create function c.return_table_without_grants() returns c.compound_key as $$
...
+  union
+  select i + 43 as col1;
+$$ language sql stable;
+
+create function c.func_returns_table_multi_col(i int) returns table (col1 int, col2 text) as $$
+  select i + 42 as col1, 'out'::text as col2
+  union
+  select i + 43 as col1, 'out2'::text as col2;
+$$ language sql stable;
+
```
benjie: We need to have some more trickier functions in here to really test this. Functions with output arguments that use compound types, output arguments that are tables (and check that relations still work on the tables); all these checks need to be done on all three types of functions: custom queries, custom mutations and computed fields.

I think we're also going to get problems for functions like the below that return raw `record` type without declaring the shape:

```sql
begin;
create function foo() returns record as $$ select 1, 'foo', 3; $$ language sql immutable;
select * from foo() as t(a1 int, a2 text, a3 int);
rollback;
```

These should be skipped by PostGraphile as we do not know how to specify the table alias required in the `select` statement. Adding them to the test suite will ensure that they are skipped, rather than causing PostGraphile to crash on startup.

packages/graphile-build-pg/src/plugins/PgRecordReturnTypesPlugin.js
```diff
@@ -0,0 +1,88 @@
...
+                  null
+                );
+                return {
+                  ...memo,
+                  [fieldName]: {
+                    type: fieldType,
+                    resolve(data) {
+                      return outputArgName !== ""
+                        ? data[outputArgName]
+                        : data[`column${idx + 1}`]; // FIXME?
```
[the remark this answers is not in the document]
benjie: Yes. They do. `columnN` seems to be the pattern that Postgres 10.x uses; but the docs say:

> (If you omit the name for an output argument, the system will choose a default column name.)
> https://www.postgresql.org/docs/10/static/sql-createfunction.html

I.e. they do not declare the actual name of the column.

Please add a comment to this code explaining why it has `column${idx + 1}` 👍

packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
mattbretl: Added a mixed named/unnamed test in d79a7ff.

benjie: 👍

benjie:
Liking this new `Resolve Conversation` feature in GitHub 👍

packages/graphile-build-pg/src/plugins/PgScalarFunctionConnectionPlugin.js
```diff
@@ -43,7 +43,10 @@ export default (function PgScalarFunctionConnectionPlugin(
         // on function arguments and return types, however maybe a later
         // version of PG will?
         const NodeType =
-          pgGetGqlTypeByTypeIdAndModifier(returnType.id, null) || GraphQLString;
+          returnType.id === "2249" // record type
+            ? getTypeByName(inflection.functionReturnsRecordType(proc))
+            : pgGetGqlTypeByTypeIdAndModifier(returnType.id, null) ||
+              GraphQLString;
```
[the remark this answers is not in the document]
mattbretl: :+1: e9b940c

mattbretl:
Not sure why that last travis run failed..  might need to increase the timeout?

packages/graphile-build-pg/src/plugins/PgBasicsPlugin.js
```diff
@@ -482,6 +482,9 @@ export default (function PgBasicsPlugin(
         functionInputType(proc: PgProc) {
           return this.upperCamelCase(`${this._functionName(proc)}-input`);
         },
+        functionReturnsRecordType(proc: PgProc) {
+          return this.upperCamelCase(`${this._functionName(proc)}-record`);
```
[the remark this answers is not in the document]
mattbretl: :+1: aecc835

benjie:
That started happening yesterday. I don't know why. Feel free to bump up the timeout - set it to 60 seconds. I wonder if it's because I update jest?

```
●badlyBehavedFunction.graphql
    Timeout - Async callback was not invoked within the 20000ms timeout specified by jest.setTimeout.
```

^ Everything cascades off this.

If the tests were running in parallel that would explain it, but that shouldn't be the case due to `lerna run --concurrency 1 test` combined with `jest -i`.

mattbretl:
No clue about the flow errors in 0666eb2

benjie:
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈ packages/graphile-build-pg/src/plugins/makeProcField.js:93:21
Propertylengthismissinginnull[1].
packages/graphile-build-pg/src/plugins/makeProcField.js
90│  const sliceAmount = computed ? 1 : 0;
91│  const argNames = proc.argTypeIds.slice(sliceAmount).reduce(
92│    (prev, _, idx) =>
93│      proc.argModes.length === 0 || // all args are `in`
94│      proc.argModes[idx] === "i" || // this arg is `in`
95│      proc.argModes[idx] === "b" // this arg is `inout`
96│        ? [...prev, proc.argNames[idx + sliceAmount] || ""]
packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
[1]43│  argModes: Array<"i" | "o" | "b" | "v" | "t"> |null
```


^ You declared that `argModes` can be null; yet you're accessing the `.length` property without checking that it's not null first which could result in a runtime type error. Change to `!proc.argModes || proc.argModes.length === 0 || ...`

```
Error ┈┈┈┈┈┈┈┈┈���┈┈ packages/graphile-build-pg/src/plugins/makeProcField.js:94:21
Cannotgetproc.argModes[idx]becauseanindexerpropertyismissingin
null[1].
packages/graphile-build-pg/src/plugins/makeProcField.js
91│  const argNames = proc.argTypeIds.slice(sliceAmount).reduce(
92│    (prev, _, idx) =>
93│      proc.argModes.length === 0 || // all args are `in`
94│proc.argModes[idx] === "i" || // this arg is `in`
95│      proc.argModes[idx] === "b" // this arg is `inout`
96│        ? [...prev, proc.argNames[idx + sliceAmount] || ""]
97│        : prev
packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
[1]43│  argModes: Array<"i" | "o" | "b" | "v" | "t"> |null
```

^ Same, but for the `[idx]` reference (need to check proc.argModes is not null)

```
Error ┈┈┈┈┈┈┈┈┈┈┈ packages/graphile-build-pg/src/plugins/makeProcField.js:102:21
Propertylengthismissinginnull[1].
packages/graphile-build-pg/src/plugins/makeProcField.js
99│  );
100│  const argTypes = proc.argTypeIds.slice(sliceAmount).reduce(
101│    (prev, typeId, idx) =>
102│      proc.argModes.length === 0 || // all args are `in`
103│      proc.argModes[idx] === "i" || // this arg is `in`
104│      proc.argModes[idx] === "b" // this arg is `inout`
105│        ? [...prev, introspectionResultsByKind.typeById[typeId]]
packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
[1]43│  argModes: Array<"i" | "o" | "b" | "v" | "t"> |null
```

^ same (for `.length`)

```
Error ┈┈┈┈┈┈┈┈┈┈┈ packages/graphile-build-pg/src/plugins/makeProcField.js:103:21
Cannotgetproc.argModes[idx]becauseanindexerpropertyismissingin
null[1].
packages/graphile-build-pg/src/plugins/makeProcField.js
100│  const argTypes = proc.argTypeIds.slice(sliceAmount).reduce(
101│    (prev, typeId, idx) =>
102│      proc.argModes.length === 0 || // all args are `in`
103│proc.argModes[idx] === "i" || // this arg is `in`
104│      proc.argModes[idx] === "b" // this arg is `inout`
105│        ? [...prev, introspectionResultsByKind.typeById[typeId]]
106│        : prev
packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
[1]43���  argModes: Array<"i" | "o" | "b" | "v" | "t"> |null
```

^ same (for `[idx]`)

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -40,6 +40,7 @@ export type PgProc = {
   returnTypeId: string,
   argTypeIds: Array<string>,
   argNames: Array<string>,
+  argModes: Array<string>,
```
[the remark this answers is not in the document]
mattbretl: Actually, we're using `coalesce` to return an empty array during introspection, so I'm dropping the `| null`: 8187b74

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.d.ts
```diff
@@ -20,6 +20,7 @@ export interface PgProc {
   returnTypeId: string;
   argTypeIds: Array<string>;
   argNames: Array<string>;
+  argModes: Array<string>;
```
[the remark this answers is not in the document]
mattbretl: Actually, we're using `coalesce` to return an empty array during introspection, so I'm dropping the `| null`: 8187b74

packages/graphile-build-pg/src/plugins/PgRecordReturnTypesPlugin.js
```diff
@@ -0,0 +1,88 @@
...
+        const returnType =
+          introspectionResultsByKind.typeById[proc.returnTypeId];
+        const returnTypeTable =
+          introspectionResultsByKind.classById[returnType.classId];
+        if (returnTypeTable) {
+          // Just use the standard table connection from PgTablesPlugin
+          return;
+        }
+        const isReturnTypeRecord = returnType => returnType.id === "2249";
+        if (!isReturnTypeRecord) {
```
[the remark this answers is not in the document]
mattbretl: Good catch. Fixed in 5a22b84

packages/graphile-build-pg/src/plugins/PgRecordReturnTypesPlugin.js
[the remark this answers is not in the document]
mattbretl: Done. https://github.com/graphile/graphile-engine/pull/296/commits/a6abe0bf0706cea1526a0ed7179e57f42ff9f9ca

packages/graphile-build-pg/src/plugins/PgRecordReturnTypesPlugin.js
```diff
@@ -0,0 +1,88 @@
...
+              }, {}),
+          },
+          {
+            __origin: `Adding record return type for ${describePgEntity(
+              proc
+            )}. You can rename the function's GraphQL field (and its dependent types) via:\n\n  ${sqlCommentByAddingTags(
+              proc,
+              {
+                name: "newNameHere",
+              }
```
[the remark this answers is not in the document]
mattbretl: :+1: d00b325

packages/graphile-build-pg/src/plugins/PgRecordReturnTypesPlugin.js
```diff
@@ -0,0 +1,88 @@
...
+            )}`,
+            isRecordReturnType: true,
```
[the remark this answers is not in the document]
mattbretl: :+1: 9eab349

packages/graphile-build-pg/src/plugins/PgRecordReturnTypesPlugin.js
```diff
@@ -0,0 +1,88 @@
...
+        newWithHooks(
+          GraphQLObjectType,
+          {
+            name: inflection.functionReturnsRecordType(proc),
+            description: `The return type of our \`${inflection.functionQueryName(
+              proc
+            )}\` query.`,
+            fields: () =>
+              outputArgNames.reduce((memo, outputArgName, idx) => {
+                const fieldName = inflection.argument(outputArgName, idx + 1);
```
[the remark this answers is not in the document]
mattbretl: Done. I named it `functionOutputFieldName` to follow the apparent conventions. bd5b3f7

packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
mattbretl: > output arguments that are tables (and check that relations still work on the tables)

Can you elaborate on your expectations for specifying tables as output arguments? Do you just want to support definitions like:

```sql
create function func_out_table3(out person) as $$
  ...
$$ language plpgsql stable;
```

These aren't currently exposed because of this: https://github.com/graphile/graphile-engine/blob/9d2855f2196424a5da1353c7c3390371d61593b1/packages/graphile-build-pg/src/plugins/PgQueryProceduresPlugin.js#L56-L63

..but that would be a simple to fix.

mattbretl: In other words, if there's a single output argument of a table type (`func(out person)`), expose it just like `func() returns person`. Any other variation of output args that include a table type would be ignored. Example of function that would be ignored:
```
create function func_out_table99(out person, out something_else text)
  returns setof record as $$ ...
```
Correct?

benjie: Oh you're right - things are way more complex than I realised - the codebase currently assumes that all arguments are input arguments (like that code you reference does) - we need to go through the current codebase and add "and is an input argument" checks in the relevant places. I wonder how computed columns should handle `inout` arguments... 🤔

You should be able to output a mix of scalar and compound types:

```sql
create type compound_type as (
  c int,
  d text
);
create table my_table (
  e int,
  f text
);

create function my_func(in a int, in b text, out x int, out y compound_type, out z my_table) as $$ ... $$ language sql;
```

mattbretl: > we need to go through the current codebase and add "and is an input argument" checks in the relevant places

Done (I think) via 39a94ef

> if there's a single table output argument like `func(out person)`, expose it just like `func() returns person`

I added tests via f537a78  to ensure that this (and it's `setof` counterpart) work.

You want to go down the rabbit hole towards your `my_func` example within this PR?

mattbretl: No apparent issues with compound types: 6e93503

Still working to get your `my_func` torture test to pass.

benjie: Should also add a relation from my_table to another table and check that PostGraphile can navigate it as expected during the query.

benjie: I think you’ll need to recurseDataGenerators on the record field of the mutation payload; can’t remember if that’s done automatically for you... might be?

benjie: I’m happy to pitch in on this if you get stuck; the code to achieve it may be somewhat complex. It’s going to be a couple weeks before I can though.

benjie:
You’re doing excellent work 👍👍🙏

packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
mattbretl: Ok, your `my_func` plus a related table is finally working: 8bf04ba. Hope I didn't butcher the implementation.. this is clearly the "make it work" phase. :sweat_smile:

mattbretl: Connections are working now (c163c2e; test: 88d39aa). I realize this hijacking of PgScalarFunctionConnectionPlugin isn't ideal, but doing it this way helps me track the changes in logic. I can refactor after everything's working as intended.

mattbretl: > I wonder how computed columns should handle inout arguments...
Any special concerns? I think it's working as expected: 06e111f

mattbretl:
Ready for another review. Queries, computed columns, and mutations are all working now. :tada: 

The SQL functions added for the mutation tests are direct copies of the ones used for queries, with `stable` changed to `volatile`.

Fair warning: In order to minimize the diff, I often jammed things in, resulting in nested ternaries :unamused:. We'll want to refactor some of that.

The viaTemporaryTable updates were particularly gruesome. With named types, `(str::${sqlTypeIdentifier}).*` works to cast the record-stored-as-text (`str`) to the named type. I found no way to cast a record-stored-as-text dynamically to a recordset; I wish something like `(str as a(b int, c text))` were possible, but PG kept throwing `ERROR:  input of anonymous composite types is not implemented`. This meant storing record as text was a non-starter, as there was no way to extract the data back out. As a workaround, I cast the individual columns as text, and stored the record as text[]. This worked great until the `unnest((${sql.value(values)})::text[])` came along. When calling `unnest` on a multidimensional array, it flattens the array before generating records; I could not find a workaround for this without resorting to a plpgsql function. So.. the `unnest` call had to become a `values` call in order to generate rows with a text[] in each row, and then I extracted column values from the text[] using the array index. :sweat_smile: I'm obviously open to saner approaches, but this was the best I could come up with.

benjie:
> I found no way to cast a record-stored-as-text dynamically to a recordset

Mutations don't deal in recordsets anyway (we convert them to arrays because you can't paginate over a mutation), does treating it as an array of records work instead?

benjie:
Wow. Sounds like this was even more of a pain than I anticipated (and I knew it was going to be a pain!) - thanks for your heroic efforts Matt!

Here's another approach that might be worth exploring

```sql
select * from (values(1, 2, 3), (2,3,4), (3,4,5)) as t(i, j, k);
```

We could also look into using `jsonb_to_recordset` and the opposite in `viaTemporaryTable` but I'm not sure if it would corrupt certain data.

packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
benjie: Just wondering what'd happen if you made the table arg itself inout.

packages/graphile-build-pg/src/plugins/makeProcField.js
```diff
@@ -190,10 +222,41 @@ export default function makeProcField(
       fieldScope.pgFieldIntrospectionTable = returnTypeTable;
       payloadTypeScope.pgIntrospectionTable = returnTypeTable;
     }
+  } else if (isRecordLike) {
+    const RecordType = getTypeByName(
+      inflection.functionReturnsRecordType(proc)
+    );
```
benjie: Handle `RecordType` not found issue (just don't make the field)

packages/graphile-build-pg/src/plugins/makeProcField.js
```diff
@@ -190,10 +222,41 @@ export default function makeProcField(
...
+    if (proc.returnsSet) {
+      if (isMutation) {
+        type = new GraphQLList(RecordType);
+      } else if (forceList) {
+        type = new GraphQLList(RecordType);
+        fieldScope.isPgFieldSimpleCollection = true;
+      } else {
+        const ConnectionType = getTypeByName(
+          inflection.scalarFunctionConnection(proc)
```
benjie: I'm okay with using this method, but please add a comment to the inflectors saying it's now a misnomer and isn't just used for scalars.

packages/graphile-build-pg/src/plugins/makeProcField.js
```diff
@@ -555,11 +620,16 @@ export default function makeProcField(
                     sql.query`select ${
                       isPgClass
                         ? sql.query`${intermediateIdentifier}.*`
-                        : sql.query`${intermediateIdentifier}.${intermediateIdentifier} as ${functionAlias}`
+                        : isPgRecord
+                          ? sql.query`${intermediateIdentifier}.*`
+                          : sql.query`${intermediateIdentifier} as ${functionAlias}`
```
benjie: I believe we used the `.*` here for performance reasons - there's an incredible performance difference between `select users from users` and `select users.* from users`. This isn't the kind of thing that typically shows up in tests, but I'd prefer the behaviour of this code didn't change for classes.

packages/graphile-build-pg/src/plugins/viaTemporaryTable.js
```diff
@@ -43,7 +43,11 @@ export default async function viaTemporaryTable(
   sqlMutationQuery: SQL,
   sqlResultSourceAlias: SQL,
   sqlResultQuery: SQL,
-  isPgClassLike: boolean = true
+  isPgClassLike: boolean = true,
+  isPgRecord: boolean = false,
+  // eslint-disable-next-line flowtype/no-weak-types
+  outputArgTypes: Array<any> = [],
+  outputArgNames: Array<string> = []
```
benjie: Can you confirm these changes are backwards compatible? `viaTemporaryTable` is now an API that plugins use, I don't want to break any of them.

benjie: I wonder if we should merge these things into one argument - an object - that we just use for records. That way changes can be made easier in future without risking breaking things.

packages/graphile-build-pg/src/plugins/viaTemporaryTable.js
```diff
@@ -83,7 +87,17 @@ export default async function viaTemporaryTable(
          * representation to make it easier to deal with below.
          */
         sql.query`(case when ${sqlResultSourceAlias} is null then null else ${sqlResultSourceAlias} end)`
-      : sql.query`(${sqlResultSourceAlias}.${sqlResultSourceAlias})::${sqlTypeIdentifier}`;
+      : isPgRecord
+        ? sql.query`array[${sql.join(
+            outputArgNames.map(
+              (outputArgName, idx) =>
+                sql.query`${sqlResultSourceAlias}.${sql.identifier(
+                  outputArgName !== "" ? outputArgName : `column${idx + 1}`
```
benjie: Add a comment here noting that `column1` etc is what PostgreSQL picks currently if there's no name, but that we have no guarantee this will always be the case.

packages/graphile-build-pg/src/plugins/viaTemporaryTable.js
```diff
@@ -101,16 +115,38 @@ export default async function viaTemporaryTable(
...
+                )} as ${sql.identifier(
+                  outputArgName !== "" ? outputArgName : `column${idx + 1}`
+                )}`
+            ),
+            ", "
+          )}
+          from (values ${sql.join(
+            values.map(value => sql.query`(${sql.value(value)}::text[])`),
+            ", "
+          )}) as x(arr)`
```
benjie: Turns out using a values list is exactly what you did 👍

benjie: I'm wondering if `jsonb_to_recordset` is safe to use for *just* `isPgRecord` (since it's new behaviour so definitely not breaking) which'd avoid this complexity.

packages/graphile-build-pg/src/plugins/makeProcField.js
mattbretl: See 92dbc7e; makeProcField is expected to return a field spec, so it has to throw instead of returning silently. I didn't want to get into refactoring the calls to makeProcField unless you think that's necessary.

packages/graphile-build-pg/src/plugins/makeProcField.js
mattbretl: At this point, I have a good handle on the logic changes, so I propose we leave PgScalarFunctionConnectionPlugin.js/scalarFunctionEdge()/scalarFunctionEdge() unmodified by this PR and create a new PgRecordFunctionConnectionPlugin.js with recordFunctionEdge()/recordFunctionConnection() inflectors. Then there would no need to explain the hijacking and fewer sprinkles of `recordType.id === "2249"` throughout. Thoughts?

packages/graphile-build-pg/src/plugins/makeProcField.js
mattbretl: For classes, the query fragment remains the same.

For scalars, it changed from `${intermediateIdentifier}.${intermediateIdentifier} as ${functionAlias}` to `${intermediateIdentifier} as ${functionAlias}`.

(It would have been clearer to just put `|| isPgRecord` next to `isPgClass`, but I was trying to follow a class-record-scalar pattern for the conditionals.)

packages/graphile-build-pg/src/plugins/viaTemporaryTable.js
mattbretl: I believe the changes to viaTemporaryTable are backwards compatible. `isPgRecord` defaults to false, so scalars will just pass through to the else clause of the two ternaries I inserted. I did have to refactor `convertFieldBack` to include more of the query, but resulting SQL is identical for both classes and scalars. `outputArgTypes` and `outputArgNames` are only referenced if `isPgRecord` is passed as true, so no breakage there.

Plugins might have issues handling all the new procs they'll be seeing on introspectionResultsByKind.proc, perhaps incorrectly inferring an incorrect value for isPgClassLike and passing that through to viaTemporaryTable.. but that's not really a breaking change in the viaTemporaryTable API itself.

As far as wrapping record-related args into an object, are you suggesting `{ outputArgTypes: Array<any> = [],
  outputArgNames: Array<string> = [] }`, or including isPgRecord as well? I think isPgRecord should be adjacent to isPgClassLike, but I'm fine putting the other two into an object.

packages/graphile-build-pg/src/plugins/viaTemporaryTable.js
mattbretl: Done. eb9382c

packages/graphile-build-pg/src/plugins/viaTemporaryTable.js
benjie: I was thinking that the presence of the `{ outputArgTypes, outputArgNames }` object would infer `isPgRecord` - send null/undefined if `!isPgRecord`

packages/graphile-build-pg/src/plugins/makeProcField.js
benjie: You're right, I misread the diff.

packages/graphile-build-pg/src/plugins/makeProcField.js
benjie: Sure 👍

packages/graphile-build-pg/src/plugins/makeProcField.js
benjie: No that's fine, I think many are wrapped with swallowError anyway. I just want to ensure that if someone deliberately skips the record plugin that they will still be able to boot their PostGraphile instance even if it has INOUT / RETURNS TABLE functions.

mattbretl:
> > I found no way to cast a record-stored-as-text dynamically to a recordset
> 
> Mutations don't deal in recordsets anyway (we convert them to arrays because you can't paginate over a mutation), does treating it as an array of records work instead?

Sorry, I misspoke. What I meant was: "I found no way to cast a record-stored-as-text dynamically to an actual row using a column definition list."

With a named type, the following method (which is how viaTemporaryTable has always operated) works just fine:
```sql
create type foo as (a int, b int);

select (str::foo).* from unnest(array['(1,11)','(2,22)']::text[]) str;

--  a | b  
-- ---+----
--  1 | 11
--  2 | 22
-- (2 rows)
```

I wish I just could replace `(str::foo).*` with something like `(cast(str as (a int, b int))).*`, but no dice.  Likewise, I found no way to turn an array of records-stored-as-text into a recordset using a column definition list:

```sql
select * from unnest(array['(1,11)','(2,22)']::record[]) as t(a int, b int);

-- ERROR:  input of anonymous composite types is not implemented
```
:unamused: 

Hence, the use of a 2D array and the ensuing complexity.

benjie:
Yeah; my initial comments were before I read the code. Do you think the jsonb_\* functions are viable candidates? I wonder if there's stuff that'd get munged by jsonb that wouldn't get munged by converting to text.

mattbretl:
> We could also look into using `jsonb_to_recordset` and the opposite in `viaTemporaryTable` but I'm not sure if it would corrupt certain data.

That was my fear as well; as ugly as the current implementation is, it uses very simple value::text casts (wrapped with `array[`...`]`). I would put more trust in that than jsonb_*.

mattbretl:
That said, you've got a lot more xp in these matters than I.. I'm more than willing to refactor to jsonb and see if anything blows up.

packages/graphile-build-pg/src/plugins/viaTemporaryTable.js
mattbretl: Done. afb4357

benjie:
Nah lets stick with casting to text. Everything supports that in Postgres. We just need to ensure that we get the right types on the output - they at least have to be named types... What happens if they have `create function foo(out r record)` or similar? We should skip over that entirely if we can.

packages/graphile-build-pg/src/plugins/makeProcField.js
mattbretl: Done. 10b9147

mattbretl:
> What happens if they have `create function foo(out r record)` or similar?

It throws upon querying `foo`. :man_facepalming: 

Fixed by updating the introspection query. 4aa79db

benjie:
Commencificating reviewification...

packages/graphile-build-pg/src/plugins/PgRecordReturnTypesPlugin.js
```diff
@@ -0,0 +1,88 @@
...
+          []
+        );
+        const outputArgTypes = proc.argTypeIds.reduce(
+          (prev, typeId, idx) =>
+            ["o", "b", "t"].includes(proc.argModes[idx])
+              ? [...prev, introspectionResultsByKind.typeById[typeId]]
+              : prev,
+          []
+        );
+        newWithHooks(
```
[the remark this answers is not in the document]
mattbretl: Understood. I just ran into a case of this (conflicts for types that don't end up being in the schema) the other day. I think we could refactor this into a `getOrCreate`-type function in the future.

benjie:
Seems that there's no null-checking on returned compound types. Might be that you need to add `addNullCase` or similar to a `queryFromResolveData` call? Specifically I think it might be here:

https://github.com/mattbretl/graphile-build/blob/9b8ee3b6caf7b00e9c7c5c7e91972c1631b3f08c/packages/graphile-build-pg/src/plugins/makeProcField.js#L382-L383

The `isTableLike` should probably include records too?

---

```sql
create schema e;
set search_path to e, public;
create table foo(id serial primary key, t text);
create table bar(id serial primary key, foo_id int references foo, i int);
create function blah(in nr int, inout txt text, out f foo, out b bar) as $$
begin
  txt = txt || nr::text;
  select * into f from foo order by id desc limit 1;
  select * into b from bar order by id desc limit 1;
end;
$$ language plpgsql stable;
```

```graphql
{
  blah(nr: 1) {
    f {
      id
    }
    b {
      id
    }
    txt
  }
}
```

```
{
  "errors": [
    {
      "message": "Cannot return null for non-nullable field Foo.id.",
      "locations": [
        {
          "line": 4,
          "column": 7
        }
      ],
      "path": [
        "blah",
        "f",
        "id"
      ],
      "stack": "Error: Cannot return null for non-nullable field Foo.id.\n    at completeValue 
```

packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
mattbretl: It Just Works :tada: 2a591ca

benjie:
Fixed the above comment.

packages/postgraphile-core/__tests__/helpers.js
```diff
@@ -3,7 +3,7 @@ const { readFile } = require("fs");
 const pgConnectionString = require("pg-connection-string");
 
 // This test suite can be flaky. Increase it’s timeout.
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 1000 * 20;
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 1000 * 60;
```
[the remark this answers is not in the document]
mattbretl: See https://github.com/graphile/graphile-engine/pull/296#issuecomment-421345376

benjie:
```
[pggql_test] # select null::e.foo is null;
 ?column?
----------
 t
(1 row)

Time: 0.262 ms
[pggql_test] # select (null,null)::e.foo is null;
 ?column?
----------
 t
(1 row)

Time: 0.308 ms
[pggql_test] # select null::e.foo is not distinct from null;
 ?column?
----------
 t
(1 row)

Time: 0.235 ms
[pggql_test] # select (null,null)::e.foo is not distinct from null;
 ?column?
----------
 f
(1 row)

Time: 0.205 ms
```

✨ Postgres!

packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
```diff
@@ -461,6 +461,312 @@ create function c.return_table_without_grants() returns c.compound_key as $$
...
+create function c.mu_out_table() returns c.person as $$
+  select * from c.person where id = 1;
+$$ language sql volatile;
+
+create function c.mu_out_table_setof() returns setof c.person as $$
+  select * from c.person;
+$$ language sql volatile;
+
+------------------- */
```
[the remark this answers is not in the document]
mattbretl: I didn't intend to commit that. Removed in fc1e5f9

(When I was testing, I created equivalent functions without the `in`/`out`/`inout`/`returns table` syntax to cross-check the APIs/queries.)

packages/graphile-build-pg/src/plugins/PgRecordFunctionConnectionPlugin.js
```diff
@@ -0,0 +1,179 @@
...
+                    }\` and cursor to aid in pagination.`,
+                    type: new GraphQLNonNull(
+                      new GraphQLList(new GraphQLNonNull(EdgeType))
+                    ),
+                    resolve(data, _args, _context, resolveInfo) {
+                      const safeAlias = getSafeAliasFromResolveInfo(
+                        resolveInfo
+                      );
+                      return data.data.map(entry => ({
+                        __cursor: entry.cursor,
```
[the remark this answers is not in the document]
mattbretl: :flushed: Fixed via 7d0afd5

packages/graphile-build-pg/src/plugins/PgRecordFunctionConnectionPlugin.js
```diff
@@ -0,0 +1,179 @@
...
+          return;
+        }
+        // TODO: PG10 doesn't support the equivalent of pg_attribute.atttypemod
+        // on function arguments and return types, however maybe a later
+        // version of PG will?
+        const NodeType = getTypeByName(
+          inflection.functionReturnsRecordType(proc)
+        );
+        if (!NodeType) {
+          return;
```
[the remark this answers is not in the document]
mattbretl: Correct.. NodeType should always be created by PgRecordReturnTypesPlugin. Added error via 007b928

packages/graphile-build-pg/src/plugins/viaTemporaryTable.js
```diff
@@ -99,16 +120,41 @@ export default async function viaTemporaryTable(
...
+      ? sql.query`\
+              select (str::${sqlTypeIdentifier}).*
+              from unnest((${sql.value(values)})::text[]) str`
+      : isPgRecord
+        ? sql.query`\
+          select ${sql.join(
+            outputArgNames.map(
+              (outputArgName, idx) =>
+                sql.query`\
+                    (arr)[${sql.literal(idx + 1)}]::${sql.identifier(
```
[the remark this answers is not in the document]
mattbretl: :pray:

mattbretl:
I think the only remaining item at this point is finalizing the mutation payload field names. Namely:
1. When we're returning a single field (`out a int` or `returns table (a int)`), what should the field be named?
2. When we're returning multiple fields, (`out a int, out b int` or `returns table (a int, b int)`) the field is currently named `result`/`results` depending on whether it's an array/setof function. Is that what we want?

packages/graphile-build-pg/src/plugins/makeProcField.js
```diff
@@ -379,11 +426,10 @@ export default function makeProcField(
         return memo;
       }, {});
       if (isMutation) {
-        const resultFieldName = getResultFieldName(
+        const resultFieldName = inflection.functionMutationResultFieldName(
           proc,
-          type,
-          rawReturnType,
-          proc.returnsSet
+          getNamedType(type).name,
```
benjie: Pass the type, rather than its name; it gives greater flexibility.
```suggestion
          getNamedType(type),
```

mattbretl: Done. c829caf

benjie:
> When we're returning a single field (`out a int` or `returns table (a int)`), what should the field be named?

Single `out` field: name it the name it's given (i.e. `a`); in the absence of a name, follow the naming conventions for `returns int`/etc for consistency. If people want it to be called `result` they can either renaming the arg or override the inflector 👍


> When we're returning multiple fields, (`out a int, out b int` or `returns table (a int, b int)`) the field is currently named `result`/`results` depending on whether it's an array/setof function. Is that what we want?

Yep, sounds good to me.

benjie:
Caveat: ensure that you can override all these names with a smart comment on the function (as you can currently).

mattbretl:
Done. Ready for review.

packages/postgraphile-core/__tests__/fixtures/mutations/mutation-return-types.graphql
```diff
@@ -110,6 +110,6 @@ mutation {
     }
   }
   mutationReturnsTableOneCol(input: { i: 20 }) {
-    integers
+    col1S
```
benjie: I hate this camel case pluralize bug. Oh well, can't fix it now (except by using a plugin).

packages/postgraphile-core/__tests__/helpers.js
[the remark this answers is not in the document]
benjie: Yeah still no idea why it's happening. Ought to look into it really.

benjie:
LGTM; I'm going to do some testing, then hopefully merge.

benjie:
I haven't managed to break it. Merging 🎉 

Next up to fix the docs!

benjie:
This has been an incredibly large effort from you Matt, excellent work - on behalf of the PostGraphile community: thanks so much!

I'm going to mark this support as experimental (i.e. we can change it in a minor release if we need to) as I suspect there might be an edge case or two we've not met yet, but I've not managed to find any 😁

PR #297, opened 2018-09-12
fix(pagination): fixes bug with hasPreviousPage when paginating without unique orderBy

benjie:
Fixes https://github.com/graphile/postgraphile/issues/844


benjie:
@jemgillam did some testing and found two issues for non-ordered queries (e.g. a collection from a function):

1. using after and offset together throws an error `Must only set offset once`
2. if you set `before:` and give it the cursor of the first row then `hasNextPage` is incorrectly `false`

benjie:
> if you set before: and give it the cursor of the first row then hasNextPage is incorrectly false

Actually this is fine - no rows are returned so there's no cursor you can use for the next page and thus hasNextPage is fine to be false. If you were paying attention to `hasPreviousPage` then you shouldn't be able to get into this state 👍

benjie:
after/offset now works on function connections (and test added to check) so we should be good to go now 👍

PR #301, opened 2018-09-19
fix: Improve error message for invalid pg configurations

petetnt:
As seen in the [docs](https://node-postgres.com/api/client), `pg` also accepts an object as the configuration:

```js
type pgConfigObject = {
  user?: string, // default process.env.PGUSER || process.env.USER
  password?: string, //default process.env.PGPASSWORD
  database?: string, // default process.env.PGDATABASE || process.env.USER
  port?: number, // default process.env.PGPORT
  connectionString?: string, // e.g. postgres://user:password@host:5432/database
  ssl?: any, // passed directly to node.TLSSocket
  types?: any, // custom type parsers
  statement_timeout?: number, // number of milliseconds before a query will time out default is no timeout
}
```

All the fields are entirely optional, would be filled with the env variables if available:

> Every field of the config object is entirely optional. A Client instance will use environment variables for all missing values.

This PR adds the option to pass in an object as the `pgConfig`.

(`ssl` and `types` are commented due to `no-weak-types` rule. The `ssl` config for example is quite complex and most likely not very relevant).

Not very familar of `graphile-engine` yet so let me know if there's any changes necessary. 

edit: Weird, build failed. It passed on my branch though https://travis-ci.com/petetnt/graphile-engine

### DIFF
--- a/packages/graphile-build-pg/src/withPgClient.js
+++ b/packages/graphile-build-pg/src/withPgClient.js
@@ -1,6 +1,7 @@
 // @flow
 import pg from "pg";
 import debugFactory from "debug";
+
 const debug = debugFactory("graphile-build-pg");
 
 function constructorName(obj) {
@@ -72,7 +73,9 @@ const withPgClient = async (
         pgClient.connect(err => (err ? reject(err) : resolve()))
       );
     } else {
-      throw new Error("You must provide a valid PG client configuration");
+      throw new Error(
+        "You must provide either a pg.Pool or pg.Client instance or a PostgreSQL connection string."
+      );
     }
     result = await fn(pgClient);
   } finally {


benjie:
If you're doing this:

```js
const postgraphile = require('postgraphile');

postgraphile({user: ..., password: ...}, ...);
```

It doesn't seem much more effort to do this:

```js
const postgraphile = require('postgraphile');
const { Pool } = require('pg');

postgraphile(new Pool({user: ..., password: ...}), ...);
```

and that way there's not another interface to maintain and test.

However, I recognise that you're doing this as part of `withPgClient` in order to create an individual client rather than a pool, so it seems to me that you have a specific use-case in mind. Please could you elaborate on how you see this being useful?

petetnt:
Hi @benjie!

Oh dang, I completely forgot to mention what this is supposed to fix 🤦‍♂️ . Sorry about that!

The issue is more on the error message the side. The `pg` docs all refer to the `pgConfig` being either a object or a string while the error message says `Error: You must provide a valid PG client configuration`.

I have always used the `object` form of the `pgConfig` (and a team member ran into a similar issue few months ago too). FWIW I would be okay with just either accepting the `new pgPool(conf)` version but not the `postgres://u:p@host/db` connection string, but if the connection string is accepted I think it would be reasonable that the configuration object would be accepted too (as it is in `pg` in general).

Alternatively we can just change the version to throw something like `Please use the connection string or a pg.Pool client as the configuration object` instead of `Error: You must provide a valid PG client configuration` when object is being passed in.

Thanks for your great work!

packages/graphile-build-pg/src/withPgClient.js
```diff
@@ -59,7 +71,11 @@ const withPgClient = async (
       const pgPool = (pgConfig: pg.Pool);
       pgClient = await pgPool.connect();
       releasePgClient = () => pgClient.release();
-    } else if (pgConfig === undefined || typeof pgConfig === "string") {
+    } else if (
+      pgConfig === undefined ||
+      typeof pgConfig === "string" ||
+      typeof pgConfig === "object"
```
benjie: Please make this check stricter, e.g. use `lodash.isPlainObject`.

benjie:
We're definitely going to keep accepting the string (it'd be a breaking change not to); but I'm not super keen on supporting the generic object because that would mean that pg's breaking config changes become our breaking changes. I've added a note about using `isPlainObject` but TBH I think I'd rather that we just edited the error message to state something like: "Please pass either a pg.Pool or pg.Client instance or a PostgreSQL connection string."

I definitely want to improve the error if it's something that's confused you/colleagues.

I'm concerned if we're not careful we'll have people feeding invalid things in as the pgConfig (e.g. the PostGraphile configuration object itself) and wondering why it isn't working. e.g. `postgraphile({dynamicJson: true})`. This seems hard to guard against other than outright rejecting configuration objects as we do currently. (And anything like this increases my support burden which reduces time I can spend working on PostGraphile.)

petetnt:
Would something like

```
} else if (isPlainObject(pgConfig)) {
  throw new Error("Please pass either a pg.Pool or pg.Client instance or a PostgreSQL connection string.");
}
```

do?

benjie:
I think literally replace the default error message:

```js
throw new Error("You must provide a valid PG client configuration");
```

With a better one

```js
throw new Error("You must provide either a pg.Pool or pg.Client instance or a PostgreSQL connection string.");
```

No need for the `else if` I think.

petetnt:
Updated in 288350c

benjie:
👍

PR #309, opened 2018-09-28
Fix types and upgrade dependencies

lifeiscontent:


benjie:
Hi @lifeiscontent, thanks for the PR. I'm sorry to say we're in the process of removing Flow and moving to TypeScript for the whole stack. Work on this has already progressed a fair way, almost everything has TypeScript definitions now and some of the modules are 100% TypeScript. I will not be merging Flow fixes/changes (especially where updating the 'flow-bin' dependency is involved) as it will likely cause conflicts with this ongoing work. 

I'm not keen on reordering the `import` statements without good reason, and this PR also has some dangerous updates to dependencies - for example it changes the GraphQL range:

```diff
-   "graphql": ">=0.9 <0.14",
+   "graphql": "^14.0.2",
```

which would be a major breaking change.

For a project like this, updating dependencies is something that needs to be done carefully to ensure that no-one's applications are broken. I have to review the change log of every dependency to make sure that the changes won't cause issues downstream. Sadly `yarn upgrade --latest` is not typically viable on an npm module so many people depend on.

I'm going to close this PR, but if there are certain (smaller) changes that you wish to make please chat with me on Discord or file an issue and we can agree on a direction 👍

A note for the future: it's normally best for PRs to be small and focussed, changing as few lines as possible - this makes it easier to review each change in isolation, and to ensure that no other functionality has been affected. Most maintainers would much rather receive 3 small self-contained PRs than one large one.

Thanks again,

Benjie.

PR #324, opened 2018-10-18
feat: `pgIgnoreIndex = false` feature to honour indexes

benjie:
- [x] Only expose conditions on fields with indexes
- [x] Only expose orderBy on fields with indexes
- [ ] Optional: only expose foreign key relations if they have indexes (??)


benjie:
I'm not happy with how the errors work in this. In particular they are repetitive within the same build, and they're going to get old fast in watch mode. Too much information is sometimes worse than too little information - users tend to ignore it as noise.

I think I'm going to remove them actually. Or hide them behind a debug flag.

Yeah, debug flag is the way to go I think. We should come up with a better system for schema warnings.

PR #325, opened 2018-10-22
feat: `pgLegacyFunctionsOnly` option

mattbretl:
This is an opt-out option for the functionality introduced in #296.

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
@@ -186,10 +186,9 @@ export default (async function PgIntrospectionPlugin(
             versionResult.rows[0].server_version_num,
             10
           );
-          const introspectionQuery = makeIntrospectionQuery(
-            serverVersionNum,
-            pgLegacyFunctionsOnly
-          );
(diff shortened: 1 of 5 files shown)


packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -185,7 +186,10 @@ export default (async function PgIntrospectionPlugin(
             versionResult.rows[0].server_version_num,
             10
           );
-          const introspectionQuery = makeIntrospectionQuery(serverVersionNum);
+          const introspectionQuery = makeIntrospectionQuery(
+            serverVersionNum,
+            pgLegacyFunctionsOnly
```
benjie: I feel like this list of options could get quite big quite fast; should we make it an object instead?

benjie:
This is fine to merge as-is, but I'd like your input on the above question first.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
mattbretl: I went back and forth.. but taking a fresh look, I think it should be an options object. I'll make the change.

Issue #331, opened 2018-11-02
Feature request: add option to sort NULLs last


wanradt:
Hi!

As Postgres by default sorts NULLs first while using descending sort, so does postgraphile. But in many/most situations is desired behaviour to put NULLS LAST.

For now there is no configuration option in postgraphile to trigger such behaviour.

TIA,

Gunnar

benjie:
We just merged this in #332 👍 

We're not currently exposing this with a top-level flag, so to use it you need to use `graphileBuildOptions`; e.g.:

```js
app.use(postgraphile(DATABASE_URL, SCHEMA, {
  graphileBuildOptions: { orderByNullsLast: true }
}));
````

christian-predebon:
Hi, I cannot find anything related to the flag `orderByNullsLast` in the docs. Is this flag still used and if so why can't I find it inside the type `PostGraphileOptions`?

benjie:
Please see the example in the comment above yours - it lives within the graphileBuildOptions

christian-predebon:
Oh right, I read it wrong. Thanks!

JohnMusleh:
is there a way to define this for a single query and not on this level ? 

assume I want to use 2 different behaviors in different queries

benjie:
Each enum value used for ordering can override the nulls first/last; but that's on an enum-value basis, not a field basis. You can also add your own orders that can specify nulls behaviour via [makeAddPgTableOrderByPlugin](https://www.graphile.org/postgraphile/make-add-pg-table-order-by-plugin/)

JohnMusleh:
Thanks @benjie , which enum value are you talking about ?
can you give an example to better understand ?

benjie:
The enum values that an `orderBy` argument uses. Here's where we add enum values for each column:

https://github.com/graphile/graphile-engine/blob/ff52f00751a1eb09e391a0ec6319815875de470a/packages/graphile-build-pg/src/plugins/PgOrderAllColumnsPlugin.js#L38

That tuple has two entries, you can add a third as `true` or `false` to force `NULLS FIRST` or `NULLS LAST` (I don't remember which way round it is though, sorry!).

barbalex:
Sorry but after trying it out I haven't succeeded so asking: How do I achieve to sort with nulls first?

- is there a `orderByNullsFirst` option?
- would it be `orderByNullsLast: false`?

No matter what I do nulls seem to end up last.

[...]

PR #332, opened 2018-11-02
feat(pg): support ordering with NULLS FIRST / NULLS LAST

mattbretl:


mattbretl:
This allows plugins to solve #331

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -465,11 +465,17 @@ class QueryBuilder {
...
                   sql.fragment`${expr} ${
                     Number(ascending) ^ Number(flip)
                       ? sql.fragment`ASC`
                       : sql.fragment`DESC`
+                  }${
+                    nullsFirst === true
+                      ? sql.fragment` NULLS FIRST`
+                      : nullsFirst === false
+                        ? sql.fragment` NULLS LAST`
+                        : null
```
benjie: I had to actually read the source to make sure `null` was okay here! [It is ✅](https://github.com/graphile/pg-sql2/blob/d99ed2313976b1193eed2ff03f6639c8a04ac32b/src/index.ts#L193)

mattbretl:
Given how the tests are structured, I can add a new integration test  in [postgraphile-core/\_\_tests\_\_/integration/schema](https://github.com/graphile/graphile-engine/tree/master/packages/postgraphile-core/__tests__/integration/schema), appending a small plugin to expose additional orderBy enums that use `NULLS LAST`. Is that what you had in mind?

mattbretl:
Or.. I could add an `orderByNullsLast` option to PgConnectionArgOrderBy.js, and use that (if provided) when calling queryBuilder.orderBy(). Then the test would be as simple as passing `orderByNullsLast: true` through `graphileBuildOptions`. That's probably all that most people (including myself) will want to accomplish with this.. overriding the PG default of NULLS FIRST for DESC.

packages/graphile-build-pg/src/QueryBuilder.js
mattbretl: Yeah, I find that really handy..  don't change it! :smile:

mattbretl:
> Basically what you're saying is "treat nulls as if they are smaller than everything" - i.e. they're first when ascending and last when descending?

Actually, no. This is about using NULLS LAST for both ASC and DESC. As a simple example, consider a `tasks` table with a nullable integer `priority` field. (Tasks can optionally be assigned a priority.) On the UI side, if you're sorting by high or low priority, you probably want to show prioritized tasks before unprioritized tasks, and that requires specifying NULLS LAST for DESC.

If 5 is high-priority and 1 is low-priority:
- High-priority tasks: `ORDER BY priority DESC NULLS LAST` (without `NULLS LAST`, you get all of the unprioritized tasks showing up first.. definitely not what you want)
- Low-priority tasks: `ORDER BY priority ASC`

If 1 is high-priority and 5 is low-priority:
- High-priority tasks: `ORDER BY priority ASC`
- Low-priority tasks: `ORDER BY priority DESC NULLS LAST` (without `NULLS LAST`, you get all of the unprioritized tasks showing up first.. probably not what you want)

> Can you flag that this behaviour will become the default in v5? I always want it so that when you reverse the order (ASC -> DESC) the rows return in the exact opposite order (which is not the case currently, if there are nulls).

It works that way currently. Records are returned in the exact opposite order, with nulls treated as greater than all non-nulls. So there should be no need to change the default in v5.

benjie:
Hmmm. Add --order-nulls-last flag then I guess.

benjie:
Ohhhh. Thanks for sharing. Reviewing now 👍

packages/graphile-build-pg/src/QueryBuilder.d.ts
```diff
@@ -25,7 +25,7 @@ export default class QueryBuilder {
   public where(exprGen: SQLGen): void;
   public whereBound(exprGen: SQLGen, isLower: boolean): void;
   public setOrderIsUnique(): void;
-  public orderBy(exprGen: SQLGen, ascending: boolean): void;
+  public orderBy(exprGen: SQLGen, ascending: boolean, nullsFirst: boolean): void;
```
benjie: Technically this should be `nullsFirst?: boolean` because it doesn't have to be specified

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -236,9 +236,9 @@ class QueryBuilder {
   setOrderIsUnique() {
     this.data.orderIsUnique = true;
   }
-  orderBy(exprGen: SQLGen, ascending: boolean = true) {
+  orderBy(exprGen: SQLGen, ascending: boolean = true, nullsFirst: boolean) {
```
benjie: Technically this should also be `nullsFirst?: boolean` because it doesn't have to be specified. Huh, TypeScript and Flow use the same syntax for this 👍

benjie: Actually it should be `nullsFirst: boolean | null = null` I think.

packages/graphile-build-pg/src/QueryBuilder.d.ts
benjie: You know what... Lets have it required.

packages/graphile-build-pg/src/QueryBuilder.js
benjie: You know what... Lets have it required. It should still be `| null` technically...

PR #334, opened 2018-11-12
do not duplicate getTypeAndIdentifiersFromNodeId code

alexppxela:
If you want to add nodeId hooks, it is necesssary to have hooks on:
- nodeId creation (ok because getNodeIdForTypeAndIdentifiers is used everywhere)
- nodeId read (not ok because getTypeAndIdentifiersFromNodeId is duplicated twice)

### DIFF
--- a/packages/graphile-build/src/plugins/NodePlugin.js
+++ b/packages/graphile-build/src/plugins/NodePlugin.js
@@ -70,7 +70,7 @@ export default (function NodePlugin(
           getTypeAndIdentifiersFromNodeId(nodeId) {
             const [alias, ...identifiers] = JSON.parse(base64Decode(nodeId));
             return {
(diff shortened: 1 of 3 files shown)


packages/graphile-build/src/plugins/NodePlugin.js
```diff
@@ -70,7 +70,7 @@ export default (function NodePlugin(
           getTypeAndIdentifiersFromNodeId(nodeId) {
             const [alias, ...identifiers] = JSON.parse(base64Decode(nodeId));
             return {
-              Type: this.getTypeByName(nodeTypeNameByAlias[alias] || alias),
+              Type: this.getNodeType(alias),
```
benjie: 👍👍👍

PR #340, opened 2018-11-17
feat: add `@deprecated` support to functions

### LINKED ISSUE #336: Add `@deprecated` support to functions
I think literally deleting this one line of code will work:

https://github.com/graphile/graphile-engine/blob/61957931d236b78c0c8791420d052ae8af22aecf/packages/graphile-build-pg/src/plugins/PgColumnDeprecationPlugin.js#L17

Would make the plugin a misnomer though.

mattbretl:
Closes #336 

This allows deprecation of tables as well. I'm not sure whether that's desirable, but I wanted to see the diff.


mattbretl:
Ready for review. (custom query was tested in the `test function` commit: e4038d61dcdf6c0fc71fd508976ca01fbf42b581)

packages/graphile-build-pg/src/plugins/PgColumnDeprecationPlugin.js
```diff
@@ -4,28 +4,19 @@ import type { Plugin } from "graphile-build";
...
+      scope: { pgFieldIntrospection },
     } = context;
     if (
-      !(isPgRowType || isPgCompoundType) ||
-      !table ||
-      table.kind !== "class" ||
       !pgFieldIntrospection ||
-      !pgFieldIntrospection.tags
+      !pgFieldIntrospection.tags ||
+      !pgFieldIntrospection.tags.deprecated
```
benjie: This is a very broad check...

- Applies to CRUD mutations if the table is deprecated... fine ✅ 
- Applies to custom mutations if the function is deprecated... fine ✅ 
- Applies to columns on tables... fine ✅ 
- Applies to computed columns... fine ✅ 
- Applies to custom queries... fine ✅ 
- Applies to views... fine ✅ 

Okay, looks good. I can't think of a good reason this would not be desired.

packages/graphile-build-pg/src/plugins/PgColumnDeprecationPlugin.js
```diff
@@ -4,28 +4,19 @@ import type { Plugin } from "graphile-build";
...
       return field;
     }
-    return pgFieldIntrospection.tags.deprecated
-      ? Object.assign({}, field, {
-          deprecationReason: Array.isArray(pgFieldIntrospection.tags.deprecated)
-            ? pgFieldIntrospection.tags.deprecated.join("\n")
-            : pgFieldIntrospection.tags.deprecated,
-        })
-      : field;
+    return Object.assign({}, field, {
```
benjie: I prefer `{...field, `; "object spread properties" is supported as of Node 8.6 (which is the minimum requirement for PostGraphile for a reason 😉 ) according to https://node.green/

packages/postgraphile-core/__tests__/integration/schema/__snapshots__/noDefaultMutations.test.js.snap
```diff
@@ -1798,7 +1798,7 @@ type Person implements Node {
 
     \\"\\"\\"The method to use when ordering \`PersonSecret\`.\\"\\"\\"
     orderBy: [PersonSecretsOrderBy!] = [PRIMARY_KEY_ASC]
-  ): PersonSecretsConnection! @deprecated(reason: \\"Please use personSecretByPersonId instead\\")
+  ): PersonSecretsConnection! @deprecated(reason: \\"This is deprecated.\\")
```
benjie: - [x] Check this

packages/postgraphile-core/__tests__/integration/schema/__snapshots__/rbac.test.js.snap
```diff
@@ -753,7 +753,7 @@ type Person implements Node {
 
     \\"\\"\\"The method to use when ordering \`PersonSecret\`.\\"\\"\\"
     orderBy: [PersonSecretsOrderBy!] = [PRIMARY_KEY_ASC]
-  ): PersonSecretsConnection! @deprecated(reason: \\"Please use personSecretByPersonId instead\\")
+  ): PersonSecretsConnection! @deprecated(reason: \\"This is deprecated.\\")
```
benjie: - [x] Check this

packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
```diff
@@ -433,6 +437,8 @@ begin
 end;
 $$ language plpgsql stable;
 
+comment on function c.badly_behaved_function() is '@deprecated This is deprecated.';
+
```
benjie: In future, putting different comments on each thing makes it easier to confirm that the right things are being deprecated for the right reasons. `This is deprecated (badly_behaved_function).` would have been safer.

packages/postgraphile-core/__tests__/integration/schema/__snapshots__/noDefaultMutations.test.js.snap
benjie: This is where we have the legacy connection for a one-to-one relation; that added a deprecation message and then this overrode it.

... 🤔 ...

Okay, yeah I think that's fine - if you deprecate the entire table then that actually wins over the connection itself being deprecated. And throwing under these circumstances does not make sense. Sold.

PR #343, opened 2018-11-20
Add a fallback value for server_version_num in Redshift

drapadubok:
Related to https://github.com/graphile/postgraphile/issues/913

Redshift doesn't seem to allow getting server_version_num via the `show server_version_num` query, so here I add a fallback value as suggested in the issue above.
What would be a good way to test it? I'm not really familiar with Lerna, and not super familiar with code structure in Typescript projects. I

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
@@ -223,14 +223,21 @@ export default (async function PgIntrospectionPlugin(
     };
     const introspectionResultsByKind = cloneResults(
       await persistentMemoizeWithKey(cacheKey, () =>
-        withPgClient(pgConfig, async pgClient => {
-          const versionResult = await pgClient.query(
-            "show server_version_num;"
-          );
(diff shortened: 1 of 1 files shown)


benjie:
Hey @drapadubok; for testing things I prefer to do the following:

- Clone graphile-build; cd into `graphile-build`; run `yarn`
- Clone postgraphile into the graphile-build folder; run `yarn`
- Run a custom script I call `rmlocal.sh` which removes a bunch of stuff from node_modules so postgraphile looks for it in a higher folder (namely `graphile-build`'s folder)
- Run the `./scripts/dev` script passing the relevant options; e.g. `./scripts/dev -- --watch -c my_db -s app_public`

`rmlocal.sh`:

```bash
rm -Rf node_modules/graphile-build node_modules/graphile-build-pg node_modules/postgraphile-core/ node_modules/graphql-parse-resolve-info node_modules/graphql node_modules/pg-sql2 node_modules/@types/graphql node_modules/@types/pg
```

Bit of a hack I know. I've considered merging postgraphile and graphile-build into the same monorepo many times. Maybe I should. I'm on the fence still.

PR #361, opened 2018-12-02
fix(perf): detect unique columns and don't add primary key to order in that case

benjie:


packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -165,7 +165,7 @@ class QueryBuilder {
 
   // Helper function
   jsonbBuildObject(fields: Array<[SQL, RawAlias]>) {
-    if (this.supportsJSONB) {
+    if (this.supportsJSONB && fields.length > 50) {
```
benjie: Fixes a performance regression for unindexed selects - takes query time from ~4s back to ~1.7s in this case.

packages/graphile-build-pg/src/plugins/PgOrderAllColumnsPlugin.js
```diff
@@ -33,6 +34,7 @@ export default (function PgOrderAllColumnsPlugin(builder) {
               value: {
                 alias: ascFieldName.toLowerCase(),
                 specs: [[attr.name, true]],
+                unique,
```
benjie: This is the fix; it only works for single-column unique indexes. We should add a solution for multi-column unique indexes at some point 👍

packages/graphile-build-pg/src/plugins/debugSql.js
```diff
@@ -68,7 +63,16 @@ function debugSql(sql) {
...
+  return colouredSql;
+}
+
+const rawDebugSql = debugFactory("graphile-build-pg:sql");
+
+function debugSql(sql) {
+  if (!rawDebugSql.enabled) {
+    return;
+  }
+  rawDebugSql("%s", "\n" + formatSQLForDebugging(sql));
```
benjie: Exporting this helper so PostGraphile can use it.

packages/postgraphile-core/__tests__/integration/__snapshots__/mutations.test.js.snap
```diff
@@ -188,7 +188,7 @@ Object {
         },
       },
       "c": Object {
-        "cursor": "WyJpZF9hc2MiLFs5MDAwLDkwMDBdXQ==",
+        "cursor": "WyJpZF9hc2MiLFs5MDAwXV0=",
```
benjie: `["id_asc",[9000,9000]]` becomes `["id_asc",[9000]]`

packages/postgraphile-core/__tests__/integration/__snapshots__/mutations.test.js.snap
```diff
@@ -203,7 +203,7 @@ Object {
         },
       },
       "e": Object {
-        "cursor": "WyJlbWFpbF9hc2MiLFsiam9obm55LmJveS5zbWl0aEBlbWFpbC5jb20iLDkwMDBdXQ==",
+        "cursor": "WyJlbWFpbF9hc2MiLFsiam9obm55LmJveS5zbWl0aEBlbWFpbC5jb20iXV0=",
```
benjie: `["email_asc",["johnny.boy.smith@email.com",9000]]` becomes `["email_asc",["johnny.boy.smith@email.com"]]` because email is unique

packages/graphile-build-pg/src/plugins/PgOrderAllColumnsPlugin.js
mattbretl: The same change should be made to the desc enum, correct?

benjie: Oops!

benjie: https://github.com/graphile/graphile-engine/pull/364

PR #364, opened 2018-12-02
fix(perf): add missing `unique` setting

benjie:


benjie:
Snapshots need updating.

PR #365, opened 2018-12-07
feat: constraints through smart comments (view foreign keys, etc)

benjie:
- [ ] relations `@foreignKey (col1, col2) references other_table (col3, col4)` (#376)
- [ ] many-to-many relationship support (#91)
- [ ] primary key for nodeID / etc `@primaryKey (col1, col2)`
- [ ] ability to override non-null-ness of columns

Fixes #376


itaied246:
Getting this error for `view` foreign key:

```sql
create table post (
  id text primary key
);
comment on table post is E'@name post_table
@omit';

create table offer (
  id serial primary key,
  post_id text references post(id) not null,
);
comment on table offer is E'@name offer_table
@omit';

create view post_view as
  SELECT 
    post.id,
    FROM mydb.post post;
comment on view post_view is E'@name posts
@uniqueKey id';

create view offer_view as
  SELECT
    offer.id,
    offer.post_id

    FROM mydb.offer offer;
comment on view offer_view is E'@name offers
@uniqueKey id
@foreignKey (post_id) references posts';
```

```js
A fatal error occurred when building the initial schema, so the process will now exit. Error details:

TypeError: Cannot read property 'namespaceId' of undefined
```

benjie:
That’s a misleading error 👍

itaied246:
Continue the previous example:

```sql
comment on view offer_view is E'@name offers
@uniqueKey id
@foreignKey (post_id) references post_view (id)';
```

Throws:

```js
A fatal error occurred when building the initial schema, so the process will now exit. Error details:

Error: No columns specified for 'mydb.post_view' (oid: 119745) and no PK found (@foreignKey (post_id) references post_view(id)).
```

benjie:
@itaied246 Thanks for these; I've added tests for them and fixed the missing parenthesis which caused the "No columns specified" error. Note you've used `@uniqueKey` but for it to work like in PostgreSQL where you can omit columns you must instead use `@primaryKey` (like you would in PostgreSQL).

PR #369, opened 2019-01-04
feat(utils): enable wrapping resolvers matching a filter

benjie:
Enables you to wrap a class of resolvers rather than only ones that have a particular name.

```js
// Example: log before and after each mutation runs
module.exports = makeWrapResolversPlugin(
  context => {
    if (context.scope.isRootMutation) {
      return { scope: context.scope };
    }
    return null;
  },
  ({ scope }) => async (resolver, user, args, context, _resolveInfo) => {
    console.log(`Mutation '${scope.fieldName}' starting with arguments:`, args);
    const result = await resolver();
    console.log(`Mutation '${scope.fieldName}' result:`, result);
    return result;
  }
);
```

### DIFF
--- a/packages/graphile-utils/README.md
+++ b/packages/graphile-utils/README.md
@@ -11,7 +11,7 @@ utilising Graphile Engine, such as the one produced by
 [PostGraphile](https://graphile.org/postgraphile).
 
 Documentation is currently available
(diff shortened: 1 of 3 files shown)


benjie:
(If reviewing, add `?w=1` to ignore whitespace changes in the tests.)

PR #371, opened 2019-01-09
fix(sql): tsrange, tstzrange, int8range and numrange now properly formatted

mathroc:
I've tried this by patching manually PgTypesPlugin.js in my node_modules and it seems to work fine

I tried to do this as a plugin as suggested but this was easier than rebuilding the whole Ranges types

I have not looked into the tests for this yet

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
@@ -1,6 +1,5 @@
 // @flow
 import type { Plugin } from "graphile-build";
-import { types as pgTypes } from "pg";
 
 import makeGraphQLJSONTypes from "../GraphQLJSON";
 
@@ -25,44 +24,6 @@ function parseInterval(str) {
   }
   return result;
 }
-
-const pgRangeParser = {
-  parse(str) {
-    const parts = str.split(",");
-    if (parts.length !== 2) {
-      throw new Error("Invalid daterange");
-    }
-
-    return {
-      start:
-        parts[0].length > 1
-          ? {
-              inclusive: parts[0][0] === "[",
-              value: parts[0].slice(1),
-            }
-          : null,
-      end:
-        parts[1].length > 1
-          ? {
-              inclusive: parts[1][parts[1].length - 1] === "]",
-              value: parts[1].slice(0, -1),
-            }
-          : null,
-    };
-  },
-
-  serialize({ start, end }) {
-    const inclusivity = {
-      true: "[]",
-      false: "()",
-    };
-
-    return [
-      start ? inclusivity[start.inclusive][0] + start.value : "[",
-      end ? end.value + inclusivity[end.inclusive][1] : "]",
-    ].join(",");
-  },
-};
 
 export default (function PgTypesPlugin(
   builder,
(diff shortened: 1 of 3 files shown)


mathroc:
Turns out values in a range can be null, so I changed that 

@benjie do you think that's okay ?


I'm trying to update the test expected result but I'm having an issue with lerna:

```
# lerna run test
info cli using local version of lerna
lerna notice cli v3.10.1
lerna info Executing command in 6 packages: "yarn run test"
lerna info run Ran npm script 'test' in 'graphile' in 0.5s:
yarn run v1.13.0
$ true
Done in 0.06s.
lerna ERR! yarn run test exited 1 in 'graphql-parse-resolve-info'
lerna ERR! yarn run test stdout:
yarn run v1.13.0
$ jest
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

lerna ERR! yarn run test stderr:
● Validation Error:

  Module ts-jest in the transform option was not found.
         <rootDir> is: /gb/packages/graphql-parse-resolve-info

  Configuration Documentation:
  https://jestjs.io/docs/configuration.html

error Command failed with exit code 1.

lerna ERR! yarn run test exited 1 in 'graphql-parse-resolve-info'
```

any idea what I should do?

benjie:
Hey @mathroc; check out the contributing document in PostGraphile, it also covers setting up Graphile Engine:

https://github.com/graphile/postgraphile/blob/master/CONTRIBUTING.md

You have to use `yarn` to install modules when developing Graphile Engine because we use "Yarn Workspaces". Maybe you need to run `yarn prepack:all` or `yarn watch` if you're already using yarn?

mathroc:
thank you, tests are now green :)

packages/postgraphile-core/__tests__/integration/__snapshots__/mutations.test.js.snap
```diff
@@ -98,7 +98,10 @@ Object {
             "inclusive": false,
             "value": 500,
           },
-          "start": null,
+          "start": Object {
+            "inclusive": false,
+            "value": null,
+          },
```
benjie: It seems to me like the meanings of these are equivalent (i.e. "there is no lower bound"); as such I think we should stick with the existing structure or risk this being a breaking change for some users.

packages/postgraphile-core/__tests__/integration/schema/__snapshots__/defaultOptions.test.js.snap
```diff
@@ -176,7 +176,7 @@ type AnIntRangeBound {
   inclusive: Boolean!
 
   \\"\\"\\"The value at one end of our range.\\"\\"\\"
-  value: AnInt!
+  value: AnInt
```
benjie: Changing an output type from non-null to nullable in GraphQL is a breaking change 😞

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
```diff
@@ -660,32 +621,17 @@ export default (function PgTypesPlugin(
...
         gqlTypeByTypeIdAndModifier[type.id][typeModifierKey] = Range;
         gqlInputTypeByTypeIdAndModifier[type.id][typeModifierKey] = RangeInput;
+        if (pgTweaksByTypeIdAndModifer[type.id] === undefined) {
+          pgTweaksByTypeIdAndModifer[type.id] = {};
+        }
+        pgTweaksByTypeIdAndModifer[type.id][
+          typeModifierKey
+        ] = fragment => sql.fragment`case when (${fragment}) is null then null else json_build_object(
+          'start', json_build_object('value', lower(${fragment}), 'inclusive', lower_inc(${fragment})),
+          'end', json_build_object('value', upper(${fragment}), 'inclusive', upper_inc(${fragment}))
```
benjie: To avoid the breaking schema changes this PR introduces, can we instead set the start/end null in SQL:

```sql
          'start', case when lower(${fragment}) is null then null else json_build_object('value', lower(${fragment}), 'inclusive', lower_inc(${fragment})) end,
          'end', case when upper(${fragment}) is null then null else json_build_object('value', upper(${fragment}), 'inclusive', upper_inc(${fragment})) end
```

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
```diff
@@ -26,44 +25,6 @@ function parseInterval(str) {
...
-    const inclusivity = {
-      true: "[]",
-      false: "()",
-    };
-
-    return [
-      start ? inclusivity[start.inclusive][0] + start.value : "[",
-      end ? end.value + inclusivity[end.inclusive][1] : "]",
-    ].join(",");
-  },
```
benjie: We seem to have lost the serialize capability for ranges. Do we not have any tests for this?!

benjie: Oh, whoops, I replaced this in the mapper but never deleted the dead code. Fine 👍

packages/postgraphile-core/__tests__/integration/__snapshots__/mutations.test.js.snap
mathroc: ahh much better, I didn't notice the whole Bound the was already nullable

mathroc:
I reverted the type change I made earlier, it's backward compatible now :)

benjie:
screenshot 2019-01-10 13 30 52

Lovely lovely lovely!

I'll take a better look at this later, but it looks mergeable 👍 Thanks @mathroc!

benjie:
@mathroc Title still says WIP; can you confirm you're happy with it?

mathroc:
yes I'm happy with it :)

```sql
create table test (
 num numrange default null,
 int8 int8range default null
);

insert into test values (
  int8range(-9223372036854775808, 9223372036854775807)
  numrange(-922337203685477580892233720368547758089223372036854775808, 922337203685477580792233720368547758079223372036854775807), 
);
```

mathroc:
ps: the range bound value type for the bigint is BigInt and BigFloat for the numeric

benjie:
Both of those are truncating; that's fine I'll fix it - I think the issue probably already existed.

mathroc:
ah yes, the bigint have the same length but the last digits are not the same !

I don't really know how this should be solved so yeah I'll let you handle that part, thank you

benjie:
No probs; I had to do it for JWT today already, so it's still fresh 😂

benjie:
Pushed - we just needed to add the tweaks to the relevant values 👍

benjie:
Thanks!

benjie:
@mathroc If you get a couple minutes could you add tests for tsrange and tstzrange? I just realised they were the reason for the PR and I’ve not tested them!

mathroc:
done

packages/postgraphile-core/__tests__/integration/queries-ranges.test.js
```diff
@@ -0,0 +1,124 @@
...
+      null,
+      {
+        pgClient: pgClient,
+      },
+      { id: row.id }
+    );
+    expect(result.errors).toBeFalsy();
+    expect(result.data.rangeTestById.tstz).toEqual({
+      start: {
+        value: "2019-01-10T17:45:56.356022-04:00",
```
benjie: This should be zulu time, I think... This might suggest we’re not properly tweaking timestamptz.

benjie: Nope, apparently this is what to_json does. It’s valid ISO8601 so I’m cool with it 👍

PR #379, opened 2019-01-12
feat(indexes): introspect index types

mlipscombe:
Adds an `indexType` property to introspected indexes, and removes the filter on indexes without expressions.  This is to be able to introspect ZomboDB indexes.

All tests pass on pg 9.4, 9.6, 10 and 11.


packages/graphile-build-pg/src/plugins/introspectionQuery.js
```diff
@@ -379,13 +380,14 @@ with
       pg_catalog.pg_index as idx
       inner join pg_catalog.pg_class idx_more on (idx.indexrelid = idx_more.oid)
       inner join class on (idx.indrelid = class.id)
+      inner join pg_class as idxclass on (idx.indexrelid = idxclass.oid)
```
benjie: This is a duplicate join, please remove

packages/graphile-build-pg/src/plugins/introspectionQuery.js
```diff
@@ -379,13 +380,14 @@ with
...
       inner join pg_catalog.pg_namespace as nsp on (nsp.oid = idx_more.relnamespace)
+      inner join pg_am as am on (am.oid = idxclass.relam)
```
benjie:
```suggestion
      inner join pg_catalog.pg_am as am on (am.oid = idxclass.relam)
```

packages/graphile-build-pg/src/plugins/introspectionQuery.js
```diff
@@ -379,13 +380,14 @@ with
...
       left join pg_catalog.pg_description as dsc on dsc.objoid = idx.indexrelid and dsc.objsubid = 0 and dsc.classoid = 'pg_catalog.pg_class'::regclass
     where
       idx.indislive is not false and
       idx.indisexclusion is not true and -- exclusion index
       idx.indcheckxmin is not true and -- always valid?
-      not (string_to_array(idx.indkey::text, ' ')::int2[] operator(pg_catalog.@>) ARRAY[0::int2]) and -- no expressions
```
benjie: Please add something to the tests to ensure this doesn't cause issues - currently our test suite does not contain indexes with expressions so we cannot be sure this doesn't cause issues.

mlipscombe:
The changes you've requested have been made.  I think this should be good to go now.

benjie:
🎉 Thanks @mlipscombe

PR #380, opened 2019-01-17
Update introspection query URL in introspectionQuery.js

petetnt:
This simple PR updates the introspection query URL in introspectionQuery.js

The file was at some point inlined from explicitly loaded `introspection-query.sql` to `introspectionQuery.js` but the URL was never changed: the URL shows up when errors happen during introspections making it a bit confusing for the end users.


benjie:
Thanks @petetnt

PR #395, opened 2019-02-04
fix(hstore): include explicit namespace when casting

### LINKED ISSUE #394: hstore mutations fail if the extension is installed to a schema
```sql
create schema ext;
create extension hstore with schema ext;

create schema app_public;

create table app_public.foo (
  id serial primary key,
  key_value ext.hstore
);
```

Mutation:
```graphql
mutation MyMutation {
  createFoo(input: {
    foo: {
      id: 1,
      keyValue: {
        mykey: "myvalue"
      }
    }
  }) {
    foo {
      nodeId
    }
  }
}
```

Response:
```json
{
  "errors": [
    {
      "message": "type \"hstore\" does not exist",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "createFoo"
      ]
    }
  ],
  "data": {
    "createFoo": null
  }
}
```
I'll be submitting a patch shortly.

mattbretl:
Fixes #394


benjie:
Great spot, simple fix - thanks.

(If you discover anywhere else we should be prefixing namespaces, except for operators, please do let me know.)

PR #401, opened 2019-02-23
fix: input field should be nullable if domain has default

### LINKED ISSUE #398: Add test for domain with default values
If having domains like

```
CREATE DOMAIN tscreated AS
  timestamptz(6)
  NOT NULL
  DEFAULT (('now'::text)::timestamp(6) with time zone);

CREATE DOMAIN tsupdated AS
  timestamptz(6)
  NOT NULL
  DEFAULT (('now'::text)::timestamp(6) with time zone);
```

and then a simple table 

```
CREATE TABLE post (
  post_id           SERIAL PRIMARY KEY,
  body_text       text NULL,
  created          tscreated,
  updated         tsupdated
);
```

I get the error `Field value.post.created of required type Tscreated! was not provided.` on mutation

```
mutation ($input: CreatePostInput!) {
  createPost (input: $input) {
    result: post {
      ...postFragment
    }
  }
}
```

Domains with default values should be nullable

mattbretl:
Fixes #398


benjie:
Thanks Matt!

PR #416, opened 2019-03-08
feat(plugins): hook priority system

benjie:
The diff in this PR is massively misleading. Using `?w=1` may help a little, but I recommend checking it out locally and using:

```
git diff master -w --word-diff-regex=[^[:space:]]
```

That'll hide **all** whitespace differences, not just on a line-by-line basis.

This should not affect the existing hook order; but it does allow plugins to register hooks *before* other hooks.


ab-pm:
Oh, this does a topological sort on the registered hook functions, so that the order of plugins is becoming less relevant? Cool!
Please document this on the *Build object* and *Hooks* pages.

benjie:
It's not ready for that yet; it's still experimental.

PR #424, opened 2019-03-18
fix(utils): fix makeExtendSchemaPlugin @pgQuery directive

benjie:
#413  introduced this feature, but it only worked when the connection and the parent type were from the same table (?!).

### DIFF
--- a/packages/graphile-utils/src/makeExtendSchemaPlugin.ts
+++ b/packages/graphile-utils/src/makeExtendSchemaPlugin.ts
@@ -732,7 +732,7 @@ function getFields<TSource>(
               );
             }
           }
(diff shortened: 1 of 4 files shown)


packages/graphile-utils/src/makeExtendSchemaPlugin.ts
```diff
@@ -732,7 +732,7 @@ function getFields<TSource>(
               );
             }
           }
-          if (directives.pgQuery && table && pgIntrospection === table) {
+          if (directives.pgQuery && table) {
```
benjie: This one line is the actual fix; the rest are just tests.

PR #435, opened 2019-04-10
fix(pg-pubsub): module augmentation

cmanou:
When merging declarations, the specified module path must exactly match the path to the actual module. Since postgraphile exports the `PostGraphileOptions` from the interfaces file the path needs to be updated to reflect this, similarly with graphile-build `Options`


benjie:
Explicit import/export seems to fix this issue, as you mentioned, so I'm opting for that instead. Thanks for bringing this to my attention.

PR #451, opened 2019-04-18
fix(live): fix live for relations with multiple keys

purge:


packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -264,8 +264,7 @@ class QueryBuilder {
                   allRequirements[key]
                 }`
             ),
-
-            ", "
+            ""
```
benjie: This is the fix.

packages/postgraphile-core/__tests__/integration/live_helpers.js
```diff
@@ -46,6 +46,10 @@ exports.liveTest = (query, variables, cb) => {
     cb = variables;
     variables = null;
   }
+
+  const errors = validate(schema, query);
+  if (errors && errors.length) throw errors[0];
```
benjie: 😳 Oops.

PR #453, opened 2019-04-19
Fix typo in big-int definition.

nilkanthjp:
Not sure if this is where this description originates.


benjie:
Need to update the test snapshots. I can do this next week 👍

nilkanthjp:
Thanks, @benjie!

Issue #191, opened 2018-03-02
Custom query format (not GraphQL). Is it possible?


KleinMaximus:
Is it possible to use graphile-builder without graphql query, but with some other syntax?
For example, to create some plugin that will convert my JSON-request format to the desired form?

benjie:
graphile-build's purpose is to build a GraphQL schema. I don't quite understand what use case you're asking for, but if it's outside the scope of GraphQL it's outside the scope of the project.

Perhaps you could expand with exactly what you want to do (maybe with some pseudocode) and we could give some pointers?

PR #467, opened 2019-05-29
feat(postgraphile): ignore schema and rely on search_path

michaelbridge:
Per [graphile/postgraphile #191](https://github.com/graphile/postgraphile/issues/191), creates an `ignoreSchemas` (default: `false`) configuration option to disable automatic prepending of the schema name to the table name.  In the case of multi-tenant applications (where there's a schema per tenant), `search_path` can instead be supplied via the callback to `pgSettings` based on the inbound JWT.

I've sought to accomplish this with minimal changes to the pre-existing codebase, however I'm open to alternative approaches if desired.

### DIFF
--- a/packages/graphile-build-pg/src/QueryBuilder.js
+++ b/packages/graphile-build-pg/src/QueryBuilder.js
@@ -43,6 +43,7 @@ type CursorComparator = (val: CursorValue, isAfter: boolean) => void;
 
 export type QueryBuilderOptions = {
   supportsJSONB?: boolean, // Defaults to true
+  ignoreSchemas?: boolean, // Defaults to false
 };
 
 class QueryBuilder {
@@ -50,6 +51,7 @@ class QueryBuilder {
   context: GraphQLContext;
   rootValue: any; // eslint-disable-line flowtype/no-weak-types
   supportsJSONB: boolean;
+  ignoreSchemas: boolean;
   locks: {
     [string]: false | true | string,
   };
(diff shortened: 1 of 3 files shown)


packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -798,7 +804,11 @@ order by (row_number() over (partition by 1)) desc`;
     } else if (type === "from") {
       if (this.data.from) {
         const f = this.data.from;
-        this.compiledData.from = [callIfNecessary(f[0], context), f[1]];
+        let _f = [callIfNecessary(f[0], context), f[1]];
+        if (this.ignoreSchemas) {
+          _f[0] = { ..._f[0], names: _f[0].names.slice(-1) };
+        }
+        this.compiledData.from = _f;
```
benjie: This assumes the from is an identifier; sometimes it’s a subquery or function call or similar. I don’t think this approach will work in general.

michaelbridge:
Thanks, benjie.  Prior to investigating core, I had explored whether this might be possible via a plugin but wasn't able to devise a solution.  I'll take another look to confirm (and would certainly appreciate any creative thoughts here).  Happy to formalize/release any plugin-based solution if it turns out to be possible.

Alternatively, I'm willing to expend effort on producing a PR as I can't wait 2-3 months for this functionality.  Candidly, though, if the effort level is too high I may need to stick with a fork.

michaelbridge:
Investigated plugin route in detail and have not devised a solution.  I've updated this PR to account properly for `type === 'IDENTIFIER`, however I understand this may not be satisfactory to you.  If not, we can close as I don't think I'll be able to devote substantial resources to a major PR.

benjie:
On the horizon is some unrelated work that happens to make doing this easier. I'm going to try and address this at that time, sorry I can't commit to a date, but I'm going to leave this PR open (and open PRs/issues nag at my soul) so that should encourage me to deal with it sooner rather than later.

benjie:
Actually since #191 is open, I'm going to close this as a duplicate. I appreciate your work on it, but I think it needs fixing in a much more invasive way, and I don't think I can give solid guidance on it right now. Make sure you subscribe to #191 so you get notified of status changes 👍

Issue #471, opened 2019-06-13
Bad pluralizations for special pluralize cases


jgzuke:
For any [special cases](https://github.com/blakeembrey/pluralize/blob/master/pluralize.js#L273) in pluralize, if your table name contains more than one word they will get pluralized incorrectly. For example the table `species` would show up in the schema as singular type `Species`, but the table `plant_species` would show up as type `PlantSpecy`.

benjie:
Any bad pluralisations can be fixed by overriding the pluralize/singularize inflectors using makeAddInflectorsPlugin: https://graphile.org/postgraphile/make-add-inflectors-plugin

You may also be interested in the `distinctPluralize` feature I added to `pg-simplify-inflector`: https://github.com/graphile-contrib/pg-simplify-inflector/blob/94cb288beefef876c91a9f4ee7f0ab8999a55186/index.js#L51-L70

Unfortunately this cannot be fixed in core as it will break peoples existing schemas. Thank you for raising the issue nonetheless 👍

ab-pm:
I have the same problem (with a table name containing `series`). How exactly can I use that `distinctPluralize` method? Do I need to skip the `simplify-inflector` plugin and write my own, or should I add another plugin that somehow hooks into the `simplify-inflector`?

benjie:
Just add another inflector plugin after simplify that overwrites `distinctPluralize`. All the inflector plugins just overwrite (or add) individual inflection methods, that's why they use `this` so they can refer to the very latest version (after all the overwrites).

ab-pm:
Thanks, looks like I need to get more familiar with the plugin system. I thought [hooks](https://www.graphile.org/graphile-build/hooks/) were just called when schema entities are created, but didn't really understand what the `extend` stuff does - so the plugins are ran on a builder for an object before it is actually built (and thereby can overwrite each others methods)? But who then calls these methods in the end, where is that documented? Especially with [`makeAddInflectorsPlugin`](https://github.com/graphile/graphile-engine/blob/master/packages/graphile-utils/src/makeAddInflectorsPlugin.ts) I am lost, as the `"inflection"` hook is not documented anywhere…
(Sorry for getting off-topic on this thread, I can open an other question issue if you want).

benjie:
> I thought hooks were just called when schema entities are created

Correct.

>  didn't really understand what the extend stuff does

[...]

PR #472, opened 2019-06-13
Run pluralize on only the last segment of a name

jgzuke:
My fix for https://github.com/graphile/graphile-engine/issues/471. Not sure if this is how you would want to resolve this but thought I would share since this solution was quick.

Switched `makeNewBuild` to use pluralize from utils and created in `formatLastSegment ` in the style of `formatInsideUnderscores`. Now `pluralize` and `singularize` will only be run on the final segment of a word, so special cased words will return proper results.

I believe this may increase the number of naming conflicts as many words would now pluralize to themselves, for example 'species'.

### DIFF
--- a/packages/graphile-build/src/utils.js
+++ b/packages/graphile-build/src/utils.js
@@ -32,11 +32,22 @@ export const formatInsideUnderscores = (fn: (input: string) => string) => (
   return `${start}${fn(middle)}${end}`;
 };
 
+export const formatLastSegment = (fn: (input: string) => string) => (
+  str: string
+) => {
+  const matches = str.match(/^([\s\S]*?)([A-Z]?[a-z0-9]*)$/);
+  if (!matches) {
+    throw new Error("Impossible?"); // Satiate Flow
+  }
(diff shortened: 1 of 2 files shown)


packages/graphile-build/src/utils.js
```diff
@@ -32,11 +32,22 @@ export const formatInsideUnderscores = (fn: (input: string) => string) => (
   return `${start}${fn(middle)}${end}`;
 };
 
+export const formatLastSegment = (fn: (input: string) => string) => (
+  str: string
+) => {
+  const matches = str.match(/^([\s\S]*?)([A-Z]?[a-z0-9]*)$/);
```
benjie: I think this regexp might be vulnerable to ReDoS. An alternative approach might be:

```js
const matches = str.match(/[A-Z_][a-z0-9]*$/);
const prefix = matches ? str.substr(0, matches.index) : "";
const word = matches ? matches[0].replace(/^_/, '') : str;
```

benjie:
Closing this since it cannot be released as part of v4 😞

PR #475, opened 2019-06-17
Warning has typo in command line option suggestion

angelosarto:
The "Failed to setup watch fixtures in Postgres database" has been corrected to suggest that the command line option is --owner-connection instead of --owner-connection-string.


angelosarto:
Super small correction in the help message -- just wanted to throw it into a PR before I forgot since it just happened and I blindly copied the --option and wondered why it didn't work!

PR #476, opened 2019-06-21
Add single relation by rest of Composite key

jgzuke:
Follow up for https://github.com/graphile/postgraphile/issues/1099

Exposed a single backwards relation when another object has reference to part of our primary key.
For example, the following table
```sql
CREATE TABLE foo("id1" SMALLINT NOT NULL);
CREATE TABLE foo_names
(
    "ex_foo_id1" SMALLINT NOT NULL REFERENCES foo(id1),
    "id2" SMALLINT NOT NULL,
    PRIMARY KEY(ex_foo_id1, id2)
);
```
would have relation
```gql
{
  fooById1(id1: 1) {
    id1
    fooNameByExFooId1AndId2(id2: 2) {
      exFooId1
      id2
    }
  }
}
```

I wasnt sure where I could add tests for this

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.js
@@ -25,6 +25,8 @@ export default (function PgBackwardRelationPlugin(
         extend,
         getTypeByName,
         pgGetGqlTypeByTypeIdAndModifier,
+        pgGetGqlInputTypeByTypeIdAndModifier,
+        gql2pg,
         pgIntrospectionResultsByKind: introspectionResultsByKind,
         pgSql: sql,
         getSafeAliasFromResolveInfo,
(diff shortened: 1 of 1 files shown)


packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.js
```diff
@@ -234,6 +236,97 @@ export default (function PgBackwardRelationPlugin(
...
+                            innerQueryBuilder.where(sql.fragment`${tableAlias}.${sqlIdentifier} = ${gql2pg(parsedResolveInfoFragment.args[columnName], type, typeModifier)}`);
+                          });
+                        }, queryBuilder.context, queryBuilder.rootValue);
+                        return sql.fragment`(${query})`;
+                      }, getSafeAliasFromAlias(parsedResolveInfoFragment.alias));
+                    }
+                  };
+                });
+                return {
+                  description: constraint.tags.backwardDescription || `Reads a single \`${tableTypeName}\` that is related to this \`${foreignTableTypeName}\`.`,
```
jgzuke: I wasn't sure what the proper description should be

packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.js
```diff
@@ -234,6 +236,97 @@ export default (function PgBackwardRelationPlugin(
...
+                      liveRecord("pg", table, record.__identifiers);
+                    }
+                    return record;
+                  }
+                };
+              }, {
+                pgFieldIntrospection: table,
+                isPgBackwardSingleRelationField: true
+              })
+            }, `Backward relation (single) for ${describePgEntity(constraint)}. To rename this relation with smart comments:\n\n  ${sqlCommentByAddingTags(constraint, {
```
jgzuke: Same as above, unsure what this should be

benjie:
My apologies on the radio silence; my last 6 weeks have been incredibly busy, but I'm back to service as normal now!

If I understand correctly, a concrete use case of this it might be like this:

```sql
create table organizations (
  id serial primary key,
  name text
);

create table members (
  organization_id int,
  member_id int,
  name text,
  primary key (organization_id, member_id)
);
```

So that you can do a query like:

```graphql
{
  organization(id: 7) {
    member(id: 23) {
      ...
    }
  }
}
```
---

Another alternative database for this could be:

```sql
create table organizations (
  id serial primary key,
  name text
);

create table members (
  id serial primary key,
  organization_id int,
  name text,
  unique (organization_id, id)
);
```

This should work in the same way, IMO. So the fact it's a primary key is not important, it just needs to be a unique key.

---

In the plugins so far we have the concept of a "forward" relation (that's where you reference something directly, e.g. you have a field `author_id` that points to author) and a "backward" relation (that's where you find everything that points to you). This one is like a "partial forward relation" - you have _part_ of the forward pointing ID.

We have an approximation of this already, e.g. you can do:

```graphql
{
  organization(id: 7) {
    members(condition: {id: 23}, first: 1) {
      nodes {
        ...
      }
    }
  }
}
```

but of course no-one wants a connection when they're just after one record. We can use the `@simpleCollections` smart comment to reduce this specific relation to a list, but then we lose the connection when we want multiple records.

I therefore see the desire for this.

However, I worry that there's going to be a lot of places in a system that has lots of multi-column unique constraints where this will apply, and thus it will have a lot of matches.

For now, I advise that you make this an external plugin (if you like, you can host it under github.com/graphile-contrib) that people can use should they want to. I definitely see quite a few people wanting this, but equally I'm not sure I want it in core.

I hope this makes sense. Let me know if you need help turning this into an external plugin.

benjie:
I'm going to close this, but please keep discussing it below :+1:

PR #479, opened 2019-06-26
feat: allow passing readCache as an object

garcianavalon:
*This is still a Work In Progress, do not merge :)*

This PR allows passing a `readCache object` to `getPostGraphileBuilder`. The functionality is backwards compatible, passing a `readCache` string with a path to a file will still work on the same way.

### DIFF
--- a/packages/postgraphile-core/__tests__/unit/read-cache.test.js
+++ b/packages/postgraphile-core/__tests__/unit/read-cache.test.js
@@ -87,6 +87,7 @@ describe("When readCache is String", () => {
     } catch (e) {
       error = e;
     }
+    expect(error).toBeDefined();
     expect(error).toMatchSnapshot();
     expect(fs.readFile).toHaveBeenCalledTimes(1);
   });
@@ -124,6 +125,7 @@ describe("when readCache is not String or Object, getPostGraphileBuilder should 
     } catch (e) {
       error = e;
     }
+    expect(error).toBeDefined();
     expect(error).toMatchSnapshot();
   });
   test("when its Array", async () => {
@@ -136,6 +138,7 @@ describe("when readCache is not String or Object, getPostGraphileBuilder should 
     } catch (e) {
       error = e;
     }
+    expect(error).toBeDefined();
     expect(error).toMatchSnapshot();
   });
   test("when its Number", async () => {
@@ -148,6 +151,7 @@ describe("when readCache is not String or Object, getPostGraphileBuilder should 
     } catch (e) {
       error = e;
     }
+    expect(error).toBeDefined();
     expect(error).toMatchSnapshot();
   });
 });
--- a/packages/postgraphile-core/__tests__/unit/read-cache.test.js
+++ b/packages/postgraphile-core/__tests__/unit/read-cache.test.js
@@ -11,11 +11,16 @@ beforeEach(() => {
 });
 
 /*
-
 Current behaviour of getPostGraphileBuilder and readCache option:
 
+If readCache is String: 
 1. cacheString is loaded by reading the file defined by readCache
 2. memoizeCache is set by JSON.parse(cacheString)
+
+If readCache is Object:
+1. memoizeCache is set as readCache directly
+
+Then
 3. persistentMemoizeWithKey is a function created dynamically that will use memoizeCache
 4. getBuilder(imported from graphile-build) is called, passing persistentMemoizeWithKey as one of the arguments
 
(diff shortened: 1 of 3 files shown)


garcianavalon:
@benjie please take a look when you have time to see if the functionality is what you required. If thats ok, I'll write the tests

benjie:
Looking good 👍

packages/postgraphile-core/src/index.ts
```diff
@@ -281,6 +281,8 @@ const getPostGraphileBuilder = async (
         `Failed to parse cache file '${readCache}', perhaps it is corrupted? ${e}`
       );
     }
+  } else if (readCache && typeof readCache === "object") {
+    memoizeCache = readCache;
```
benjie:
```suggestion
    memoizeCache = readCache;
  } else {
    throw new Error(`'readCache' not understood; expected string or object, but received '${typeof readCache}'`);
```

garcianavalon:
Thanks for the suggestion, added it.

I expect to finish tests by this week

benjie:
Great :+1:

garcianavalon:
Hi @benjie during the weekend I'll get back to this now that the test PR is merged. 

At some point, travis started failing. Checking at the logs, looks like its a linter error. I haven't seen any specifics about linting on the CONTRIBUTING docs, I assumed it would be some configuration stored in the repo. Is there anything else I need to set up on my environment regarding linters?

benjie:
@garcianavalon Good spot; we should add a `yarn prettier:fix` command to root. You can run prettier against the relevant file via

```
npx prettier --write packages/postgraphile-core/src/index.ts
```

My editors (Vim and VSCode) automatically fix any lint errors on save so I don't think about it. Please feel free to open a dedicated issue about this :+1:

garcianavalon:
@benjie I think this PR is ready (removing WIP)

- Fixed an error that caused getPostGraphileBuilder to throw an error if cache was null or undefined
- Pulled master for the readCache tests from the other PR. All tests run successfully
- Added new tests to verify object input and error throwing with invalid input
- Refactored test into describe blocks for readability (I had to update the snapshots due to the change in test name, in case you wonder why they appear as modified. I made sure to run tests *before* changing them though)

I'm waiting for good ol' travis to finish running the pipeline, but I think the linter errors have gone away on their own. I also use VSCode and is also linting on save automagically, an error might have sliped in when adding suggestions I guess.

packages/postgraphile-core/__tests__/unit/read-cache.test.js
```diff
@@ -67,22 +45,109 @@ test("when no readCache flag, persistentMemoizeWithKey should be undefined", asy
...
+    // call our method with invalid readCache value and check error
+    let error;
+    try {
+      await getPostGraphileBuilder({}, [], {
+        readCache: [],
+      });
+    } catch (e) {
+      error = e;
+    }
+    expect(error).toMatchSnapshot();
```
benjie: This isn't actually throwing - the error is `undefined`. You might want to add `expect(error).toBeTruthy()` to each of these tests, snapshots should only be used for looking for changes over time (e.g. error message text, new error properties, etc) rather than asserting behaviour.

packages/postgraphile-core/src/index.ts
```diff
@@ -265,22 +265,35 @@ export const getPostGraphileBuilder = async (
...
-    } catch (e) {
+      );
+      try {
+        memoizeCache = JSON.parse(cacheString);
+      } catch (e) {
+        throw new Error(
+          `Failed to parse cache file '${readCache}', perhaps it is corrupted? ${e}`
+        );
+      }
+    } else if (typeof readCache === "object") {
```
benjie:
```suggestion
    } else if (typeof readCache === "object" && !Array.isArray(readCache)) {
```

packages/postgraphile-core/__tests__/unit/read-cache.test.js
garcianavalon: you are right! I'll add that check

packages/postgraphile-core/src/index.ts
garcianavalon: I always forget how stupid js can be, thanks for the addition

garcianavalon:
> LGTM; I'm a little concerned that the cache might be modified but so long as it's not combined with watch mode that should be fine (and they shouldn't be combined... that doesn't make sense!). Tiny tweak to logic and tests required.

we could deep clone the object instead of direct assignment if you think thats a concern

benjie:
I considered it, but 🤷‍♂️ I think wait for someone to complain ;) Performance-wise avoiding a deep copy is preferable.

benjie:
The snapshots need updating (run the tests with `-u` to update the snapshots):

```
  Snapshot name: `when readCache is not String or Object, getPostGraphileBuilder should error when its Array 1`
    - Snapshot
    + Received
    - undefined
    + [Error: 'readCache' not understood; expected string or object, but received 'object']
      137 |       error = e;
      138 |     }
    > 139 |     expect(error).toMatchSnapshot();
          |                   ^
      140 |   });
      141 |   test("when its Number", async () => {
      142 |     // call our method with invalid readCache value and check error
      at Object.toMatchSnapshot (__tests__/unit/read-cache.test.js:139:19)
```

garcianavalon:
sorry @benjie I got a job spike and I've been quite busy. I expect to finish it this week

benjie:
No worries :+1:

packages/postgraphile-core/__tests__/unit/read-cache.test.js
garcianavalon: done

garcianavalon:
@benjie updated snapshots, added assert to verify that errors are not  `undefined`

packages/postgraphile-core/src/index.ts
```diff
@@ -265,22 +265,35 @@ export const getPostGraphileBuilder = async (
...
+        throw new Error(
+          `Failed to parse cache file '${readCache}', perhaps it is corrupted? ${e}`
+        );
+      }
+    } else if (typeof readCache === "object" && !Array.isArray(readCache)) {
+      memoizeCache = readCache;
+    } else {
       throw new Error(
-        `Failed to parse cache file '${readCache}', perhaps it is corrupted? ${e}`
+        `'readCache' not understood; expected string or object, but received '${typeof readCache}'`
```
benjie:
```suggestion
        `'readCache' not understood; expected string or object, but received '${Array.isArray(readCache) ? 'array' : typeof readCache}'`
```

garcianavalon:
Applied your suggestion, updated snapshots 👍

benjie:
Great; thanks!

CI is failing because one of the files doesn't pass the "prettier" check. It seems we don't have a `lint:fix` command :flushed: Can you run something like:

```
prettier --write packages/postgraphile-core/src/index.ts
```

please? Sorry :smile:

I should really make this easier for people to comply with.

garcianavalon:
Haha sure no problem!

garcianavalon:
Done! lets see if travis likes it now :D

benjie:
:tada: CI passes :+1:

benjie:
Thanks @garcianavalon; merged! :tada:

garcianavalon:
💃 awesome!

PR #485, opened 2019-07-04
Update error message for watch

djake:
Hello, this is a first PR for this project. I hope you don't mind I skipped the "ask first" instruction because this is a minor change to an error message, so simply typing this description was more time consuming than the actual PR.

While trying to setup new roles, I broke watching and got an error like:
```
Failed to setup watch fixtures in Postgres database ️️⚠️
This is likely because the PostgreSQL user in the connection string does not have sufficient privileges; you can solve this by passing the 'owner' connection string via '--owner-connection' / 'ownerConnectionString'. If the fixtures already exist, the watch functionality may still work.
Enable DEBUG='graphile-build-pg' to see the error
```

I am using the CLI and passing configuration via the options prop in `.postgraphilerc.js`. Per this message, I added `options.ownerConnectionString`. It appears that it should be options.ownerConnection. My guess is the existing language is meant for `library.ownerConnection`. Anyhow, I thought this PR might provide some clarity for any future newbie who stumbles on it.

If this is unwelcome or simply reflects some user error on my part, feel free to close. If you think it might be preferable to use the same property name in both cases or some other alternative, I'm happy to try that instead, I would like to contribute in some way as I make use of this project.

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
@@ -888,7 +888,7 @@ export default (async function PgIntrospectionPlugin(
                     );
                     console.warn(
                       chalk.yellow(
-                        "This is likely because the PostgreSQL user in the connection string does not have sufficient privileges; you can solve this by passing the 'owner' connection string via '--owner-connection' / 'ownerConnectionString'. If the fixtures already exist, the watch functionality may still work."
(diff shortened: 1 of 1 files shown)


benjie:
Hi @djake; thanks for your PR! The "ask first" is just there to help avoid people getting frustrated if their hard work were to be rejected.

You're absolutely right that the CLI option is `ownerConnection`; everything in `.postgraphilerc.js` is a camelCase'd version of the CLI arguments (in this case `--owner-connection`) rather than library options.

I'm very unhappy with this state of affairs; so am planning to replace it:

https://github.com/graphile/postgraphile/issues/705

As such, I'd rather not merge this PR because I'd rather it was not necessary at all. I'm going to leave it open to nag me to address the underlying problem of postgraphilerc.

Thanks again!

benjie:
On second thoughts I'm closing it. I know the RC needs dealing with :+1:

PR #487, opened 2019-07-19
Replace Object.assign() with object spread operator

singingwolfboy:
The [object spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals) is shorter and clearer to read than `Object.assign()`. This pull request changes several usages of `Object.assign()` to the object spread operator.

Note that there are several other usages of `Object.assign()` in the codebase that I did _not_ change in this pull request. It looks like the code is sometimes doing some fancy stuff with object prototypes, and I didn't want to risk accidentally breaking something. This pull request only changes code that is clearly expecting a plain object literal out of the result of `Object.assign()`, which makes them safe to rewrite using the object spread operator.

### DIFF
--- a/packages/graphile-build-pg/src/inflections.js
+++ b/packages/graphile-build-pg/src/inflections.js
@@ -86,270 +86,260 @@ export const newInflector = (
   }
 
   return deprecateEverything(
-    preventEmptyResult(
-      Object.assign(
-        {
-          pluralize,
-          argument(name: ?string, index: number) {
-            return camelCase(name || `arg${index}`);
-          },
-          orderByType(typeName: string) {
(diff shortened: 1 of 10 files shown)


benjie:
Nice; this was all written before I decided that Node 8.6 was going to be the minimum required version for PostGraphile v4. I believe that the spread operator is also faster than `Object.assign` so this looks good. It's going to take me a few days to get around to reviewing this properly, but it looks good from a quick scan :+1:

benjie:
Travis has failed due to temporary network issues. Closing and re-opening to trigger rebuild.

PR #488, opened 2019-07-19
Make sure the name property is on the field spec

singingwolfboy:
I noticed that some fields didn't have the `name` field properly populated on the spec -- which is ironic, considering that `fieldWithHooks` must be called with a `fieldName` argument. This pull request uses that argument to set that name on the spec, although other specs can override it as necessary.

### DIFF
--- a/packages/graphile-build/src/makeNewBuild.js
+++ b/packages/graphile-build/src/makeNewBuild.js
@@ -641,7 +641,7 @@ export default function makeNewBuild(builder: SchemaBuilder): { ...Build } {
                   `|${getNameFromType(Self)}.fields.${fieldName}`
                 );
                 newSpec.args = newSpec.args || {};
-                newSpec = Object.assign({}, newSpec, {
+                newSpec = Object.assign({}, { name: fieldName }, newSpec, {
                   args: builder.applyHooks(
                     this,
                     "GraphQLObjectType:fields:field:args",


benjie:
Hey DB :smiley:  

The spec here is a [`GraphQLFieldConfig` from graphql-js](https://graphql.org/graphql-js/type/#graphqlobjecttype) which does not accept a `name` field:

```js
type GraphQLFieldConfig = {
  type: GraphQLOutputType;
  args?: GraphQLFieldConfigArgumentMap;
  resolve?: GraphQLFieldResolveFn;
  deprecationReason?: string;
  description?: ?string;
}
```

The field name is available on the context a few lines down (line 650) via `context.field.name` which will be available in args hooks.

I'm going to close this, but feel free to follow up with further questions.

singingwolfboy:
Ah! I made this pull request because I noticed that [`PgConnectionArgCondition` is trying to access `field.name` when logging a message](https://github.com/graphile/graphile-engine/blob/9fc302602c6eda46e6423aefd420c2ef7d5c49eb/packages/graphile-build-pg/src/plugins/PgConnectionArgCondition.js#L189), and I was getting `undefined`, which I assumed was a bug in the field definition. But maybe it's a bug in the logging code, instead!

benjie:
Ah, indeed a bug in the logging code. 😳

singingwolfboy:
> The field name is available on the context a few lines down (line 650) via `context.field.name` which will be available in args hooks.

I just did some more investigation, and I discovered that the `field` variable in `PgConnectionArgCondition` *is* `context.field.name`! And since that value does _not_ have a `name` field, it seems like your statement here is incorrect (unintentionally, I'm sure). Is there another way of getting the field name from a  `GraphQLObjectType:fields:field:args` hook?

benjie:
(I believe @singingwolfboy answered his own question in https://github.com/graphile/graphile-engine/pull/489 )

PR #501, opened 2019-08-09
Unit Test read cache

garcianavalon:
This is still a Work in progress, I'm opening the PR to discuss the changes.

This PR add unit tests for `readCache` option in `getPostGraphileBuilder`. See test suite for details

There are no changes to `src` files (except exporting `getPostGraphileBuilder` function) , so it should be safe to merge once accepted.

### DIFF
--- a/packages/postgraphile-core/__tests__/unit/read-cache.test.js
+++ b/packages/postgraphile-core/__tests__/unit/read-cache.test.js
@@ -62,3 +62,23 @@ test("when no readCache flag, persistentMemoizeWithKey should be undefined", asy
   } = graphileBuild.getBuilder.mock.calls[0][1];
   expect(persistentMemoizeWithKey).toBeUndefined();
 });
+
+test("when cache file has invalid content, getPostGraphileBuilder should error", async () => {
+  // mock fs.readFile to return an object
+  fs.readFile.mockImplementationOnce((path, options, cb) =>
+    cb(null, "thisisnotjson")
+  );
+
+  // call our method and check error
+
+  let error;
+  try {
+    await getPostGraphileBuilder({}, [], {
+      readCache: true,
+    });
+  } catch (e) {
+    error = e;
+  }
+  expect(error).toEqual(expect.any(Error));
+  expect(fs.readFile).toHaveBeenCalledTimes(1);
+});
--- a/packages/postgraphile-core/__tests__/unit/read-cache.test.js
+++ b/packages/postgraphile-core/__tests__/unit/read-cache.test.js
@@ -36,7 +36,7 @@ test("when cache file has empty object, persistentMemoizeWithKey should be a fun
   graphileBuild.getBuilder.mockResolvedValueOnce(expectedOuput);
   // call our method and test output
   const output = await getPostGraphileBuilder({}, [], {
-    readCache: true,
+    readCache: "path/to/cache",
   });
   expect(output).toBe(expectedOuput);
   expect(fs.readFile).toHaveBeenCalledTimes(1);
@@ -74,7 +74,7 @@ test("when cache file has invalid content, getPostGraphileBuilder should error",
   let error;
   try {
     await getPostGraphileBuilder({}, [], {
-      readCache: true,
+      readCache: "path/to/cache",
     });
   } catch (e) {
     error = e;
--- a/packages/postgraphile-core/__tests__/unit/read-cache.test.js
+++ b/packages/postgraphile-core/__tests__/unit/read-cache.test.js
@@ -34,13 +34,13 @@ test("when cache file has content, persistentMemoizeWithKey should be a valid fu
   );
 
   // mock getBuilder to fake output
-  const expectedOuput = {};
-  graphileBuild.getBuilder.mockResolvedValueOnce(expectedOuput);
+  const expectedOutput = {};
+  graphileBuild.getBuilder.mockResolvedValueOnce(expectedOutput);
   // call our method and test output
   const output = await getPostGraphileBuilder({}, [], {
     readCache: "path/to/cache",
   });
-  expect(output).toBe(expectedOuput);
+  expect(output).toBe(expectedOutput);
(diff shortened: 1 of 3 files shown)


garcianavalon:
@benjie this should cover the most trivial cases. 

The next step I can think of would be using snapshot testing to verify the function created in `persistentMemoizeWithKey` with some non-trivial input from a cache file (I can generate one or two in a sample project and use them as fixtures). Should I implement this or are we happy with the current tests?

Another option I can think of is extracting `persistentMemoizeWithKey` creation to its own module and unit test that thoroughly but I feel like that exceeds the scope of this PR and also introduces changes in `src` files which I'll rather avoid for now.

packages/postgraphile-core/__tests__/unit/read-cache.test.js
```diff
@@ -0,0 +1,84 @@
...
+
+  let error;
+  try {
+    await getPostGraphileBuilder({}, [], {
+      readCache: "path/to/cache",
+    });
+  } catch (e) {
+    error = e;
+  }
+  expect(error).toEqual(expect.any(Error));
```
benjie: Can we use an inline snapshot for this error please? I'd like to track it changing :)

packages/postgraphile-core/__tests__/unit/read-cache.test.js
```diff
@@ -0,0 +1,84 @@
...
+  fs.readFile.mockImplementationOnce((path, options, cb) => cb(null, "{}"));
+
+  // mock getBuilder to fake output
+  const expectedOuput = {};
+  graphileBuild.getBuilder.mockResolvedValueOnce(expectedOuput);
+  // call our method and test output
+  const output = await getPostGraphileBuilder({}, [], {
+    readCache: "path/to/cache",
+  });
+  expect(output).toBe(expectedOuput);
```
benjie: Since this assertion is the same as the one on line 57, rather than an empty object could we define something like `{"__test": true}` and assert that matches? Thanks! :pray:

garcianavalon: Sure, I'll take a look tomorrow

packages/postgraphile-core/__tests__/unit/read-cache.test.js
garcianavalon: Makes sense, thanks for the review

packages/postgraphile-core/__tests__/unit/read-cache.test.js
garcianavalon: Hey @benjie I've reviewed your comment but I'm not sure what you mean. The check:
```javascript
const output = await getPostGraphileBuilder({}, [], {});
expect(output).toBe(expectedOuput);
```
its not about the value of `expectedOutput` but rather the fact that we got this output. This value is coming from [here](https://github.com/graphile/graphile-engine/blob/master/packages/postgraphile-core/src/index.ts#L365). Because this is a unit test, we are mocking `getBuilder` which is from another package:
```js
  // mock getBuilder to fake output
  const expectedOuput = {};
  graphileBuild.getBuilder.mockResolvedValueOnce(expectedOuput);
```
So which value we put there is completely irrelevant. Is only about checking that  `getPostGraphileBuilder` got to the end. Given than we test the important outcome (that we pass the `persistentMemoizedWithKey` to `getBuilder`) in the next lines, we could completely remove the  `expectedOutput` check
```js
expect(graphileBuild.getBuilder).toHaveBeenCalledTimes(1);
  // check persistentMemoizeWithKey, the actual "result" of readCache flag
  const {
    persistentMemoizeWithKey,
  } = graphileBuild.getBuilder.mock.calls[0][1];
  expect(persistentMemoizeWithKey).toBeUndefined();
```
Either we remove it or we keep it as is, but changing the mocked value will not provide us with better checks (we would be testing jest mock functions hehe)

packages/postgraphile-core/__tests__/unit/read-cache.test.js
garcianavalon: done

packages/postgraphile-core/__tests__/unit/read-cache.test.js
```diff
@@ -0,0 +1,84 @@
...
+  // check persistentMemoizeWithKey, the actual "result" of readCache flag
+  const {
+    persistentMemoizeWithKey,
+  } = graphileBuild.getBuilder.mock.calls[0][1];
+  expect(typeof persistentMemoizeWithKey).toBe("function");
+});
+
+test("when no readCache flag, persistentMemoizeWithKey should be undefined", async () => {
+  // mock getBuilder to fake output
+  const expectedOuput = {};
```
benjie: Typo (missing `t`)

packages/postgraphile-core/__tests__/unit/read-cache.test.js
benjie: Sorry yeah, this was the wrong place to add this suggestion - leave this line as-is, and change line 32 to return `'{"__test": true}'` instead of `"{}"`.

We should then call the generated `persistentMemoizeWithKey` at the bottom of the test with a know value and assert that it matches the expected value based on the cache data we've fed in, e.g.:

```js
expect(persistentMemoizeWithKey('__test')).toEqual(true);
expect(() => persistentMemoizeWithKey('unknown_key')).toThrow();
```

benjie:
Please remove `WIP` from the title when you think your code's ready to merge :)

packages/postgraphile-core/__tests__/unit/read-cache.test.js
garcianavalon: aahh makes sense, it is a good check. I'll add it

packages/postgraphile-core/__tests__/unit/read-cache.test.js
garcianavalon: took me a while to find where it was...

garcianavalon:
@benjie added changed requested, ready to merge

PR #508, opened 2019-08-22
fix(introspection): corrected bitmasks for mutability checks

rudism:
closes https://github.com/graphile/postgraphile/issues/1138


benjie:
Great spot; our logic has UPDATE and INSERT the wrong way around.

CMD_UPDATE (2), CMD_INSERT (3) and CMD_DELETE (4) are defined [here](https://github.com/postgres/postgres/blob/66bde49d96a9ddacc49dcbdf1b47b5bd6e31ead5/src/include/nodes/nodes.h#L667-L678):

```c
typedef enum CmdType
{
	CMD_UNKNOWN,
	CMD_SELECT,					/* select stmt */
	CMD_UPDATE,					/* update stmt */
	CMD_INSERT,					/* insert stmt */
	CMD_DELETE,
	CMD_UTILITY,				/* cmds like create, destroy, copy, vacuum,
								 * etc. */
	CMD_NOTHING					/* dummy command for instead nothing rules
								 * with qual */
} CmdType;
```

That's used [here](https://github.com/postgres/postgres/blob/5ee190f8ec37c1bbfb3061e18304e155d600bc8e/src/backend/rewrite/rewriteHandler.c#L2698-L2703) to build the bitmap:

```c
			if (trigDesc->trig_insert_instead_row)
				events |= (1 << CMD_INSERT);
			if (trigDesc->trig_update_instead_row)
				events |= (1 << CMD_UPDATE);
			if (trigDesc->trig_delete_instead_row)
				events |= (1 << CMD_DELETE);
```

So I can confirm that INSERT is the 4th bit (00001000), UPDATE is the 3rd bit (00000100), and DELETE is the 5th bit (00010000).

PR #511, opened 2019-08-24
Remove @types/graphql dependency

none23:
When installing postgraphile, I get warnings:
```
warning postgraphile > @types/graphql@14.5.0: This is a stub types definition. graphql provides its own type definitions, so you do not need this installed.
warning postgraphile > postgraphile-core > @types/graphql@14.5.0: This is a stub types definition. graphql provides its own type definitions, so you do not need this installed.
warning postgraphile > postgraphile-core > graphile-build > @types/graphql@14.5.0: This is a stub types definition. graphql provides its own type definitions, so you do not need this installed.
warning postgraphile > postgraphile-core > graphile-build > graphql-parse-resolve-info > @types/graphql@14.5.0: This is a stub types definition. graphql provides its own type definitions, so you do not need this installed.

```
This PR, together with https://github.com/graphile/postgraphile/pull/1142 should fix them


benjie:
Thanks for this PR!

This change was only added in graphql@14.5.0 three days ago: https://github.com/graphql/graphql-js/releases/tag/v14.5.0, so we need to also update our graphql dependency so we can use the types from within it. To do so, don't edit the compatibility range (that would be a breaking change), instead delete this section:

https://github.com/graphile/graphile-engine/blob/e95f1564ed37444108511e103e4042c3c3c85929/yarn.lock#L4793-L4798

from the `yarn.lock` and then re-run yarn and it should install the latest version.

I'm happy to merge this, if it affects anyone negatively they should be able to solve it by installing an older version of @types/graphql manually.

none23:
I also modified .flowconfig to make travis-ci pass (flow 0.66 didn't like flow 0.106 syntax used by graphql-js)

PR #512, opened 2019-08-27
docs(CONTRIBUTING): standardise database setup instructions

ab-pm:
See https://github.com/graphile/postgraphile/issues/1136, https://github.com/graphile/postgraphile/pull/1145

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,8 +64,7 @@ Graphile Engine has matured into the production-ready system it is today.
 ## Development
 
 Below is a quick-start, for more detailed instructions, please [see the
-CONTRIBUTING.md doc in
-PostGraphile](https://github.com/graphile/postgraphile/blob/master/CONTRIBUTING.md).
+CONTRIBUTING.md documentation in PostGraphile](https://github.com/graphile/postgraphile/blob/master/CONTRIBUTING.md).
 
 ```bash
 yarn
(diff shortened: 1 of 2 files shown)


packages/pg-pubsub/__tests__/subs.test.ts
```diff
@@ -23,7 +23,7 @@ const init = async (options: PostGraphileOptions = {}) => {
     throw new Error("CTX wasn't torn down");
   }
   const pgPool = new Pool({
-    connectionString: process.env.TEST_DATABASE_URL || "pggql_test",
```
ab-pm: Given that all other packages require the `TEST_DATABASE_URL` to have a variable anyway, I thought it wouldn't be an issue to drop this fallback. (I also removed the "historic" name from the rest of the docs - sorry @benjie)

benjie:
Excellent work; thanks for your help on this! I’ll review it next week 👍 (ping me if I forget)

README.md
```diff
@@ -76,15 +75,9 @@ yarn watch
 `yarn watch` will keep monitoring and compiling the babel files, so open
 another terminal to run the tests:
 
-(Note: before you can run the tests, you'll need to configure your PostgreSQL
-server to support logical decoding for our live queries tests. See
-[the @graphile/lds README](packages/lds/README.md#postgresql-configuration) for how
-to enable `wal_level = logical`.)
```
benjie: Why was this paragraph removed?

ab-pm: This seemed like a duplication of information, and not to fit in the light overview ("quick start") of the README. The prerequisites necessary to run the tests are documented extensively in the linked CONTRIBUTING now, which seemed like the better place for this paragraph.

benjie: I think we should bring it back, otherwise 95% of the people who try the quickstart will fail and not know why. We could simplify it:

```
(Note: your PostgreSQL server must be
[configured for logical decoding](packages/lds/README.md#postgresql-configuration).)
```

benjie: @ab-pm Just a nudge that this PR can be merged after this small change.

ab-pm:
Should I rebase on the current master?

benjie:
I've pressed the "Update branch" button, which is fine. We squash so rebase isn't required to keep history linear.

PR #370, opened 2019-01-04
chore(types): convert to TypeScript

benjie:
This is currently the result of running this against the codebase:
https://github.com/bcherny/flow-to-typescript/pull/12

TODO:

- [x] Automated conversion on all Flow files
- [x] Run prettier on all converted files to make them match the Flow as much as possible
- [x] Delete the .js files, hopefully git recognises the .js -> .ts as basically a rename?
- [ ] Review differences, fix where appropriate
- [ ] Remove Flow from everywhere
- [ ] Add relevant TypeScript compilation stuff
- [ ] Ensure all the tests pass
- [ ] Have a hard-earned beer 🍺 

Convert script:

```bash
git checkout 686ee4e40a9f301d91cda7d42e393582fae94026 .
for I in packages/*/src/*.js packages/*/src/**/*.js; do echo $I; ./flow-to-typescript/dist/src/cli.js -i "$I" -o "${I/.js/.ts}" && rm "$I"; done
yarn prettier:fix
git add packages
```

PR #516, opened 2019-09-03
chore(types): convert to TypeScript

benjie:
This uses my latest iteration of flow2typescript, replacing #370 

(I need to switch all the build tools over to TypeScript, get rid of flow, and fix any remaining TypeScript issues, but this is an epically good start!)


benjie:
Closing in favour of https://github.com/graphile/graphile-engine/pull/573

PR #519, opened 2019-09-10
fix: remove sortable/filterable from functions with required args

mattbretl:


packages/graphile-build-pg/src/plugins/PgConditionComputedColumnPlugin.js
```diff
@@ -21,7 +21,7 @@ function getCompatibleComputedColumns(build, table) {
     const { pseudoColumnName } = computedColumnDetails;
 
     // Must have only one required argument
-    const nonOptionalArgumentsCount = proc.argDefaultsNum - proc.inputArgsCount;
+    const nonOptionalArgumentsCount = proc.inputArgsCount - proc.argDefaultsNum;
```
benjie: :interrobang: :flushed:

packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
```diff
@@ -358,6 +358,8 @@ create function a.post_computed_compound_type_array(post a.post, object c.compou
 create function a.post_computed_text_array(post a.post) returns text[] as $$ select ARRAY['str1','str2','str3']; $$ language sql stable;
 create function a.post_computed_interval_array(post a.post) returns interval[] as $$ select ARRAY[interval '12 seconds', interval '3 hours', interval '34567 seconds']; $$ language sql stable;
 create function a.post_computed_interval_set(post a.post) returns setof interval as $$ begin return next interval '12 seconds'; return next interval '3 hours'; return next interval '34567 seconds'; end; $$ language plpgsql stable;
+create function a.post_computed_with_required_arg(post a.post, i int) returns int as $$ select 1; $$ language sql stable strict;
+comment on function a.post_computed_with_required_arg(post a.post, i int) is E'@sortable\n@filterable';
```
benjie: Could you add an example with an optional argument too, please?

```suggestion
comment on function a.post_computed_with_required_arg(post a.post, i int) is E'@sortable\n@filterable';
create function a.post_computed_with_optional_arg(post a.post, i int = 1) returns int as $$ select 1; $$ language sql stable strict;
comment on function a.post_computed_with_optional_arg(post a.post, i int) is E'@sortable\n@filterable';
```

mattbretl: :+1: Done

PR #256, opened 2018-07-17
Add support for postgres inet type

syndesis:
Add support for postgres inet type

PR #520, opened 2019-09-12
feat(pg): add support for cidr and macaddr types

higherorderfunctor:
Adds support for postgres cidr and macaddr types following the advice in #256 and example of the inet type from #279.


packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
```diff
@@ -341,6 +341,14 @@ export default (function PgTypesPlugin(
...
         "An IPv4 or IPv6 host address, and optionally its subnet."
       );
+      const CidrType = stringType(
+        inflection.builtin("CidrAddress"),
+        "An IPv4 or IPv6 CIDR address."
+      );
+      const MacAddrType = stringType(
+        inflection.builtin("MacAddress"),
+        "A 6-byte MAC address."
+      );
```
benjie: I believe that cidr and macaddr currently use the `String` type; if you could make this an opt-in (non-breaking) change I'd be happy to merge it. To do so we need to do a couple of small changes:

First, this section needs to use a new flag `pgUseCustomNetworkScalars`. Add the new flag (`pgUseCustomNetworkScalars = false`) here:
https://github.com/graphile/graphile-engine/blob/e1d167e30f725707483fbea66b9bed28b80c5caf/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js#L33

Secondly, change these two defs to use the new flag: `const CidrType = pgUseCustomNetworkScalars ? stringType(...) : GraphQLString`

This would make this available using `graphileBuildOptions`. I don't think we should expand it beyond here as we'll just default it to on in V5.

higherorderfunctor:
Is there a mechanism at the stage where types are mapped to detect the postgres version, or does that occur before connecting to the database?  I considered adding in macaddr8 with this commit, but it was not available until version 10.

mattbretl:
You can safely add `macaddr8` to PgTypesPlugin the same way you've added the others. On PG 9.6 or below, that GraphQL type will never be used; no issue there.

Writing tests is a bit more complicated, but doable. We have a PG10-specific schema here, which is only used when the PG version is 10 or higher:
https://github.com/graphile/graphile-engine/blob/b0e19c73f1f78c62c4be1a2974a4f2404c538aa5/packages/postgraphile-core/__tests__/pg10.sql

If you update that file to include a table with a `macaddr8` column, it'll show up in the schema tests. We don't have any query or mutation tests that are PG10-specific, but if you get to that point (and Benjie thinks they're necessary) we can provide guidance.

mattbretl:
I lied. We do have PG10-specific mutation tests: https://github.com/graphile/graphile-engine/blob/b0e19c73f1f78c62c4be1a2974a4f2404c538aa5/packages/postgraphile-core/__tests__/fixtures/mutations/pg10.identity.graphql

As long as you prefix the `.graphql` filename with `pg10.`, it'll be skipping if the PG version is <10: https://github.com/graphile/graphile-engine/blob/b0e19c73f1f78c62c4be1a2974a4f2404c538aa5/packages/postgraphile-core/__tests__/integration/mutations.test.js#L98

PG10-specific queries haven't been implemented, but the approach would be same.

higherorderfunctor:
Updated the PR with the requested changes and rebased on top of master.  I had to make a few design decisions in the test suite to test the various schema permutations with the default option, opted-in option, and pg10 specific tests (mutations and queries).

Let me know if there is anything I can improve to get this pulled.  Thank you for your feedback.

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
higherorderfunctor: Updated the PR with the requested changes, macaddr8 type (pg10+), and rebased on top of master. I had to make a few design decisions in the test suite to test the various schema permutations with the default option, opted-in option, and pg10 specific tests (mutations and queries).

Let me know if there is anything I can improve to get this pulled. Thank you for your feedback.

higherorderfunctor:
> I lied. We do have PG10-specific _mutation_ tests: https://github.com/graphile/graphile-engine/blob/b0e19c73f1f78c62c4be1a2974a4f2404c538aa5/packages/postgraphile-core/__tests__/fixtures/mutations/pg10.identity.graphql
> 
> As long as you prefix the `.graphql` filename with `pg10.`, it'll be skipped if the PG version is <10:
> 
> https://github.com/graphile/graphile-engine/blob/b0e19c73f1f78c62c4be1a2974a4f2404c538aa5/packages/postgraphile-core/__tests__/integration/mutations.test.js#L98
> 
> PG10-specific _query_ tests haven't been implemented, but the approach would be same.

Thank you for the lead.  I updated the PR following the hint.  I did add some pg10 test queries, though it follows what is in queries.test.js currently for skipping tests/schema selection over how it was done in mutations.test.js.

packages/postgraphile-core/__tests__/integration/mutations.test.js
```diff
@@ -90,19 +110,29 @@ beforeAll(() => {
...
           console.log("Skipping test as PG version is less than 10");
           return;
         }
-        schemaToUse = pg10Schema;
+        if (fileName.startsWith("pg10.network_types.")) {
+          schemaToUse = pg10UseCustomNetworkScalarsSchema;
+        } else {
+          schemaToUse = pg10Schema;
+        }
+      } else if (fileName.startsWith("network_types")) {
```
benjie: Missing period

```suggestion
      } else if (fileName.startsWith("network_types.")) {
```

packages/postgraphile-core/__tests__/integration/queries.test.js
```diff
@@ -16,6 +16,11 @@ function readFile(filename, encoding) {
   });
 }
 
+async function getServerVersionNum(pgClient) {
+  const versionResult = await pgClient.query("show server_version_num;");
+  return parseInt(versionResult.rows[0].server_version_num, 10);
+}
```
benjie: Please move this and the other copy to helpers and import both from there.

higherorderfunctor:
Corrections made, look forward to additional feedback.

I did add a generic skip in queries.test.js for files prefixed with "pg10." if serverVersionNum < 100000.

benjie:
(AFAICT no other changes are required.)

higherorderfunctor:
> This looks great; but you seem to have forgotten to commit the `pg10.network_types.*` snapshot file?

Aren't these results baked into these files?

**packages/postgraphile-core/__tests__/integration/__snapshots__/mutations.test.js.snap** 
...
exports[`pg10.network_types.graphql 1`] = `
...

**packages/postgraphile-core/__tests__/integration/__snapshots__/queries.test.js.snap**
...
exports[`pg10.network_types.createNetwork.graphql 1`] = `
...

Anticipating what you might be looking for, I added a schema test case into integration/schema/pg10.test.js, with the new option on, to test the macaddr8 type.

I also added a completely new schema test, integration/schema/network_types.test.js, that tests with the new option on and off on the network_types schema added to the kitchen sink.

This generated the snap files I believe you are looking for?  I apologize for missing these, still learning this package/testing practices.

Do you want me to force squash this into a single commit before pulling?

benjie:
We do squash merges so the state of your git history is unimportant. In general, I'd advise that you do not rebase after someone has reviewed your code unless they (or the project) ask you to - it makes reviewing just the changes a challenge and often necessitates a full re-review.

You're right; I was looking for the schema dump as I couldn't see the new types being used - thanks for reading between the lines :)

Will take a look now.

PR #527, opened 2019-09-24
Pin prettier==1.18.2

singingwolfboy:
Prettier recommends pinning an exact version of the package, because they sometimes introduce stylistic changes in patch releases. See https://prettier.io/docs/en/install.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "lerna": "^3.13.4",
     "pg": "^7.10.0",
     "postgraphile": "^4.4.0",
-    "prettier": "^1.17.0",
+    "prettier": "1.18.2",
     "typescript": "^3.6.3"
   },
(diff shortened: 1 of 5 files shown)


benjie:
No thanks; I'm happy automatically rewriting the codebase when we upgrade prettier ─ it's normally an improvement (often significantly so, in places).

singingwolfboy:
Is there a particular reason why prettier is pinned in `pg-sql2`, but nowhere else?

benjie:
pg-sql2 was only recently brought "into the fold", so I must have overlooked this.

Actually; none of the packages should have `prettier` as a dependency since we now (as of yesterday) use root-level `eslint` for everything. Fixed in https://github.com/graphile/graphile-engine/pull/528

Issue #455, opened 2019-04-26
Schema generated by Function returning Table doesn't have totalCount and pageInfo


xialvjun:
**Function returning Table**:
![image](https://user-images.githubusercontent.com/4413712/56795926-1de32100-6844-11e9-85c7-4e8f1b2f1f43.png)

**Schema generated**:
![image](https://user-images.githubusercontent.com/4413712/56795996-40753a00-6844-11e9-9cdc-5cf7c29720cb.png)

benjie:
This issue has been open for a long time and no-one else has show interest in it, so I'm going to close it.

PR #529, opened 2019-09-25
feat(connections): expose totalCount on custom connections

mattbretl:
This implements the `totalCount` portion of this feature request: https://github.com/graphile/graphile-engine/issues/455

(Implementing pageInfo would be much tougher because we'd need a unique key on the return type. Functions are essentially a black box, so we can't determine unique keys from introspection.)


packages/graphile-build-pg/src/plugins/PgConnectionTotalCount.js
```diff
@@ -7,25 +7,18 @@ export default (function PgConnectionTotalCount(builder) {
...
         Self,
       } = context;
 
-      if (
-        !isPgRowConnectionType ||
-        !table ||
-        table.kind !== "class" ||
-        !table.namespace
-      ) {
+      if (!isPgRowConnectionType || !nodeType || !nodeType.name) {
```
benjie: I think this is a breaking change for people using `makeExtendSchemaPlugin` to add their own connections:

https://github.com/graphile/graphile-engine/blob/b5d9e99fefa0af027bfcaeae06443eff00412aea/packages/graphile-utils/src/makeExtendSchemaPlugin.ts#L721-L725

Please can you support the old `pgIntrospection` in addition to `nodeType`?

mattbretl: Good catch. Done. https://github.com/graphile/graphile-engine/pull/529/commits/a75c4d7a462df01bdb83a193eb01d55ff8698537

benjie:
Thanks Matt! :tada:

PR #532, opened 2019-10-02
fix(types): export more types inc PgIntrospectionResultByKind

hansololai:
added export PgIndex from the PgIntrospectionPlugin. 

added a PgIntrospectionResultByKind type.


packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -199,6 +199,23 @@ export type PgEntity =
   | PgExtension
   | PgIndex;
 
+export type PgIntrospectionResultsByKind = {
```
benjie: Please use this type; e.g. as the return value for the `introspect` function.

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -199,6 +199,23 @@ export type PgEntity =
...
+  __pgVersion: number,
+  attribute: PgAttribute[],
+  attributeByClassIdAndNum: { [string]: { [string]: PgAttribute } },
```
benjie: Please give these indexers names so that when we convert to TypeScript it won't just be `a` and `b`. For example this one would be:

```suggestion
  attributeByClassIdAndNum: { [classId: string]: { [num: string]: PgAttribute } },
```

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
hansololai: how do I do that? should I annotate that on the js file? or export a function overload in d.ts file?

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
hansololai: cool, thanks

hansololai:
> Nice; do you want to duplicate the `PgIntrospectionResultsByKind` to `PgIntrospectionPlugin.d.ts` and import/export it in `graphile-build-pg/src/index.dts` also?

sounds good, I added these in d.ts and index.d.ts

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
benjie: Yes; this is using flowtype so you can do it similarly to TypeScript:


```js
  async function introspect(): Promise<PgIntrospectionResultsByKind> {
```

I just figure without that the type isn't actually _used_ anywhere, so it's value is less (and it's likely to come out of sync over time).

benjie:
This is looking great :raised_hands: Just need to make sure `PgIntrospectionResultsByKind` is used in the Flow code somewhere :)

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
hansololai: it is pretty hard tho, because the `introspect` is not exported. The `d.ts` file is really used by users to know the module's exported member's type.  So I added a jsDoc annotation to the `introspect` function, now it will be checked by typescript. I also added export of the function type, and the export default function

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.d.ts
```diff
@@ -1,3 +1,5 @@
+import { SchemaBuilder, Options } from "../../../graphile-build/src";
```
benjie: We cannot do this relative require into a different package. Instead:

```suggestion
import { SchemaBuilder, Options } from "graphile-build";
```

benjie: (Apparently I cannot push this fix myself, so I'll wait on you to accept it. @ me when it's ready.)

hansololai: I see, thanks. @benjie I just committed it.

PR #533, opened 2019-10-08
chore(tests): cleaner GraphQL schema snapshots

singingwolfboy:
Same change as in https://github.com/graphile-contrib/pg-many-to-many/pull/27


packages/graphile-build/__tests__/enum.test.js
```diff
@@ -1,7 +1,16 @@
...
 
+const GraphQLSchemaSerializer = {
+  test(val) {
+    return isSchema(val);
+  },
+  serialize(schema) {
+    return printSchema(schema);
+  },
+};
+expect.addSnapshotSerializer(GraphQLSchemaSerializer);
```
benjie: Please can you put this in `packages/graphile-build/__tests__/helpers.js` and then import that instead:

```js
require("./helpers");
```

packages/pg-pubsub/__tests__/subs.test.ts
```diff
@@ -15,8 +17,21 @@ import { runQuery, TestCtx } from "./runQuery";
...
+  // so that it conforms to the shape of `pretty-format.NewPlugin`,
+  // but the Jest types don't seem to know about that type yet,
+  // so we'll name it `print` instead to satisfy the Jest types.
+  // https://github.com/facebook/jest/blob/master/packages/jest-snapshot/src/plugins.ts
+  // https://github.com/facebook/jest/blob/master/packages/pretty-format/src/types.ts#L115
+  print(schema: GraphQLSchema) {
+    return printSchema(schema);
+  },
+};
+expect.addSnapshotSerializer(GraphQLSchemaSerializer);
```
benjie: Or, more specifically, this one since it has the helpful comments :grin:

packages/pg-pubsub/__tests__/subs.test.ts
```diff
@@ -15,8 +17,21 @@ import { runQuery, TestCtx } from "./runQuery";
...
 
-const printIntrospectionResults = (json: any) =>
-  printSchema(buildClientSchema(json.data));
+const GraphQLSchemaSerializer = {
+  test(val: any) {
+    return isSchema(val);
+  },
+  // It would be nice if we could name this method `serialize`,
+  // so that it conforms to the shape of `pretty-format.NewPlugin`,
+  // but the Jest types don't seem to know about that type yet,
```
benjie: From looking at the source, it looks like they should as it's declared inside pretty-format's `.d.ts` files here:

https://unpkg.com/browse/pretty-format@24.9.0/build/types.d.ts

Is our version of pretty-format out of date? If so, try deleting it's definition in `yarn.lock` and running yarn again to get the latest. If not I'm happy to merge as-is and research more later.

packages/pg-pubsub/__tests__/subs.test.ts
```diff
@@ -78,7 +93,8 @@ describe("Middleware defaults", () => {
       {},
       async (json, _req, res) => {
         expect(res.statusCode).toEqual(200);
-        expect(printIntrospectionResults(json)).toMatchSnapshot();
+        const schema = buildClientSchema(json.data);
```
benjie: Let's add another assertion just in case
```suggestion
        expect(json.errors).toBeFalsy();
        const schema = buildClientSchema(json.data);
```

packages/graphile-build/__tests__/enum.test.js
benjie: Given this is used throughout, actually putting it in a helpers file that everyone can use would be awesome. Maybe we should add a private subpackage just for this so we can require it from the other projects?

packages/pg-pubsub/__tests__/subs.test.ts
singingwolfboy: As far as I can tell, this is happening because we depend on `@types/jest`, which is not kept perfectly in sync with the actual types in Jest. As far as I can tell, `@types/jest` was created because Jest was originally written in Flow. However, now Jest has migrated to Typescript, but `@types/jest` still exists -- and is apparently still necessary, for some reason. I tried removing `@types/jest` from our `package.json` files and then running `yarn` to uninstall it, but the tests failed to run. I got the following error:

```
__tests__/index.test.ts:6:1 - error TS2582: Cannot find name 'test'. Do you need to install type definitions for a test runner? Try `npm i @types/jest` or `npm i @types/mocha`.
```

Ironic, huh?

packages/graphile-build/__tests__/enum.test.js
singingwolfboy: I think a different package is probably the right solution here, but I'm not exactly sure what you mean by "private subpackage". Should I make a new directory under `packages` in this repo? How do I make it private? Shouldn't the versioning be independent of everything else in the `packages` directory? I've never used Lerna before, and I'm not sure how to configure it properly to make this work...

packages/pg-pubsub/__tests__/subs.test.ts
singingwolfboy: This issue seems relevant: https://github.com/facebook/jest/issues/8794

benjie: Noted. Merge as-is then :)

packages/postgraphile-core/__tests__/integration/schema/core.js
```diff
@@ -12,6 +11,6 @@ exports.test = (schemas, options, setup, finalCheck = () => {}) => () =>
       }
     }
     const schema = await createPostGraphileSchema(client, schemas, options);
-    expect(printSchemaOrdered(schema)).toMatchSnapshot();
+    expect(schema).toMatchSnapshot();
```
benjie:
```suggestion
    expect(lexicographicSortSchema(schema)).toMatchSnapshot();
```

packages/postgraphile-core/__tests__/integration/schema/__snapshots__/disabledTagsNoLegacyRelations.test.js.snap
```diff
@@ -1628,1025 +1628,1025 @@ enum LeftArmsOrderBy {
...
-    \\"\\"\\"
+    """
     input: JsonbIdentityMutationPlpgsqlWithDefaultInput!
   ): JsonbIdentityMutationPlpgsqlWithDefaultPayload
+  jsonIdentityMutation(
+    """
+    The exclusive input argument for this mutation. An object type, make sure to see documentation for this object’s fields.
+    """
+    input: JsonIdentityMutationInput!
+  ): JsonIdentityMutationPayload
```
benjie: This has been reordered I think?

EDIT: ah, it's because my old schema sorter and GraphQL's differ slightly. This is fine.

ab-pm:
Wow, this is awesome! But why don't you want to sort schemas by default?

Btw, in my own project I even omit field descriptions from the schema to get an even cleaner snapshot diff:
```ts
import { GraphQLSchema, buildASTSchema, lexicographicSortSchema, parse, printSchema, visit } from 'graphql';
const simplified: GraphQLSchema = lexicographicSortSchema(buildASTSchema(visit(parse(printSchema(schema)), {
	StringValue(node, key) {
		return key == 'description' ? null : node;
	}
})));
```
Might be interesting to provide as an option.

singingwolfboy:
@ab-pm: quoting @benjie's [resolved comment](https://github.com/graphile/graphile-engine/pull/533#discussion_r334270938):

> Ah; these are now being alphabetically sorted. This can be misleading as order is significant in GraphQL (e.g. it impacts the order that things are show in auto-complete/in the GraphQL docs) and can be used to group related things together (e.g. showing simple fields before showing relations). There are some places where the schema should be compared sorted, and some where order is significant. Typically I want to be aware if the schema order has changed as it will affect people dumping their schema downstream. For now; lets try and match the previous snapshots (but with the unnecessary backslashes removed).

Unfortunately, I don't think that Jest snapshot matchers can expose a configuration interface, so there's no way to provide an option to change these. However, if I'm wrong, then feel free to file an issue on the repository suggesting this change -- or even better, send a pull request! 😄

ab-pm:
Thanks, I had not seen that.

Yeah, it seems Jest snapshot serialisers are not configurable, unless you add them explicitly with [`expect.addSnapshotSerializer`](https://jestjs.io/docs/en/expect#expectaddsnapshotserializerserializer) where one could call a function with options that creates the serialiser.

benjie:
The field order is _officially_ significant (Lee Byron himself states so); so by default we should honour it. `lexicographicSortSchema` is always available if you want to compare a sorted schema :+1:

PR #534, opened 2019-10-08
feat(utils): @pgQuery support for scalars

phryneas:
We had already chatted about this - this PR allows for scalars to be added as subqueries using the `@pgQuery` directive with the `fragment` argument.

If this is merged, it would be great if you could mention my employer [Mayflower](https://mayflower.de/) in the release notes, as this PR was done on company time. But if this something you don't do, thats fine, too :smiley:

### DIFF
--- a/packages/graphile-utils/__tests__/ExtendSchemaPlugin-pg.test.js
+++ b/packages/graphile-utils/__tests__/ExtendSchemaPlugin-pg.test.js
@@ -722,8 +722,11 @@ it("allows to retrieve a single scalar value", async () => {
               myCustomScalar: Int! @pgQuery(
                 fragment: ${embed(sql.fragment`(SELECT 100)`)}
               )
-              myCustomScalarWithFunction: Int! @pgQuery(
-                fragment: ${embed(queryBuilder => sql.fragment`(SELECT 101)`)}
+              myCustomScalarWithFunction: String! @pgQuery(
+                fragment: ${embed(
+                  queryBuilder =>
+                    sql.fragment`(${queryBuilder.getTableAlias()}.name || ' ' || ${queryBuilder.getTableAlias()}.email)`
+                )}
               )
               myCustomScalarWithFunctionAndArgument(test: Int!): Int! @pgQuery(
                 fragment: ${embed(
@@ -737,8 +740,7 @@ it("allows to retrieve a single scalar value", async () => {
       }),
     ],
   });
-  const printedSchema = printSchema(schema);
-  expect(printedSchema).toMatchSnapshot();
+  expect(schema).toMatchSnapshot();
   const pgClient = await pgPool.connect();
   try {
     const { data, errors } = await graphql(
@@ -751,6 +753,9 @@ it("allows to retrieve a single scalar value", async () => {
             myCustomScalar
             myCustomScalarWithFunction
             myCustomScalarWithFunctionAndArgument(test: 102)
+            m100: myCustomScalar
+            mAlice: myCustomScalarWithFunction
+            m103: myCustomScalarWithFunctionAndArgument(test: 103)
           }
         }
       `,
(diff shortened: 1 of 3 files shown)


phryneas:
As for that failing test... I have no idea what that means. Doesn't seem to fail in every environment?

phryneas:
Okay, figured it out - my test-SQL needed a CAST for Postgres 9. Tests are now working fine :)

benjie:
Awesome; from a superficial glance this looks great but I'll try and review it properly next week. In the mean time, please change your `CAST(${...} AS type)` to `${...}::type` for consistency with the rest of the project :grin:

phryneas:
sure thing :+1: 
should be consistent now.

packages/graphile-utils/__tests__/ExtendSchemaPlugin-pg.test.js
```diff
@@ -718,3 +718,116 @@ it("allows adding a single table entry to a nested type", async () => {
...
+                )}
+              )
+            }
+          `,
+        };
+      }),
+    ],
+  });
+  const printedSchema = printSchema(schema);
+  expect(printedSchema).toMatchSnapshot();
```
singingwolfboy: Once https://github.com/graphile/graphile-engine/pull/533 lands, can you rebase this pull request on top of it and simplify this test? You'll no longer need to call `printSchema()` yourself, you can just do `expect(schema).toMatchSnapshot()`. 😄 

I'm not sure when that PR will land, but I wanted to give you a heads-up.

phryneas: will do, please ping me when it lands

packages/graphile-utils/__tests__/ExtendSchemaPlugin-pg.test.js
```diff
@@ -718,3 +718,116 @@ it("allows adding a single table entry to a nested type", async () => {
...
+    disableDefaultMutations: true,
+    appendPlugins: [
+      makeExtendSchemaPlugin(build => {
+        const { pgSql: sql } = build;
+        return {
+          typeDefs: gql`
+            extend type User {
+              myCustomArrayOfScalars: [String!]! @pgQuery(
+                fragment: ${embed(
+                  sql.fragment`array(SELECT name from ${sql.fragment`graphile_utils.pets`})`
```
benjie: This is equivalent to:

```suggestion
                  sql.fragment`array(SELECT name from graphile_utils.pets)`
```

I've never used the [array constructor](https://www.postgresql.org/docs/12/sql-expressions.html#SQL-SYNTAX-ARRAY-CONSTRUCTORS) in this way; I've always done `select array_agg(name) from ...` instead. It seems that this might actually be [faster because it's simpler](https://dba.stackexchange.com/questions/159710/why-is-array-agg-slower-than-the-non-aggregate-array-constructor) so I should definitely experiment with its usage in PostGraphile.

benjie: Filed: https://github.com/graphile/graphile-engine/issues/543

packages/graphile-utils/__tests__/ExtendSchemaPlugin-pg.test.js
```diff
@@ -718,3 +718,116 @@ it("allows adding a single table entry to a nested type", async () => {
...
+    const { data, errors } = await graphql(
+      schema,
+      `
+        query {
+          user: userById(id: 1) {
+            id
+            name
+            myCustomScalar
+            myCustomScalarWithFunction
+            myCustomScalarWithFunctionAndArgument(test: 102)
```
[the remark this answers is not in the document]
phryneas: I'm guessing you want one alias for each one of them, not the same one twice?

phryneas:
@benjie I think I adressed everything

PR #537, opened 2019-10-12
feat(QueryBuilder): new methods for managing QB children

singingwolfboy:
This adds three new public methods to the `QueryBuilder` class: `buildChild`, `buildNamedChildFrom`, and `getNamedChild`. These new methods will enable new capabilities for PostGraphile plugins. @benjie and I discussed this via Discord a few days ago, so I know he approves of these new methods.

I do have a few qustions:
* Should we also add a `buildNamedChild` method, which is like `buildNamedChildFrom` but does not call the `select`, `from`, or `lock` methods? It seems like an obvious extension of this API, but I'm not sure if it's actually useful/necessary. It might be better to wait until we find a use-case before building it.
* Should I add automated tests for this change? I don't see any automated tests for the `QueryBuilder` class at all.

### DIFF
--- a/packages/graphile-build-pg/src/QueryBuilder.d.ts
+++ b/packages/graphile-build-pg/src/QueryBuilder.d.ts
@@ -93,6 +93,13 @@ export default class QueryBuilder {
     addNotDistinctFromNullCase?: boolean;
     useAsterisk?: boolean;
   }): SQL;
+  public buildChild(): QueryBuilder;
+  public buildNamedChildFrom(
+    name: string,
+    from: SQLGen,
+    field: SQLGen
+  ): QueryBuilder;
+  public getNamedChild(name: string): QueryBuilder | undefined;
 
   // ----------------------------------------
 
--- a/packages/graphile-build-pg/src/QueryBuilder.d.ts
+++ b/packages/graphile-build-pg/src/QueryBuilder.d.ts
@@ -94,13 +94,13 @@ export default class QueryBuilder {
     useAsterisk?: boolean;
   }): SQL;
   public buildChild(): QueryBuilder;
-  public buildNamedChildFrom(
-    name: string,
+  public buildNamedChildSelecting(
+    name: RawAlias,
     from: SQLGen,
     field: SQLGen,
-    alias: SQLAlias
+    alias?: SQLAlias
   ): QueryBuilder;
-  public getNamedChild(name: string): QueryBuilder | undefined;
+  public getNamedChild(name: RawAlias): QueryBuilder | undefined;
 
   // ----------------------------------------
 
--- a/packages/graphile-build-pg/src/QueryBuilder.d.ts
+++ b/packages/graphile-build-pg/src/QueryBuilder.d.ts
@@ -97,7 +97,7 @@ export default class QueryBuilder {
   public buildNamedChildSelecting(
     name: RawAlias,
     from: SQLGen,
-    field: SQLGen,
+    selectExpression: SQLGen,
     alias?: SQLAlias
   ): QueryBuilder;
   public getNamedChild(name: RawAlias): QueryBuilder | undefined;
--- a/packages/graphile-build-pg/src/QueryBuilder.d.ts
+++ b/packages/graphile-build-pg/src/QueryBuilder.d.ts
@@ -97,7 +97,8 @@ export default class QueryBuilder {
   public buildNamedChildFrom(
     name: string,
     from: SQLGen,
-    field: SQLGen
+    field: SQLGen,
+    alias: SQLAlias
   ): QueryBuilder;
   public getNamedChild(name: string): QueryBuilder | undefined;
 
--- a/packages/graphile-build-pg/src/QueryBuilder.js
+++ b/packages/graphile-build-pg/src/QueryBuilder.js
@@ -73,7 +73,6 @@ class QueryBuilder {
   context: GraphQLContext;
   rootValue: any; // eslint-disable-line flowtype/no-weak-types
   supportsJSONB: boolean;
-  isSingleFieldSelector: boolean;
   locks: {
     [string]: false | true | string,
   };
@@ -127,7 +126,8 @@ class QueryBuilder {
   lockContext: {
     queryBuilder: QueryBuilder,
   };
-  _children: Map<string, QueryBuilder>;
+  _children: Map<RawAlias, QueryBuilder>;
+  _isSingleFieldSelector: boolean;
 
   constructor(
     options: QueryBuilderOptions = {},
@@ -689,19 +689,20 @@ ${sql.join(
       useAsterisk?: boolean,
     } = {}
   ) {
-    if (this.isSingleFieldSelector) {
+    let processedOptions: typeof options = options;
+    if (this._isSingleFieldSelector) {
       if (Object.keys(options).length > 0) {
         throw new Error(
-          "Can't use QueryBuilder.build(options) in isSingleFieldSelector"
+          "Do not pass options to QueryBuilder.build() when using `buildNamedChildSelecting`"
         );
       }
-      options = {
+      processedOptions = {
         asRaw: true,
       };
     } else {
       if (options.asRaw) {
         throw new Error(
-          "We don't support using asRaw except with buildNamedChildFrom"
+          "We don't support using asRaw except with buildNamedChildSelecting"
         );
       }
     }
@@ -713,7 +714,7 @@ ${sql.join(
       addNullCase = false,
       addNotDistinctFromNullCase = false,
       useAsterisk = false,
-    } = options;
+    } = processedOptions;
 
     this.lockEverything();
     if (onlyJsonField) {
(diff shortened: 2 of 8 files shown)


benjie:
> It might be better to wait until we find a use-case before building it.

Agree; that's why we've allowed space for it with the naming.

> Should I add automated tests for this change? I don't see any automated tests for the QueryBuilder class at all.

I'm not a huge fan of unit tests; particularly for units like QueryBuilder that are constantly evolving. I prefer integration tests. QueryBuilder is _very heavily_ tested because literally every GraphQL query that relates to PostgreSQL is using QueryBuilder.

Before we can merge this, we will need it to be used in some tests. A simple way to do this would be to do an integration test where you add a plugin in the tests that uses this functionality, and then run some queries through it to make sure they work. Have a look at these schemas for example:

https://github.com/graphile/graphile-engine/blob/7007a4308ce3cb12426c8559b29cf4672ec40a06/packages/postgraphile-core/__tests__/integration/queries.test.js#L61-L121

(I really need to get around to restructuring these tests.)

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -126,6 +126,7 @@ class QueryBuilder {
   lockContext: {
     queryBuilder: QueryBuilder,
   };
+  children: Map<string, QueryBuilder>;
```
benjie: Lets rename this to `_children` to make it clear it's not intended to be a public API.

```suggestion
  _children: Map<string, QueryBuilder>;
```

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -874,6 +876,26 @@ order by (row_number() over (partition by 1)) desc`;
...
+    child.parentQueryBuilder = this;
+    return child;
+  }
+  buildNamedChildFrom(name: string, from: SQLGen, field: SQLGen) {
+    if (this.children.has(name)) {
+      throw new Error(`QueryBuilder already has a child named ${name}`);
+    }
+    const child = this.buildChild();
+    child.from(from);
+    child.select(field, "_");
```
benjie: Let's use a more readable alias here. We use `"value"` elsewhere, but if we could come up with something sensible that was easier to search for that'd be great. Failing that, value is fine:

```suggestion
    child.select(field, "value");
```

packages/graphile-build-pg/src/QueryBuilder.d.ts
```diff
@@ -93,6 +93,14 @@ export default class QueryBuilder {
     addNotDistinctFromNullCase?: boolean;
     useAsterisk?: boolean;
   }): SQL;
+  public buildChild(): QueryBuilder;
+  public buildNamedChildFrom(
+    name: string,
+    from: SQLGen,
+    field: SQLGen,
+    alias: SQLAlias
```
benjie: I think this can be optional:

```suggestion
    alias?: SQLAlias
```

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -73,6 +73,7 @@ class QueryBuilder {
   context: GraphQLContext;
   rootValue: any; // eslint-disable-line flowtype/no-weak-types
   supportsJSONB: boolean;
+  isSingleFieldSelector: boolean;
```
benjie: Lets make it clear this is internal:
```suggestion
  _isSingleFieldSelector: boolean;
```

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -669,13 +682,31 @@ ${sql.join(
...
       onlyJsonField?: boolean,
       addNullCase?: boolean,
       addNotDistinctFromNullCase?: boolean,
       useAsterisk?: boolean,
     } = {}
   ) {
+    if (this.isSingleFieldSelector) {
+      if (Object.keys(options).length > 0) {
+        throw new Error(
+          "Can't use QueryBuilder.build(options) in isSingleFieldSelector"
```
benjie: This may not make sense to the person who receives it; maybe something like this?
```suggestion
          "Do not pass options to QueryBuilder.build() when using `buildNamedChildFrom`"
```

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -669,13 +682,31 @@ ${sql.join(
...
+        );
+      }
+      options = {
```
benjie: Please don't reassign function arguments; need to store this into a separate variable.

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -874,6 +907,32 @@ order by (row_number() over (partition by 1)) desc`;
...
     this.lock("select");
   }
+  buildChild() {
+    const options = { supportsJSONB: this.supportsJSONB };
+    const child = new QueryBuilder(options, this.context, this.rootValue);
+    child.parentQueryBuilder = this;
+    return child;
+  }
+  buildNamedChildFrom(
+    name: string,
```
benjie: Let's allow `name` to be `string | symbol` because symbols are guaranteed to be unique. This could end up being pretty handy... Maybe.

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -874,6 +907,32 @@ order by (row_number() over (partition by 1)) desc`;
...
+    from: SQLGen,
+    field: SQLGen,
+    alias: SQLAlias
```
benjie:
```suggestion
    alias?: SQLAlias
```

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -874,6 +907,32 @@ order by (row_number() over (partition by 1)) desc`;
     this.lock("selectCursor");
     this.lock("select");
   }
+  buildChild() {
+    const options = { supportsJSONB: this.supportsJSONB };
+    const child = new QueryBuilder(options, this.context, this.rootValue);
+    child.parentQueryBuilder = this;
+    return child;
+  }
+  buildNamedChildFrom(
```
benjie: Given this enters into a new special mode, I think the name needs to reflect this. I'm not sure what a good name would be though, but this one doesn't seem quite right. Maybe:

```
  buildNamedSingleFieldChild(
```

what do you think?

---

Thinking about it more, I'm not sure it should be limited to a single field; for example the following should be valid:

```sql
select * from a where (a1, a2) in (select b1, b2 from b)
```

So maybe it's more that it has a fixed select clause:

```
buildNamedChildSelecting(name, expression, from, alias?)
```

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -570,6 +573,16 @@ ${sql.join(
...
+  buildSelectRaw() {
+    this.lockEverything();
+    return sql.join(
+      this.compiledData.select.map(
+        ([sqlFragment, alias]) =>
+          sql.fragment`${sqlFragment} as ${sql.identifier(alias)}`
+      ),
+      ", "
+    );
+  }
```
benjie: If we change to `buildNamedChildSelecting` or similar, I think this can be replaced with something much simpler, such as:

```js
buildSelectRaw() {
  return this._fixedSelectionExpression;
}
```

singingwolfboy: What would `this._fixedSelectionExpression` be, and how would it get defined? I'm assuming that it's something different from `this._isSingleFieldSelector`, but I'm not following you on exactly what it _should_ be.

benjie:
@singingwolfboy I've refactored this a bit; let me know if you're happy with it and if so we can merge. :+1:

benjie:
(The main change is that I've made the "exclusive select body" into it's own method so other plugins can use it for other things, and so that it doesn't need an unnecessary alias.)

singingwolfboy:
Looks great to me!

ab-pm:
I believe I understand the reasoning behind these new methods, but it would be nice if they could be documented somehow with usage instructions for plugin authors - either here on the thread, as comments in the querybuilder source code, or even on the documentation page.

Btw, should the children that are used by the builtin plugins (postgraphile-core) become named children?

benjie:
It's rare that named children are required. This release isn't officially out yet, so documentation is not ready yet.

ab-pm:
OK, then maybe I didn't understand their purpose completely. But it's good to know that documentation will come with the release :-)

benjie:
I've spent all day writing docs for the new release; here's some basic documentation for this feature but I've not done much detail because I don't expect many people will need it.

https://www.graphile.org/postgraphile/make-extend-schema-plugin/#querybuilder-named-children

ab-pm:
I thought of using a named child query builder instead of a JOIN for interfaces, when selecting fields that are stored in a shared table. So I'd want to add the subquery on-demand in the data generators, and re-use it when it already exists. However, each field uses it's own `select()`, and I don't know these beforehand, so I cannot use that `fixedSelectExpression`. I'm not sure why that is a requirement for use of `buildNamedChildSelecting`? I'd much rather use the original `buildNamedChildFrom`. Or am I overlooking something and this is the completely wrong tool?

benjie:
I’m not sure this is the right approach for that. I don’t like the “add it if it doesn’t already exist” approach, really. Still, I’m open to you extending this area of the API if it’s the only way to make it work.

ab-pm:
Yeah, there might be a better approach that involves more lookahead planning, but I guess making it more easy to reuse querybuilder children is still nice.

benjie:
We've left space in the API for `buildNamedChild` deliberately; but I don't want to actually add that functionality until we have a concrete use case for it where it's a good fit.

PR #538, opened 2019-10-13
markdown-doctest 1.0.0

singingwolfboy:
This allows us to drop Babel 6 from `yarn.lock`!


benjie:
I couldn't find the CHANGELOG entry for 1.0.0; so I looked through the commits and saw that it was primarily you migrating everything to modern tooling :raised_hands:

PR #540, opened 2019-10-13
chore(deps): remove babel core bridge

singingwolfboy:
Looks like we're still marking `babel-core` as a devDependency (as opposed to `@babel/core`). Since we're on Babel 7 everywhere, I don't think we need `babel-core` anymore.


benjie:
Yeah; I believe we needed this for jest for a while but we updated that a while back!

PR #542, opened 2019-10-15
fix(jwt): allow JWT exp to be bigint

benjie:
Fixes https://github.com/graphile/postgraphile/issues/1166

Technically we can only support ints up to 2^53-1 but that's probably enough for this ;)


benjie:
:+1:  Ready to merge

PR #548, opened 2019-10-29
wal2json is now available on Homebrew

singingwolfboy:
https://github.com/Homebrew/homebrew-core/pull/45010

### DIFF
--- a/packages/lds/README.md
+++ b/packages/lds/README.md
@@ -35,7 +35,13 @@ You need `wal2json`; this is available on:
 - Amazon RDS
 - (please send a PR adding more compatible providers)
 
-If you don't already have it, you can install it. On your Unix-based OS (assuming `pg_config` is in your path, and points to the correct PostgreSQL installation) you can add it in a few seconds with:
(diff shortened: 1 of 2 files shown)


benjie:
(The latter is particularly problematic if you have postgis installed as it doesn’t allow the data migration to take place under certain circumstances, e.g. if homebrew deletes the old postgres binaries before you’ve had a chance to migrate the data.)

singingwolfboy:
> Also can you confirm that doing the homebrew install won’t force you to upgrade postgres to the latest (e.g. 12)

I'm pretty sure that Homebrew _does_ automatically upgrade dependencies when installing a new package, and postgres is a dependency of wal2json. Maybe it's simpler to just not mention the Homebrew option, given all of these caveats...

benjie:
I think that might be wise; alternatively we can just suggest afterwards "wal2json is also available via Homebrew on Mac". I think the current install instructions are probably least likely to cause the user issues 👍

benjie:
I'm going to close this for now :+1:

PR #551, opened 2019-11-04
feat: allow granular list nulls in nullability

kevboh:
In the makeChangeNullability plugin, for list types, adds the ability to specify if both the list type itself should be nullable _and_ if its member type should be nullable, allowing you to get the following GraphQL type signature:

```
someList: [MyType!]!
```

### DIFF
--- a/packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
+++ b/packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
@@ -3,7 +3,7 @@ import { GraphQLInputFieldConfig, GraphQLFieldConfig } from "graphql";
 
 interface ChangeNullabilityRules {
   [typeName: string]: {
-    [fieldName: string]: boolean;
+    [fieldName: string]: boolean | { namedType?: boolean; listType?: boolean };
   };
 }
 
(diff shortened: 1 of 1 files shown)


kevboh:
I would like to also update the corresponding docs, but I'm not sure where they're stored.

benjie:
I've thought about this a bit more and it's possible to have things like:

```graphql
type Bar {
  foo: [[[Int!]]!]
}
```

So `listType` and `namedType` aren't quite enough.

Maybe we could do `listType: [boolean]`

Then we could have:

```js
{ Bar: { foo: {
  listType: [false, true, false],
  namedType: true
}}}
```

That'd change it to:

```graphql
type Bar {
  foo: [[[Int]!]]!
}
```

I think this is going to have to walk the types manually using the `.kind` (or `.kindOf`? I forget) property on the types, calling `getNullableType` at each point. We should also allow `null`/`undefined` to just keep the previous nullability.

I totally understand if you'd rather I wrote this :grin:

benjie:
Regarding the docs, at the top of every page is a "Suggest improvements to this page" link that takes you to the edit flow on GitHub. https://www.graphile.org/postgraphile/make-change-nullability-plugin/

kevboh:
@benjie oh, interesting. That change would definitely require more learning around the GraphQL APIs on my part, but I could do it if you're okay with this stretching out a few weeks—I'm pretty slammed at the moment.

benjie:
Me too; stretch it out :+1:

benjie:
@kevboh I'm going to close this for now as no-one else has requested this feature; feel free to reopen it when you get a chance to address the feedback :+1:

kevboh:
@benjie yep, no worries. The hacky plugin we've been using for this continues to work well, but isn't really portable. Happy to put in the work to make it so if others need it.

PR #552, opened 2019-11-04
RFC: phase-dependent typings for Builder

phryneas:
So, I already have some collection of ad-hoc typings for Builder, and I thought of doing something useful with them.

This would expose different interfaces like `BuildPropertiesIntroducedIn_GraphQLObjectType` that can be extended from plugin's type definitions. 
These extended types will be available in (in this case) the `GraphQLObjectType` phase as Partial (because these typing can't track if they've already been set within the phase, and fully in all following phases.

I also added such type augmentations for `PgIntrospectionPlugin` and `pgBasicsPlugin`, as these are the one that have most impact on my code.

As a first result, this uncovered some errors in the typings of `makeExtendSchemaPlugin` as the graphql types there were referenced as instance types, not their respective constructors/classes.

This has a LOT of `any` in it right now, but that already is better than the current state of affairs, as it at least describes the existence of keys. Where I could easily add types, I did so.

But at the moment, I'd rather wait for your input, if something like this is a direction you'd like to be going instead of charging further ahead ;)

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.d.ts
+++ b/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.d.ts
@@ -195,3 +195,10 @@ export default function PgIntrospectionPlugin(
   builder: SchemaBuilder,
   options: Options
 ): Promise<void> | void;
+
+import "graphile-build";
+declare module "graphile-build" {
+  interface BuildPropertiesIntroducedIn_build {
+    pgIntrospectionResultsByKind: PgIntrospectionResultsByKind;
+  }
+}
--- a/packages/graphile-build/src/makeNewBuild.d.ts
+++ b/packages/graphile-build/src/makeNewBuild.d.ts
@@ -1,0 +1,69 @@
+import {
+  PgProc,
+  PgType,
+  PgClass,
+  PgAttribute,
+  PgConstraint,
+} from "graphile-build-pg";
+import { GraphQLNamedType } from "graphql";
+
+export interface Build {
+  options: any;
+  graphileBuildVersion: any;
+  versions: any;
+  hasVersion(
+    packageName: string,
+    range: string,
+    options?: { includePrerelease?: boolean }
+  ): boolean;
+  graphql: typeof import("graphql");
+  parseResolveInfo: any;
+  simplifyParsedResolveInfoFragmentWithType: any;
+  getSafeAliasFromAlias: any;
+  /** @deprecated */
+  getAliasFromResolveInfo: any;
+  getSafeAliasFromResolveInfo: any;
+  resolveAlias(data: any, _args: any, _context: any, resolveInfo: any): any;
+  addType(type: GraphQLNamedType, origin?: string): void;
+  getTypeByName(typeName: any): any;
+  extend: any;
+  newWithHooks(
+    Type: any,
+    spec: any,
+    inScope: any,
+    performNonEmptyFieldsCheck?: any
+  ): any;
+  /** @deprecated */
+  fieldDataGeneratorsByType: any;
+  fieldDataGeneratorsByFieldNameByType: any;
+  fieldArgDataGeneratorsByFieldNameByType: any;
+
+  inflection: Inflection;
+
+  swallowError: any;
+  /** @deprecated EXPERIMENTAL, API might change! */
+
+  resolveNode: any;
+  status: {
+    currentHookName: any;
+    currentHookEvent: any;
+  };
+  liveCoordinator: any;
+  scopeByType: Map<any, any>;
+  [str: string]: any;
+}
+
+export interface Inflection {
+  pluralize(...args: any[]): string;
+  singularize(...args: any[]): string;
+  upperCamelCase(...args: any[]): string;
+  camelCase(...args: any[]): string;
+  constantCase(...args: any[]): string;
+  // Built-in names (allows you to override these in the output schema)
+  builtin(...args: any[]): string;
+  // When converting a query field to a subscription (live query) field, this allows you to rename it
+  live(...args: any[]): string;
+  // Try and make something a valid GraphQL 'Name'
+  coerceToGraphQLName(name: string): string;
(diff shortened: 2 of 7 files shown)


packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.d.ts
```diff
@@ -195,3 +195,10 @@ export default function PgIntrospectionPlugin(
   builder: SchemaBuilder,
   options: Options
 ): Promise<void> | void;
+
+import "graphile-build";
+declare module "graphile-build" {
+  interface BuildPropertiesIntroducedIn_build {
+    pgIntrospectionResultsByKind: PgIntrospectionResultsByKind;
+  }
+}
```
phryneas: This is a good example on how most plugins would be using it to extend Build.

In the build phase, `pgIntrospectionResultsByKind` will be `undefined | PgIntrospectionResultsByKind` - in all later phases, it will be `PgIntrospectionResultsByKind`. 
In earlier phases (okay, there's nothing earlier thatn build, but you get the gist), the property wouldn't be present.

packages/graphile-utils/src/makeExtendSchemaPlugin.ts
```diff
@@ -81,8 +92,20 @@ export interface ExtensionDefinition {
...
 
+type GraphQLConstructorType =
+  | typeof GraphQLScalarType
+  | typeof GraphQLObjectType
+  | typeof GraphQLInterfaceType
+  | typeof GraphQLUnionType
+  | typeof GraphQLEnumType
+  | typeof GraphQLInputObjectType
+  | typeof GraphQLList
+  | typeof GraphQLNonNull
```
benjie: I don't think List or NonNull should be here?

packages/graphile-utils/src/makeExtendSchemaPlugin.ts
```diff
@@ -81,8 +92,20 @@ export interface ExtensionDefinition {
...
+  | typeof GraphQLDirective
+  | Resolvers;
```
benjie: This seems out of place?

packages/graphile-build/src/makeNewBuild.d.ts
```diff
@@ -0,0 +1,69 @@
...
+  ): boolean;
+  graphql: typeof import("graphql");
+  parseResolveInfo: any;
+  simplifyParsedResolveInfoFragmentWithType: any;
+  getSafeAliasFromAlias: any;
+  /** @deprecated */
+  getAliasFromResolveInfo: any;
+  getSafeAliasFromResolveInfo: any;
+  resolveAlias(data: any, _args: any, _context: any, resolveInfo: any): any;
+  addType(type: GraphQLNamedType, origin?: string): void;
```
benjie: This should be the class type: GraphQLConstructorType or whatever.

```suggestion
  addType(type: GraphQLConstructorType, origin?: string): void;
```

benjie:
Interesting concept!

Once the `build` hook is complete, the `Build` object gets frozen so it cannot be augmented further (only the object itself is frozen, children can still change) so you only need two objects really: base Build and the Build after the `build` hook has completed.

It'd be interesting knowing which things you can/cannot rely on in the `build` and `inflection` hooks, but alas I don't think we can actually implement it, since the build object passed to each `build` hook will differ based on the previous hooks that have executed and these run in a user-configurable (and augmentable) order. So I worry this is not returning sufficient value to justify the increased complexity/maintenance costs. I think it might be better solved by making the `Build` type into `BaseBuild & Partial<BuildExtensions>` for the `build` and `inflection` hooks which are called before `Build` is finalised, and `BaseBuild & BuildExtensions` for those thereafter. Plugins should then extend `BuildExtensions`.

My plan is to use "declaration merging" in the TypeScript conversion to model these objects a lot better; you can read more about that here: https://github.com/graphile/graphile-engine/issues/553#issuecomment-549786447

I'd advise that you don't invest too much time in our TypeScript typings right now unless you can apply it to our conversion process.

I'm going to close this for now (as mentioned in the above comment I'm trying to close all the PRs before moving to TypeScript), but please raise any small typing fixes in their own PRs (for example your change to `makeExtendSchemaPlugin`).

Thanks again for putting this time in; I hadn't considered the `BaseBuild & Partial<BuildExtensions>` approach before but I think it'll improve typings significantly (though it will probably cause some churn for users who have TypeScript plugins!).

phryneas:
> Once the `build` hook is complete, the `Build` object gets frozen so it cannot be augmented further (only the object itself is frozen, children can still change) so you only need two objects really: base Build and the Build after the `build` hook has completed.

Ah, that was the part of information I was missing. I had seen some plugins that changed stuff on build in later phases, but now that I think about it - they changed just children.

> It'd be interesting knowing which things you can/cannot rely on in the `build` and `inflection` hooks, but alas I don't think we can actually implement it, since the build object passed to each `build` hook will differ based on the previous hooks that have executed and these run in a user-configurable (and augmentable) order. So I worry this is not returning sufficient value to justify the increased complexity/maintenance costs. 

Yeah, I had thought about actually implementing something like that, but given that the order is determined dynamically just by "before" and "after" references on other plugins, there's no way that could be implemented in TypeScript. I guess that's for the better :D

I think it might be better solved by making the `Build` type into `BaseBuild & Partial<BuildExtensions>` for the `build` and `inflection` hooks which are called before `Build` is finalised, and `BaseBuild & BuildExtensions` for those thereafter. Plugins should then extend `BuildExtensions`.

Sounds like a great plan!

> I'm going to close this for now (as mentioned in the above comment I'm trying to close all the PRs before moving to TypeScript), but please raise any small typing fixes in their own PRs (for example your change to makeExtendSchemaPlugin).

Will do, probably during my next slacktime or so :+1: 


*I'm really looking forward to that TS transistion :)*

benjie:
> Will do, probably during my next slacktime or so +1

Awesome :raised_hands: 

> I'm really looking forward to that TS transistion :)

You and me both :sweat_smile:

PR #555, opened 2019-11-05
feat(utils): `@pgQuery.source` can be function

benjie:
E.g. if we need to enable a subquery that references the parent query builder's alias.

```graphql
foo: ItemConnection @pgQuery(
  source: ${embed(
    ({ parentQueryBuilder }) => sql.fragment`(
  select items.*
  from mindtree_public.items
  inner join lateral unnest(${parentQueryBuilder.getTableAlias()}."starredTopicIds") with ordinality as star(id, idx) on (star.id = items.id)
  where items.type = 'TOPIC'
  order by idx asc
)`
  )}
  withQueryBuilder: ${...}
)
```

NOTE: we should also allow blocking the default value for the `orderBy` argument which is added automatically (I think?) - e.g. in this case it's important the default order is `NATURAL` since the result set is already ordered (but the order value does not leak).


benjie:
:+1:  Ready for merge.

PR #558, opened 2019-11-06
fix(watch): don't built schema twice in watch mode

### LINKED ISSUE #550: graphile-build-pg in watch mode detects a "schema change" on startup
When I run PostGraphile in watch mode during development, it always builds the schema twice on startup. This is inconvenient when also running it from `nodemon` to watch code changes, which means that startup is frequent.

I believe this is caused by https://github.com/graphile/graphile-engine/blob/1689f6691a0716d41d256de3499c9da40407348f/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js#L923-L924
(and I can resolve it by commenting out that line) but I am not certain what "on reconnect" means here (in which cases that would happen). Maybe moving this into the `_reconnect` method would be the solution?

benjie:
Fixes #550

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
@@ -861,7 +861,7 @@ export default (async function PgIntrospectionPlugin(
       this._start();
     }
 
-    async _start() {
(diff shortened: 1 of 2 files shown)


ab-pm:
:+1:

PR #560, opened 2019-11-07
sponsorship: update READMEs

jemgillam:
Thank you to all our Featured Sponsors, past and present, for supporting our work 🤩


benjie:
Thanks for the PR, Jem :grin: This only updated the root README, we need the READMEs in all the packages to be updated. You can do so using the `scripts/sponsors` script - you need to update [the snippet in that script](https://github.com/graphile/graphile-engine/blob/358687d79d613012995974aa8ce30c5859801859/scripts/sponsors#L6-L19) and then run it - it should automatically update all the READMEs to match :rocket:

Issue #460, opened 2019-05-16
Add @omit single smart comment


tinymarsracing:
This would be useful for cleaning up your schema. I think an omit for queries by unique indexes (that are generated in addition to byId) is also not available.

---

BENJIE'S NOTES:

I've renamed the issue; we're currently proposing an `@omit single` smart comment which would:

- on a table `foos`: omit the single root fields `foo(nodeId: ID)` and `fooByUniqueKey(uniqueKey: ...)`
- on a unique constraint: omit the root field `fooByUniqueKey(uniqueKey: ...)`
- on a foreign key constraint: omit the "forward" relationship (it would not omit the backward relationship, even if it's a 1-to-1 relationship, to do so you'd use `@omit many` on the constraint)

benjie:
The subtleties here can be handled by placing the `@omit` on the constraints. An omit on the table directly would apply to all constraints. Effectively we want to prevent these constraints/tables being added to the _root_ of the GraphQL schema (i.e. to the Query object).

At the moment we use `CRUDFOAMX` as the [omit options](https://www.graphile.org/postgraphile/smart-comments/#omitting). Since we can't use `R` for "root" again, I think `Q` for `Query` might make sense. Any other suggestions?

tinymarsracing:
I realize now that a part of this was [already documented](https://www.graphile.org/postgraphile/smart-comments/#unique-key-record-finders-1). But I'm confused now. So, I guess we have a couple of queries we could potentially omit from root:
- allThings --> "@omit all" on table
- thingByPk --> "@omit" on constraint pk
- thingByAnyOtherUniqueConstraint --> "@omit" on that constraint
- thing (= thingByNodeId) --> ???
If we want to save some time:
- all root queries --> "@omit" on table

Okay, and now I think I get it. Currently, "@omit" also omits mutations etc. But we might want to keep them. But I guess in the CRUDFOAMX, I would expect that "@omit read" would do exactly that (omit all root queries). When I tried that however, it threw an error. And then "@omit" would be for omitting the whole CRUDFOAMX. Also, in the table, I would expect it to say: "R | read | omit root queries" and not "omit completely" which is not "@omit read", but "@omit"? I hope you can follow me. :)

benjie:
The thing is, `@omit read` to me means remove any way of reading that value - not just at the root, but also anywhere else. Omitting from the root is different.

For example, imagine a many-to-many table situation:

```sql
create table a (
  id serial primary key
);
create table b (
[...]

PR #565, opened 2019-11-14
feat(omit) take "many" on constraints into account

### LINKED ISSUE #505: backward relations should take "@omit many" on foreign table and constraints into account
I want to omit some "many" fields on a table (or really, on the whole table).
Unfortunately, the `@omit many` smart comment only allows me to annotate the table with the foreign key constraint, dropping all its backward relations - instead of having me annotate the referenced table, on which the fields to be dropped are generated. I would enjoy having an `omit(foreignTable, "many")` check added to the line https://github.com/graphile/graphile-engine/blob/b6fbfb849ee50e29b6c53114cb159dc4f12ddd19/packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.js#L242 but I guess that would be a backwards-incompatible change. So if that is not possible, I'd suggest allowing `many` also on the individual constraints or fields, i.e. at https://github.com/graphile/graphile-engine/blob/b6fbfb849ee50e29b6c53114cb159dc4f12ddd19/packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.js#L69 https://github.com/graphile/graphile-engine/blob/b6fbfb849ee50e29b6c53114cb159dc4f12ddd19/packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.js#L103

ab-pm:
Closes #505.

### DIFF
--- a/packages/postgraphile-core/__tests__/integration/schema/omit-rename.omitstuff.test.js
+++ b/packages/postgraphile-core/__tests__/integration/schema/omit-rename.omitstuff.test.js
@@ -56,13 +56,25 @@ comment on table d.films is E'@omit read,all,update,create,delete,many';
 );
 
 test(
-  "omit and omit many on table",
+  "omit and omit everything on table",
   core.test(
     ["d"],
     {},
     `
 comment on table d.films is E'@omit *';
 comment on table d.tv_shows is E'@omit';
+`
+  )
+);
+
+test(
+  "omit many on constraint and table",
+  core.test(
+    ["d"],
+    {},
+    `
+comment on constraint post_author_id_fkey on d.post is E'@omit many\n@fieldName author';
+comment on constraint tv_shows_studio_id_fkey on d.tv_shows is E'@omit many';
 comment on table d.tv_episodes is E'@omit many';
 `
   )
--- a/packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.ts
+++ b/packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.ts
@@ -88,11 +88,16 @@ export default (function PgBackwardRelationPlugin(
       return extend(
         fields,
         foreignKeyConstraints.reduce((memo, constraint) => {
-          if (omit(constraint, "read") || omit(constraint, "many")) {
+          if (omit(constraint, "read")) {
             return memo;
           }
           const table =
             introspectionResultsByKind.classById[constraint.classId];
+          if (!table) {
+            throw new Error(
+              `Could not find the table that referenced us (constraint: ${constraint.name})`
+            );
+          }
           if (!table.isSelectable) {
             // Could be a composite type
             return memo;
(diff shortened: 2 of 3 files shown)


packages/postgraphile-core/__tests__/integration/schema/omit-rename.omitstuff.test.js
```diff
@@ -56,13 +56,25 @@ comment on table d.films is E'@omit read,all,update,create,delete,many';
...
+
+test(
+  "omit many on constraint and table",
+  core.test(
+    ["d"],
+    {},
+    `
+comment on constraint post_author_id_fkey on d.post is E'@omit many\n@fieldName author';
+comment on constraint tv_shows_studio_id_fkey on d.tv_shows is E'@omit many';
 comment on table d.tv_episodes is E'@omit many';
```
ab-pm: I've moved this line into the new test, in the previous "omit and omit many on table" test it had no effect - the only table referencing `tv_episodes` was `tv_shows`, which was omitted entirely.

packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.js
```diff
@@ -64,7 +64,7 @@ export default (function PgBackwardRelationPlugin(
       return extend(
         fields,
         foreignKeyConstraints.reduce((memo, constraint) => {
-          if (omit(constraint, "read")) {
+          if (omit(constraint, "read") || omit(constraint, "many")) {
```
benjie: This will make the `many` also apply to singular backward relations which I don't think is the correct behaviour. Can we move this check down to `shouldAddManyRelation`? If we do so then I think we can class this as a bugfix rather than a breaking change - since `@omit many` on the constraint didn't do anything before, but omitting the backward many relation would be the expected outcome.

ab-pm: Hm, but I think we should have a way to omit singular backward relations as well (without omitting the forward relation), and `@omit many` is the best I can think of for that.
In #460 I proposed `@omit one` for the forward relation, but I don't know anything better than to use `@omit many` also on a one-to-one relation.

benjie: No, if we kept this it'd be a breaking change because the result would be unexpected. Regarding `@omit one` I'd have to evaluate it. Lets put this on ice for now and come up with a full solution to #460 before we try and change what the smart comment omits mean too much.

ab-pm: OK, I had changed it already to apply only on many-relationships, see the second commit. Until we come up with a solution, maybe emit a warning (error?) when there's an `@omit many` comment on a one-to-one constraint?

ab-pm:
Now with Typescript!

benjie:
Hey; sorry it's taken me a while to get around to this! Please could you merge the latest master :+1:

PR #567, opened 2019-11-19
feat(pg) single backward relations with arguments if primary key covered partially

ab-pm:
Integrating [the plugin](https://gist.github.com/ab-pm/0f0b021db2c541008875b581f91404ee) I had posted on Discord into core.

Basic tests (creates the fields it should, queries work as expected, can be omitted using new `@omit single` #460) included based on the existing test schema `c`.
Not sure if separate (smaller) schema is needed to go over all edge cases as well, including renaming and custom inflection?


ab-pm:
Does this need to go behind a feature flag? I figured this might break existing schemas that a) have such foreign references on parts of their primary keys and b) have a `@foreignFieldName` that doesn't properly singularise, leading to a name collision between the "many" backwards relation and the new single-but-parameterised relation.

benjie:
@ab-pm Thanks for the PR. I'm going to leave this hanging for a while because I want to figure out what I'm doing with the existing core plugins in v5 before extending the codebase further.

packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.ts
```diff
@@ -131,27 +132,49 @@ export default (function PgBackwardRelationPlugin(
...
 
+          const primaryKeyConstraint = table.primaryKeyConstraint;
+          const primaryKeys =
+            primaryKeyConstraint && primaryKeyConstraint.keyAttributes;
+          const uncoveredPrimaryKeys =
+            primaryKeys && keys.every(attr => primaryKeys.includes(attr))
+              ? primaryKeys.filter(attr => !keys.includes(attr))
+              : [];
+          const parameterKeys = uncoveredPrimaryKeys.map(key => ({
+            ...key,
```
benjie: There's less risk of conflicts here if we don't splat this out:

```suggestion
            key,
```

packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.ts
```diff
@@ -131,27 +132,49 @@ export default (function PgBackwardRelationPlugin(
...
           const singleRelationFieldName = isUnique
             ? inflection.singleRelationByKeysBackwards(
                 keys,
                 table,
                 foreignTable,
                 constraint
               )
+            : uncoveredPrimaryKeys.length
+            ? inflection.rowByRelationBackwardsAndUniqueKeys(
+                uncoveredPrimaryKeys,
```
benjie:
```suggestion
                uncoveredPrimaryKeys.map(details => details.key),
```

packages/graphile-build-pg/src/plugins/PgBackwardRelationPlugin.ts
```diff
@@ -131,27 +132,49 @@ export default (function PgBackwardRelationPlugin(
...
-
           const shouldAddSingleRelation =
-            isUnique && legacyRelationMode !== ONLY;
+            (isUnique && legacyRelationMode !== ONLY) ||
+            (!isUnique &&
+              !!uncoveredPrimaryKeys.length &&
+              primaryKeyConstraint &&
+              !omit(table, "single") &&
+              !omit(primaryKeyConstraint, "single") &&
+              !omit(constraint, "single"));
```
benjie: This is a lot of potential `omit`s. I'd like to really tighten up how `omit` works in V5, I think in this case it should be something like `!omit(constraint, "backward")`

PR #571, opened 2019-11-23
feat(introspection): Add manual schema reload notification

michaelbeaumont:
We're in a situation where we don't have super user access to the database (Cloud SQL) but want postgraphile to reload the schema when it changes, so we manually notify postgraphile when we update it using the following function:

```
  perform pg_notify(
    'postgraphile_watch',
    json_build_object(
      'type',
      'ddl',
      'payload',
      (select json_agg(json_build_object('schema', NULL, 'command', 'ignored by postgraphile')))
    )::text
  );
```
it works fine but feels a bit wrong so I thought I'd make this simple PR to add explicit support for it and allow the simpler: 

```
  perform pg_notify(
    'postgraphile_watch',
    json_build_object(
      'type',
      'manual',
    )::text
  );
```

Does this make sense?

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.ts
+++ b/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.ts
@@ -1145,7 +1145,7 @@ export default (async function PgIntrospectionPlugin(
         } else if (payload.type === "manual") {
           this._handleChange();
         } else {
-          throw new Error(`Payload type '${payload.type}' not recognised`);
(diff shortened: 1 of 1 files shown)


benjie:
LGMT! Proper review in the week, but it looks good to merge.

benjie:
(Sorry; the TypeScript conversion last week took a lot longer than I anticipated - I'll be reviewing this next week.)

benjie:
Hey @michaelbeaumont please could you re-apply these changes onto the latest master? Sorry :)

PR #573, opened 2019-11-29
feat(types): convert graphile-build and -pg to TypeScript

benjie:
This is 4 full days of work :sweat_smile: 

The initial conversion was automated; but then I went through and massively refactored all our types and then made all the code work with them. This should make discovering where types are defined and what they have available on them much  much easier to figure out.

I hope to follow this up at some point with some doc autogen. TBH I've not even added comments to most of these types yet; for now I was just focussed on getting the types actually added!

Anyone who uses TypeScript and has written custom plugins will likely have to refactor a little as all of the type definitions have changed; but regular JS users should not notice much difference.

We need to refactor some types in PostGraphile once this is merged before we can release both.


packages/graphile-build-pg/src/omit.ts
```diff
@@ -75,8 +77,19 @@ function parse(arrOrNot, errorPrefix = "Error") {
...
+    | "read"
+    | "update"
+    | "delete"
+    | "filter"
+    | "order"
+    | "all"
+    | "many"
+    | "execute"
+    | "base"
+    | string
```
benjie: (I know the individual permissions above are redundant because of the `string`. I want to keep them anyway.)

packages/graphile-build-pg/src/QueryBuilder.ts
```diff
@@ -1,32 +1,39 @@
...
+import chunk = require("lodash/chunk");
+import { PgClass, PgType } from "./plugins/PgIntrospectionPlugin";
+import { GraphileResolverContext } from "graphile-build";
+
+export { GraphileResolverContext };
+
+type SQL = import("pg-sql2").SQL;
+export { sql, SQL };
 
 // eslint-disable-next-line flowtype/no-weak-types
```
singingwolfboy: You can remove this `eslint-disable` comment, and all similar comments related to Flow, right?

packages/graphile-build-pg/src/plugins/PgColumnsPlugin.ts
```diff
@@ -1,8 +1,31 @@
...
+  parentQueryBuilder: QueryBuilder
+) => SQL;
+
+declare module "graphile-build" {
+  interface Build {
+    pgGetSelectValueForFieldAndTypeAndModifier: PgGetSelectValueForFieldAndTypeAndModifier;
+  }
+}
+
+const nullableIf = <T extends import("graphql").GraphQLNullableType>(
```
singingwolfboy: Why not import these types from "graphql" at the top of this file?

packages/graphile-build-pg/src/QueryBuilder.ts
benjie: Yeah; in fact that was why CI was failing :+1:

packages/graphile-build-pg/src/plugins/PgColumnsPlugin.ts
benjie: It's important that user plugins do not import `graphql` _values_ directly, and instead use the `graphql` available on the `Build` object to help prevent `graphql` version conflicts. To help with this, we're careful to try and avoid `import ... from "graphql"` in the core plugins to lead by example. In `graphile-build` itself this isn't super important, but in `graphile-build-pg` it is.

To help ensure that we only use the _types_ (as opposed to _values_) we use `import` in type positions.

ab-pm: Is it possible to make a linting rule for this? I expect `import`s to happen accidentally where the IDE is trying to be helpful and offering automated imports…

benjie: Yeah; that's exactly what happens. Had to rewrite most of these manually. If you're aware of a lint rule we can enable for this I'm all for it :+1:

benjie: (In flow you could do `import type` which would only import types and not values. Alas TypeScript does not draw this distinction.)

ab-pm: I guess [this one](https://eslint.org/docs/rules/no-restricted-imports) could do it? (Scoped to graphile-build-pg)
```
"no-restricted-imports": ["error", [{
  "name": "graphql",
  "message": "Please refer to build.graphql instead."
}]]
```
Afaik TypeScript only has the feature of not compiling an import if the imported values were only used in type declarations.

benjie: PR to add that lint rule would be welcome :)

PR #574, opened 2019-11-30
feat(utils): makeExtendSchemaPlugin accepts typeDef array

jkantr:
This falls in line a little closer with behavior seen in Apollo server and (I believe) the reference implementation. It also makes it slightly easier to modularize schema extensions local to a project using postgraphile.

First time looking at core of the project, so I hope I got the general idea right. I'm also not *super* familiar with the graphql AST so if this is too naive of an implementation I can go back to the drawing board :) fwiw it appears to work, and it should maintain order of the typeDefs as in the given array (which I imagine would be the important part).


packages/graphile-utils/src/makeExtendSchemaPlugin.ts
```diff
@@ -118,11 +119,26 @@ export default function makeExtendSchemaPlugin(
...
+            "The first argument to makeExtendSchemaPlugin must be generated by the `gql` helper, or an array of the same."
+          );
+        }
       }
+
+      const typeDefsArr = Array.isArray(typeDefs) ? typeDefs : [typeDefs];
+      const mergedTypeDefinitions = typeDefsArr.reduce(
+        (definitions, typeDef) => definitions.concat(typeDef.definitions),
+        [] as DefinitionNode[]
+      );
```
benjie: Please can you move the assertions above into the reducer?

```ts
      const typeDefsArr = Array.isArray(typeDefs) ? typeDefs : [typeDefs];
      const mergedTypeDefinitions = typeDefsArr.reduce(
        (definitions, typeDef) => {
          if (!(typeDef as any) || (typeDef as any).kind !== "Document") {
            throw new Error(
              "The first argument to makeExtendSchemaPlugin must be generated by the `gql` helper, or an array of the same."
            );
          }
          definitions.push(...typeDef.definitions);
          return definitions;
        },
        [] as DefinitionNode[]
      );
```

That way we only have to loop once, and we only throw the error in one place. I've also changed `concat` to `push(...` for [performance reasons](https://jsperf.com/javascript-array-concat-vs-push/294).

benjie:
This is perfect @jkantr; please can you merge in the latest master (don't rebase), resolve the conflicts and then we're good to merge! :tada:

PR #578, opened 2019-12-04
chore: TypeScript project references; fix some types

benjie:
Implements project references and firms up a few more types. https://www.typescriptlang.org/docs/handbook/project-references.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ node_modules
 /.vscode
 *.tgz
 yarn-error.log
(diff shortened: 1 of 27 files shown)


benjie:
We could switch to running a root `tsc` for build now, I think, but we'll leave it as-is in this PR and add it in a separate PR I think.

benjie:
Before merge: ensure this works with a clean checkout with no caches/etc.

PR #586, opened 2019-12-11
fix(smart-tags): enable using Smart Tags with @foreignKey and @primaryKey

benjie:
Fixes https://github.com/graphile/postgraphile/issues/1207


.flowconfig
```diff
@@ -3,6 +3,7 @@
 <PROJECT_ROOT>/packages/.*/index\.js\.flow
 <PROJECT_ROOT>/packages/.*/node8plus/.*\.flow
 <PROJECT_ROOT>/postgraphile
+<PROJECT_ROOT>/packages
```
benjie: We're moving to TypeScript so rather than fighting Flow I've just disabled it.

PR #587, opened 2019-12-13
fix(utils): fix smart tags regression on unqualified tables

keithlayne:
Verified that the new test failed without the change.  Not a spectacular test, but does what it advertises.


packages/graphile-utils/__tests__/makePgSmartTagsPlugin.test.js
```diff
@@ -50,51 +91,59 @@ it("allows adding a custom single field to PG schema", async () => {
...
-            Object {
-              "email": "alice@example.com",
-              "testSmartTagByEmail": null,
-            },
-            Object {
-              "email": "bob@example.com",
-              "testSmartTagByEmail": Object {
-                "email": "bob@example.com",
-                "val": 42,
+    expect(data).toMatchInlineSnapshot(inlineSnapshot);
```
benjie: `toMatchInlineSnapshot` is an updateable snapshot (run `jest -u` to update it) so it should not be a variable reference.

keithlayne: Sorry, should have dug in to that to do it correctly, but unfortunately I don't know  jest.  Thanks for fixing.

benjie: I figured; no need to apologise, the comment was just for guidance.

PR #588, opened 2020-01-03
Ranges can be unbounded

singingwolfboy:
https://www.postgresql.org/docs/12/rangetypes.html#RANGETYPES-INFINITE

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgTypesPlugin.ts
+++ b/packages/graphile-build-pg/src/plugins/PgTypesPlugin.ts
@@ -817,9 +817,16 @@ export default (function PgTypesPlugin(
             subtype.id,
             typeModifier
           );
+          const gqlRangeInputSubType = getGqlInputTypeByTypeIdAndModifier(
+            subtype.id,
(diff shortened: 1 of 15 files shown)


packages/graphile-build-pg/src/plugins/PgTypesPlugin.ts
```diff
@@ -874,7 +881,7 @@ export default (function PgTypesPlugin(
               fields: {
                 value: {
                   description: "The value at one end of our range.",
-                  type: new GraphQLNonNull(gqlRangeSubType),
+                  type: gqlRangeInputSubType,
```
singingwolfboy: I don't understand how this was working before, since there was a mismatch between input types and output types. I think that the Typescript definitions are too loose around this.

benjie:
We represent unbounded ranges by having the entire entry be null. e.g. `{start: {value: 3, inclusive: true}}` represents the range `[3,)` because it has no "end".

benjie:
Replaced by #591

Issue #595, opened 2020-01-23
Filtering subscription events


enisdenjo:
Instead of spamming the client with many messages and filtering them down on the client or maybe clutter the DB with custom GraphQL specific logic; move the filtering logic to the graphile engine.

Ability to filter published events would be very useful allowing the client to make more focused subscriptions just by increasing specificity through arguments.

Extending the usage example from the [docs](https://www.graphile.org/postgraphile/subscriptions/#writing-the-plugin), proposed filtered extension would look something like this:

Extra trigger listening only to `name` changes:
```sql
CREATE TRIGGER gql_user_name_update
  AFTER UPDATE ON app_public.users
  FOR EACH ROW
  WHEN (OLD.name <> NEW.name) -- triggered only when the name changes
  EXECUTE PROCEDURE app_public.graphql_subscription(
    'userNameChanged', -- event
    'graphql:user:$1', -- topic
    'id' -- subject
  );
```

Extended filtered plugin:
```js
function filter(event, _args, _resolveContext, _resolveInfo) {
  return payload.event === "userNameChanged";
}

module.exports = makeExtendSchemaPlugin(({ pgSql: sql }) => ({
  typeDefs: gql`
    type UserSubscriptionPayload {
      # This is populated by our resolver below
      user: User

      # This is returned directly from the PostgreSQL subscription payload (JSON object)
      event: String
    }

    extend type Subscription {
      # Triggered whenever the user changes, including name changes
      currentUserUpdated: UserSubscriptionPayload
        @pgSubscription(
          topic: ${embed(currentUserTopicFromContext)}
        )

      # Triggered ONLY when the name changes.
      currentUserNameUpdated: UserSubscriptionPayload
        @pgSubscription(
          topic: ${embed(currentUserTopicFromContext)},
          filter: ${embed(filter)}
        )
    }
  `,

  resolvers: {
    UserSubscriptionPayload: {
      // This method finds the user from the database based on the event
      // published by PostgreSQL.
      //
      // In a future release, we hope to enable you to replace this entire
      // method with a small schema directive above, should you so desire. It's
      // mostly boilerplate.
      async user(
        event,
        _args,
        _context,
        { graphile: { selectGraphQLResultFromTable } }
      ) {
        const rows = await selectGraphQLResultFromTable(
          sql.fragment`app_public.users`,
          (tableAlias, sqlBuilder) => {
            sqlBuilder.where(
[...]

PR #596, opened 2020-01-23
feat(subscriptions): @pgSubscription now supports filter

enisdenjo:
Closes: #595


benjie:
@enisdenjo This looks great; if you want it any time soon though you need to rebase it on the v4 branch. `master` now represents work in progress against v5, which won't be available for quite a while.

enisdenjo:
@benjie I won't need it soon, but others might (including me, but I don't know it yet haha); so here is the same PR rebased on `v4`.

PR #598, opened 2020-01-28
fix(introspection): don't query roles when ignoring RBAC

jnbarlow:
This fix is to make the introspection query not iterate over every role in the database on startup when `--no-ignore-rbac` is not present. (We are the crazy people with 4000+ roles and the 16 minute start times)

We patched our local copy of Postgraphile by just removing the `union all` below, but we wanted to make sure this didn't get broken by future versions and wanted to contribute back.

I'm new to typescript and I think I followed everything correctly.  Everything built and appeared to start up correctly with and without the `--no-ignore-rbac` parameter.

### DIFF
--- a/packages/graphile-build-pg/src/plugins/introspectionQuery.js
+++ b/packages/graphile-build-pg/src/plugins/introspectionQuery.js
@@ -9,9 +9,19 @@
  */
 function makeIntrospectionQuery(
   serverVersionNum: number,
-  options: { pgLegacyFunctionsOnly?: boolean } = {}
+  options: {
+    pgLegacyFunctionsOnly?: boolean,
+    pgIgnoreRBAC?: boolean,
+  } = {}
 ): string {
-  const { pgLegacyFunctionsOnly } = options;
+  const { pgLegacyFunctionsOnly = false, pgIgnoreRBAC = true } = options;
+  const unionRBAC = `
+    union all
+      select pg_roles.oid _oid, pg_roles.*
+      from pg_roles, accessible_roles, pg_auth_members
+      where pg_auth_members.roleid = pg_roles.oid
+        and pg_auth_members.member = accessible_roles._oid 
+  `;
   return `\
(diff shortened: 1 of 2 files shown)


packages/graphile-build-pg/src/plugins/introspectionQuery.ts
```diff
@@ -23,11 +33,7 @@ with
     select oid _oid, pg_roles.*
```
benjie: You should also drop the `recursive` from line 32 if `pgIgnoreRBAC`.

jnbarlow:
@benjie Addressed CR comments.  Please review at your convenience.

jnbarlow:
Also, I'm not sure how to rerun the ci..I think it's angry from when I switched the targets.

benjie:
Nothing a little turning it off and on again can't fix... :wink:

packages/graphile-build-pg/src/plugins/introspectionQuery.js
```diff
@@ -9,9 +9,19 @@
...
 function makeIntrospectionQuery(
   serverVersionNum: number,
-  options: { pgLegacyFunctionsOnly?: boolean } = {}
+  options: {
+    pgLegacyFunctionsOnly?: boolean,
+    pgIgnoreRBAC?: boolean,
+  } = {}
 ): string {
-  const { pgLegacyFunctionsOnly } = options;
+  const { pgLegacyFunctionsOnly, pgIgnoreRBAC } = options;
```
benjie:
```suggestion
  const { pgLegacyFunctionsOnly = false, pgIgnoreRBAC = true } = options;
```

jnbarlow: NP.. I'm playing in your sandbox now, gotta follow your rules ;)

jnbarlow:
@benjie added the defaults, squished it all back down to one commit.

PR #612, opened 2020-03-26
feat(subscriptions): @pgSubscription now supports initial events

enisdenjo:
## Description

Extends the `@pgSubscription` directive with an `initialEvent` argument allowing you to guarantee data freshness by emitting an event which triggers the related resolver as soon as the user subscribes.

## Example
```js
// FileSubscriptionPlugin.js

import { makeExtendSchemaPlugin, gql, embed } from 'graphile-utils';

function buildTopic(args) {
  return `fileChanged:${args.id}`;
}

function buildInitialEvent(args) {
  // the `topic` value will be injected in the returning event
  return { subject: args.id };
}

export const FileSubscriptionPlugin = makeExtendSchemaPlugin(({ pgSql: sql }) => ({
  typeDefs: gql`
    type FileChangedPayload {
      event: String!
      file: File!
    }

    extend type Subscription {
      fileChanged(id: UUID!): FileChangedPayload!
        @pgSubscription(
          topic: ${embed(buildTopic)}
          initialEvent: ${embed(buildInitialEvent)}
        )
    }
  `,
  resolvers: {
    FileChangedPayload: {
      // will be triggered as soon as the user subscribes because of the initialEvent argument
      async file(event, _args, _context, { graphile: { selectGraphQLResultFromTable } }) {
        /**
         * given that the fileChanged id argument is: '456f', the initial event will look like this:
         * ```
         * { topic: 'fileChanged:456f', subject: '456f' }
         * ```
         */

        const rows = await selectGraphQLResultFromTable(
          sql.fragment`public.file`,
          (tableAlias, sqlBuilder) => {
            sqlBuilder.where(sql.fragment`${tableAlias}.id = ${sql.value(event.subject)}`);
          },
        );
        return rows[0];
      },
    },
  },
}));
```

### DIFF
--- a/packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
+++ b/packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
@@ -102,7 +102,8 @@ const PgSubscriptionResolverPlugin: Plugin = function(builder, { pubsub }) {
                 "filter provided to pgSubscription must be a function"
               );
             }
-            asyncIterator = withFilter(() => asyncIterator, filter)(
+            const oldIterator = asyncIterator;
+            asyncIterator = withFilter(() => oldIterator, filter)(
               parent,
               args,
               resolveContext,
@@ -116,8 +117,8 @@ const PgSubscriptionResolverPlugin: Plugin = function(builder, { pubsub }) {
                 "initialEvent provided to pgSubscription must be a function"
               );
             }
-
-            return (async function* subscribeWithInitialEvent() {
+            const oldIterator = asyncIterator;
+            asyncIterator = (async function* subscribeWithInitialEvent() {
               const event = await initialEvent(
                 args,
                 resolveContext,
@@ -132,7 +133,7 @@ const PgSubscriptionResolverPlugin: Plugin = function(builder, { pubsub }) {
                 yield { ...event, topic };
               }
               for await (const val of {
-                [Symbol.asyncIterator]: () => asyncIterator,
+                [Symbol.asyncIterator]: () => oldIterator,
               }) {
                 yield val;
               }
--- a/packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
+++ b/packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
@@ -4,8 +4,40 @@ import { PubSub, withFilter } from "graphql-subscriptions";
 
 const debug = debugFactory("pg-pubsub");
 
-function isPubSub(pubsub: any): pubsub is PubSub {
+function isPubSub(pubsub: unknown): pubsub is PubSub {
   return !!pubsub;
+}
+
+function withInitialEvent(
+  topic: string,
+  initialEvent: (
+    args: { [key: string]: unknown },
+    resolveContext: unknown,
+    resolveInfo: unknown
+  ) => unknown
+) {
+  return async function* subscribeWithInitialEvent(
+    asyncIterator: AsyncIterator<unknown>,
+    _parent: unknown,
+    args: { [key: string]: unknown },
+    resolveContext: unknown,
+    resolveInfo: unknown
+  ) {
+    const event = await initialEvent(args, resolveContext, resolveInfo);
+    if (event != null) {
+      if (typeof event !== "object") {
+        throw new Error(
+          "event returned from pgSubscription initialEvent must be an object"
+        );
+      }
+      yield { ...event, topic };
+    }
+
+    // @ts-ignore code is actually fine, but async iterators are currently an ES.next feature
+    for await (const val of asyncIterator) {
+      yield val;
+    }
+  };
 }
 
 /*
@@ -46,10 +78,10 @@ const PgSubscriptionResolverPlugin: Plugin = function(builder, { pubsub }) {
       }
       return extend(field, {
         subscribe: async (
-          parent: any,
-          args: any,
-          resolveContext: any,
-          resolveInfo: any
+          parent: unknown,
+          args: { [key: string]: unknown },
+          resolveContext: unknown,
+          resolveInfo: unknown
         ) => {
           const topic =
             typeof topicGen === "function"
@@ -65,7 +97,7 @@ const PgSubscriptionResolverPlugin: Plugin = function(builder, { pubsub }) {
             typeof unsubscribeTopicGen === "function"
               ? await unsubscribeTopicGen(args, resolveContext, resolveInfo)
               : unsubscribeTopicGen;
-          let asyncIterator = pubsub.asyncIterator(topic);
+          let asyncIterator = await pubsub.asyncIterator(topic);
           if (unsubscribeTopic) {
             // Subscribe to event revoking subscription
             const unsubscribeTopics: Array<string> = Array.isArray(
@@ -102,27 +134,14 @@ const PgSubscriptionResolverPlugin: Plugin = function(builder, { pubsub }) {
                 "initialEvent provided to pgSubscription must be a function"
               );
             }
-            const oldIterator = asyncIterator;
-            asyncIterator = (async function* subscribeWithInitialEvent() {
-              const event = await initialEvent(
-                args,
-                resolveContext,
-                resolveInfo
-              );
-              if (event != null) {
-                if (typeof event !== "object") {
-                  throw new Error(
-                    "initialEvent returning event must be an object"
-                  );
-                }
-                yield { ...event, topic };
-              }
-              for await (const val of {
-                [Symbol.asyncIterator]: () => oldIterator,
-              }) {
-                yield val;
-              }
-            })();
+            const oldAsyncIterator = asyncIterator;
+            asyncIterator = withInitialEvent(topic, initialEvent)(
+              oldAsyncIterator,
+              parent,
+              args,
+              resolveContext,
+              resolveInfo
+            );
           }
 
           // filter should always be the last adjustment
@@ -134,7 +153,8 @@ const PgSubscriptionResolverPlugin: Plugin = function(builder, { pubsub }) {
                 "filter provided to pgSubscription must be a function"
               );
             }
-            return withFilter(() => asyncIterator, filter)(
+            const oldAsyncIterator = asyncIterator;
+            asyncIterator = withFilter(() => oldAsyncIterator, filter)(
               parent,
               args,
               resolveContext,
--- a/packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
+++ b/packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
@@ -96,21 +96,6 @@ const PgSubscriptionResolverPlugin: Plugin = function(builder, { pubsub }) {
             });
           }
 
-          if (filter) {
-            if (typeof filter !== "function") {
-              throw new Error(
-                "filter provided to pgSubscription must be a function"
-              );
-            }
-            const oldIterator = asyncIterator;
-            asyncIterator = withFilter(() => oldIterator, filter)(
-              parent,
-              args,
-              resolveContext,
-              resolveInfo
-            );
-          }
-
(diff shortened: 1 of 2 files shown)


packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
```diff
@@ -101,13 +102,41 @@ const PgSubscriptionResolverPlugin: Plugin = function(builder, { pubsub }) {
...
+              );
+            }
+
+            return (async function* subscribeWithInitialEvent() {
+              const event = await initialEvent(
+                args,
+                resolveContext,
+                resolveInfo
+              );
+              if (event === null || typeof event !== "object") {
```
benjie: TypeScript can infer more things if you switch this order.

```suggestion
              if (typeof event !== "object" || event === null) {
```

benjie: Actually, I think returning `null` or `undefined` should be entirely valid - we just auto-skip it.

```js
// Only send initialEvent when return value is not null/undefined.
if (event != null) {
  if (typeof event !== "object") {
    throw ...
  }
  yield { ...event, topic };
}
for await ...
```

packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
```diff
@@ -101,13 +102,41 @@ const PgSubscriptionResolverPlugin: Plugin = function(builder, { pubsub }) {
...
+                throw new Error(
+                  "initialEvent returning event must be an object"
+                );
+              }
+              yield { ...event, topic };
+              for await (const val of {
+                [Symbol.asyncIterator]: () => asyncIterator,
+              }) {
```
benjie: Is this really required? I thought that the following would work?
```suggestion
              for await (const val of asyncIterator) {
```

packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
```diff
@@ -101,13 +102,41 @@ const PgSubscriptionResolverPlugin: Plugin = function(builder, { pubsub }) {
                 "filter provided to pgSubscription must be a function"
               );
             }
-            return withFilter(() => asyncIterator, filter)(
+            asyncIterator = withFilter(() => asyncIterator, filter)(
```
benjie:
```suggestion
            const oldIterator = asyncIterator;
            asyncIterator = withFilter(() => oldIterator, filter)(
```

packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
```diff
@@ -101,13 +102,41 @@ const PgSubscriptionResolverPlugin: Plugin = function(builder, { pubsub }) {
...
           }
+
+          if (initialEvent) {
+            if (typeof initialEvent !== "function") {
+              throw new Error(
+                "initialEvent provided to pgSubscription must be a function"
+              );
+            }
+
+            return (async function* subscribeWithInitialEvent() {
```
benjie:
```suggestion
            const oldIterator = asyncIterator;
            asyncIterator = (async function* subscribeWithInitialEvent() {
```

packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
benjie:
```suggestion
              for await (const val of oldIterator) {
```

enisdenjo: It is valid, but TS complains. Instead of doing a `// @ts-ignore`, I just complied.

packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
enisdenjo: Didn't know about the TS trick, nice! Regarding the empty event, I agree.

packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
enisdenjo: https://github.com/graphile/graphile-engine/pull/612#discussion_r401530121

enisdenjo:
I don't actually understand how would you like me to split the logic up?

Filter was not working actually, but it was just a matter of switching up the order.

packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
benjie: Hmmm... We should appease TypeScript rather than using this non-obvious syntax.

benjie:
I tried opening a PR to your PR but GitHub baulked; so here's the code I ended up with:

```ts
import debugFactory from "debug";
import { Plugin } from "graphile-build";
import { PubSub, withFilter } from "graphql-subscriptions";

const debug = debugFactory("pg-pubsub");

function isPubSub(pubsub: any): pubsub is PubSub {
  return !!pubsub;
}

// TODO: add types
function withInitialEvent<T>(initialEvent) {
  if (typeof initialEvent !== "function") {
    throw new Error(
      "`initialEvent` provided to `@pgSubscription` must be a function"
    );
  }
  return (async function* subscribeWithInitialEvent<T>(
    asyncIterator: AsyncIterableIterator<T>,
    parent: any,
    args,
    resolveContext, resolveInfo
  ): AsyncIterableIterator<T> {
    const event = await initialEvent(
      args,
      resolveContext,
      resolveInfo
    );
    if (event != null) {
      if (typeof event !== "object") {
        throw new Error(
          "The event returned from `initialEvent` must be an object"
        );
      }
      yield { ...event, topic };
    }
    for await (const val of  oldIterator) {
      yield val;
    }
  }
}

/*
 * This plugin looks for the `@pgSubscription` directive, and adds the
 * `subscribe` method.
 */

const PgSubscriptionResolverPlugin: Plugin = function(builder, { pubsub }) {
  if (!isPubSub(pubsub)) {
    debug("Subscriptions disabled - no pubsub provided");
    return;
  }
  builder.hook(
    "GraphQLObjectType:fields:field",
    (field, build, graphileContext) => {
      const { extend } = build;
      const {
        scope: { isRootSubscription, fieldDirectives },
      } = graphileContext;
      if (!isRootSubscription) {
        return field;
      }
      if (!fieldDirectives) {
        return field;
      }
      const { pgSubscription } = fieldDirectives;
      if (!pgSubscription) {
        return field;
      }
      const {
        topic: topicGen,
        unsubscribeTopic: unsubscribeTopicGen,
        filter,
        initialEvent,
      } = pgSubscription;
      if (!topicGen) {
        return field;
      }
      return extend(field, {
        subscribe: async (
          parent: any,
          args: any,
          resolveContext: any,
          resolveInfo: any
        ) => {
          const topic =
            typeof topicGen === "function"
              ? await topicGen(args, resolveContext, resolveInfo)
              : topicGen;
          if (!topic) {
            throw new Error("Cannot subscribe at this time");
          }
          if (typeof topic !== "string") {
            throw new Error("Invalid topic provided to pgSubscription");
          }
          const unsubscribeTopic =
            typeof unsubscribeTopicGen === "function"
              ? await unsubscribeTopicGen(args, resolveContext, resolveInfo)
              : unsubscribeTopicGen;
          let asyncIterator: AsyncIterableIterator<any> = await pubsub.asyncIterator(topic);
          if (unsubscribeTopic) {
            // Subscribe to event revoking subscription
            const unsubscribeTopics: Array<string> = Array.isArray(
              unsubscribeTopic
            )
              ? unsubscribeTopic
              : [unsubscribeTopic];
            const unsubscribeIterators = unsubscribeTopics.map(t => {
              const i = pubsub.asyncIterator(t);
              i["topic"] = t;
              return i;
            });
            unsubscribeIterators.forEach(unsubscribeIterator => {
              unsubscribeIterator.next().then(() => {
                debug(
                  "Unsubscribe triggered on channel %s",
                  unsubscribeIterator["topic"]
                );
                if (asyncIterator.return) {
                  asyncIterator.return();
                }
                unsubscribeIterators.forEach(i => {
                  if (i.return) {
                    i.return();
                  }
                });
              });
            });
          }

          if (initialEvent) {
            asyncIterator = wrapWithInitialEvent(initialEvent)(
              asyncIterator,
              parent,
              args,
              resolveContext,
              resolveInfo
            );
          }

          // filter should always be the last adjustment
          // since we want to feed all current and future
          // custom emissions to the filter
          if (filter) {
            if (typeof filter !== "function") {
              throw new Error(
                "filter provided to pgSubscription must be a function"
              );
            }
            const oldAsyncIterator = asyncIterator;
            asyncIterator = withFilter(() => oldAsyncIterator, filter)(
              parent,
              args,
              resolveContext,
              resolveInfo
            );
          }

          return asyncIterator;
        },
        ...(field.resolve
          ? null
          : {
              resolve<T>(event: T): T {
                return event;
              },
            }),
      });
    },
    ["PgSubscriptionResolver"]
  );
};

export default PgSubscriptionResolverPlugin;
```

This was entirely written into the GitHub interface, so you're going to need to do some heavy editing, but it perhaps gives you a better idea of the changes I was intending?

benjie:
(Suggest you paste it over your code and then review the diff.)

enisdenjo:
@benjie I have made the requested changes. However, everything is `unknown`; I would suggest doing the definitions properly throughout whole `pg-pubsub`, over just a few parts.

I'd suggest opening an additional PR where the types would be introduced. I can do that when I acquire  some spare time, you agree?

packages/pg-pubsub/src/PgSubscriptionResolverPlugin.ts
```diff
@@ -4,10 +4,42 @@ import { PubSub, withFilter } from "graphql-subscriptions";
...
+    if (event != null) {
+      if (typeof event !== "object") {
+        throw new Error(
+          "event returned from pgSubscription initialEvent must be an object"
+        );
+      }
+      yield { ...event, topic };
+    }
+
+    // @ts-ignore code is actually fine, but async iterators are currently an ES.next feature
```
benjie: Seems it requires Node 10.3.0+ but we need to support Node 8.6. May need to write a heavier wrapper :disappointed:

You can leave this with me if you're not in a rush (I think you have your own fork?)

enisdenjo: No rush. Do you need anything else from me? Access to the forked repo?

benjie:
We really need some tests for this; but I don't think we have tests for any of the subscriptions things currently :(

enisdenjo:
Agreed... In my company we are transitioning all our sub modules to Graphile, so there is an ever growing need for them.

However, I am clogged with work so I cannot spend time writing them. But would be happy to when time frees up (not soon TBH).

carlobeltrame:
I was just looking for this functionality when I came across this PR. Unfortunately, I am not deep enough into the project to contribute tests myself... but nice to know that the need for this feature is common.

benjie:
(Rebased)

benjie:
Have tested this with the following plugin:

```js
const { makeExtendSchemaPlugin, gql, embed } = require("graphile-utils");
const topicFromArgs = () => {
  return `graphql:test`;
};

module.exports = makeExtendSchemaPlugin(() => ({
  typeDefs: gql`
      type TestSubscriptionPayload {
        event: String
      }
      extend type Subscription {
        test: TestSubscriptionPayload @pgSubscription(
          topic: ${embed(
            topicFromArgs
          )}
          initialEvent: ${embed(
            () => {return {event: 'initialev'};}
          )}
        )
      }
    `,
  resolvers: {},
}));
```

And the query:

```graphql
subscription MySubscription {
  test {
    event
  }
}
```

And you can later do a manual notify:

```sql
NOTIFY "graphql:test", '{"event":"hello"}';
```

Everything seems to be in order; even if we return `null` from the initialEvent callback, or omit it entirely.

Issue #600, opened 2020-02-07
Feature request: add input support for PostgreSQL domain over a composite type


jcgsville:
I'm submitting a bug report

PostGraphile version: v4.5.5

Minimal SQL file that can be loaded into a clean database:

```sql
drop schema if exists my_schema cascade;
drop domain if exists two_ints_constrained;
drop type if exists two_ints;

create schema my_schema;

create type two_ints as (
	one		integer,
	two     integer
);

create domain two_ints_constrained as two_ints NOT NULL;

create table my_schema.foo (
	f 		two_ints_constrained,
	g		integer,
	UNIQUE (f, g)
);
```

Steps to reproduce:

Run `postgraphile -c postgres://postgres@localhost:5432/my_db -s my_schema`

Current behavior:
It fails to build the schema with the following error

```
 serious error occurred when building the initial schema. Exiting because `retryOnInitFail` is not set. Error details:

Error: Could not find input type for key 'f' on type 'Foo'
    at args.keysIncludingMeta.reduce (/path/to/postgraphile/node_modules/graphile-build-pg/node8plus/plugins/PgRowByUniqueConstraint.js:108:25)
```

Expected behavior:
It behaves the same as if the the table definition does not use the domain, and just the composite type. An example is the following: (this works no problem)

```sql
create table my_schema.foo (
	f 		two_ints,
	g		integer,
	UNIQUE (f, g)
);
```

benjie:
We need a block like this:

https://github.com/graphile/graphile-engine/blob/a9746f1d444fa6a39ab18cd983776cd68792c83e/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js#L765-L797

But for input types. PR welcome :+1:

ben-pr-p:
Should this be closed with #615 merged?

PR #615, opened 2020-04-08
feat(pg): support domain constrained composite types

jcgsville:
Addresses https://github.com/graphile/graphile-engine/issues/600

Input types now support domains that constrain composite types instead of failing to generate the schema on startup.

I didn't add any testing as I was a bit unclear what the testing story is for a change like this. Happy to make any additions in this area with a quick point in the right direction.


benjie:
Thanks @jcgsville; this is effectively a new type so we should add it to `b.types`:

https://github.com/graphile/graphile-engine/blob/e863fd37bddd1a7d40bd6e5e90f5065d1e64949b/packages/postgraphile-core/__tests__/kitchen-sink-schema.sql#L237

Also add a mutation test for it in here:

https://github.com/graphile/graphile-engine/blob/v4/packages/postgraphile-core/__tests__/fixtures/mutations/types.graphql

And a query test here:

https://github.com/graphile/graphile-engine/blob/v4/packages/postgraphile-core/__tests__/fixtures/queries/types.graphql

These are all integration tests, effectively.

Changes to the above will require a lot of snapshot updates, you can run `yarn --cwd packages/postgraphile-core jest -u` to update them. You may need to fix a few other tests too - let the test results guide you.

Thanks!

jcgsville:
Thanks for the guidance @benjie! I added some tests, which pass with Postgres 11, but it looks like Postgres 10 and earlier did not let you create a domain over a composite type, so the tests fail there. Is there a mechanism testing things only in certain versions?

benjie:
@jcgsville Not a good one, but we did have to avoid PG10 tests running in PG9.6: https://github.com/graphile/graphile-engine/blob/9dca5c8631e6c336b59c499d901c774d41825c60/packages/postgraphile-core/scripts/test#L47-L54

I'm not happy with our testing system in general (it's very legacy, clunky and slow), and am planning to overhaul it as part of the work I'm doing on V5.

benjie:
@jcgsville Are you able to add a relevant workaround for PG11 as indicated above?

benjie:
Merged `v4` branch and then moved the test of this feature to the PG11 test suite.

benjie (approved):
Thanks for the fix :raised_hands:

jcgsville:
Sorry to have not gotten to the testing tweaks sooner. Thanks for ferrying it the rest of the way through 🙂

benjie:
No worries 🙌

PR #634, opened 2020-07-29
chore(deps): update a few dependencies in yarn.lock

benjie:


benjie:
Frustratingly these semver-compatible changes have caused Node compatibility issues. E.g. the `semver` module itself now requires Node>=10; so something has clearly jumped a major version here.

I'm tempted to drop Node 8 support given it's been out of support since December. I really don't want to do a major bump to PostGraphile purely so we can update some dependencies :roll_eyes:

benjie:
Took a more selective approach.

PR #652, opened 2020-09-08
feat: add UUID validation so errors are raised earlier

### LINKED ISSUE #644: Insufficient validation of UUID type
If a user submits a query, passing an invalid UUID value such as `"undefined"` or `"27"`, we currently pass this through to Postgres as a string, and have Postgres cast it to UUID, which fails at runtime, causing the entire SQL query to fail. This may cause GraphQL clients to make invalid assumptions about the data since the error occurs at the root field rather than the nested field, potentially triggering a client to update their cache to say that root field is invalid.

To solve this, we should make sure that we'd expect PostgreSQL to be able to coerce the value to UUID before actually passing it to PostgreSQL, and if not then we should throw the relevant error in the relevant place.

benjie:
Previously if a user submitted a query, passing an invalid UUID value such as `"undefined"` or `"27"`, we would pass this through to Postgres as a string, and have Postgres cast it to UUID, which fails at runtime, causing the entire SQL query to fail within the database. This may cause GraphQL clients to make invalid assumptions about the data since the error occurs at the root field rather than the nested field, potentially triggering a client to update their cache to say that root field is invalid.

This PR solves this by making sure that we'd expect PostgreSQL to be able to coerce the value to UUID before actually passing it to PostgreSQL, throwing a scalar coercion error otherwise and not allowing the query to reach the database.

Fixes #644

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
@@ -214,17 +214,23 @@ export default (function PgTypesPlugin(
       );
       addType(GQLIntervalInput, "graphile-build-pg built-in");
 
-      const stringType = (name, description) =>
+      const stringType = (name, description, coerce) =>
         new GraphQLScalarType({
           name,
           description,
           serialize: value => String(value),
-          parseValue: value => String(value),
+          parseValue: coerce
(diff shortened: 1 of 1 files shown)


heri16:
This doesn't work for us as some of our uuids are not RFC 4122, but are md5 digests.

PR #678, opened 2020-10-16
fix(types): fix parsing interval to support fractional seconds

benjie:
## Description

@keithlayne raised a PR #676 noting that fractional seconds were being truncated by PostGraphile. Upon further investigation I determined that the `postgres-interval` module was doing a lot more than we needed, so I've removed it and replaced it with a lean reimplementation of just the parse function (which should also have better performance, lower memory footprint, and smaller bundle size).

Fixes #676 

## Performance impact

Ever so slight improvement.

## Security impact

None known.


benjie:
I've [raised an issue](https://github.com/bendrucker/postgres-interval/issues/33) with `postgres-interval` to see if they'd like the changes upstreamed.

PR #687, opened 2020-11-09
Send a manual notification at the end of watch-fixtures.sql

mathroc:
If the postgraphile server is started before the aatch-fixtures.sql file was loaded, this allows it to rebuild the GraphQL schema and then pick up on changes

useful when executed from the graphile-migrate afterReset hook


benjie:
I don't think I want to add this as it could cause unnecessary schema rebuilds when running multiple PostGraphile instances; might be better if you just add that in a second afterReset hook?

mathroc:
> I don't think I want to add this as it could cause unnecessary schema rebuilds when running multiple PostGraphile instances

ah yes I did not think about that

> might be better if you just add that in a second afterReset hook?

:+1:

PR #689, opened 2020-11-11
feat(deps): support for GraphQL v15

benjie:
Adds support for GraphQL v15; to minimise the schema diffs it also wraps all of our built-in descriptions (to match GraphQL v14's output).

Performance and security impact are dictated by GraphQL v15 vs GraphQL v14 - but ultimately this is the user's choice so is beyond the scope of this PR.

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgRowNode.js
+++ b/packages/graphile-build-pg/src/plugins/PgRowNode.js
@@ -153,11 +153,17 @@ export default (async function PgRowNode(builder, { subscriptions }) {
                   fieldName,
                   ({ getDataFromParsedResolveInfoFragment }) => {
                     return {
-                      description: `Reads a single \`${TableType.name}\` using its globally unique \`ID\`.`,
+                      description: build.wrapDescription(
+                        `Reads a single \`${TableType.name}\` using its globally unique \`ID\`.`,
+                        "field"
+                      ),
                       type: TableType,
                       args: {
                         [nodeIdFieldName]: {
-                          description: `The globally unique \`ID\` to be used in selecting a single \`${TableType.name}\`.`,
+                          description: build.wrapDescription(
+                            `The globally unique \`ID\` to be used in selecting a single \`${TableType.name}\`.`,
+                            "field"
+                          ),
                           type: new GraphQLNonNull(GraphQLID),
                         },
                       },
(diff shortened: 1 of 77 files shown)


benjie:
I've added code that wraps all our old descriptions so that the schema diff is reduced there; ~~I'm now going to sort all the schemas that use `buildClientSchema` which now seems to trigger the schema to be ordered~~.

packages/graphile-build-pg/src/plugins/PgRowNode.js
```diff
@@ -153,11 +153,17 @@ export default (async function PgRowNode(builder, { subscriptions }) {
...
+                        `Reads a single \`${TableType.name}\` using its globally unique \`ID\`.`,
+                        "field"
+                      ),
                       type: TableType,
                       args: {
                         [nodeIdFieldName]: {
-                          description: `The globally unique \`ID\` to be used in selecting a single \`${TableType.name}\`.`,
+                          description: build.wrapDescription(
+                            `The globally unique \`ID\` to be used in selecting a single \`${TableType.name}\`.`,
+                            "field"
```
benjie: Yay I found a bug, reading all this code was worthwhile :laughing: 
```suggestion
                            "arg"
```

packages/graphile-build/src/plugins/NodePlugin.js
```diff
@@ -222,11 +229,17 @@ export default (function NodePlugin(
...
+                "Fetches an object given its globally unique `ID`.",
+                "field"
+              ),
               type: getTypeByName(inflection.builtin("Node")),
               args: {
                 [nodeIdFieldName]: {
-                  description: "The globally unique `ID`.",
+                  description: build.wrapDescription(
+                    "The globally unique `ID`.",
+                    "field"
```
benjie: And another!
```suggestion
                    "arg"
```

benjie (commented):
I've reviewed everything except packages/graphile-utils/__tests__/__snapshots__/ExtendSchemaPlugin-pg.test.js.snap, packages/graphile-utils/__tests__/__snapshots__/makeAddPgTableConditionPlugin.test.js.snap and packages/postgraphile-core/__tests__/integration/schema/__snapshots__/simplePrint.test.js.snap. Those diffs are too large to sensibly compare given GraphQL v15 reorders the schema (this is an argument for lexicographically sorting the GraphQL schema, but we've chosen not to do this because we want to know _where_ we're adding the fields).

Issue #691, opened 2020-11-12
Adding default condition doesn't work as expected (requires change to GraphQL Spec)


benjie:
Adding a default value for a condition (e.g. `fields.includeArchived = { type: GraphQLBoolean, defaultValue: false}`) doesn't apply unless you pass an empty condition object (e.g. `{myCollection { ... } }` doesn't respect it, but `{myCollection(condition: {}) { ... } }` does). However this cannot be solved by setting the `defaultValue` of the `condition` argument to be `{}`; though this gets correctly exposed in GraphiQL:

![Screenshot_20201112_132612](https://user-images.githubusercontent.com/129910/98945605-a559b900-24ea-11eb-84be-d8d47555159f.png)

The results of running a query show that this condition wasn't actually applied, since the two queries above would produce different results when they should not.

I don't know why this is, but at the moment I'm leaning towards it being a bug in the way that Graphile Engine processes arguments (perhaps not applying defaults recursively?). It might actually be a GraphQL thing, but that has yet to be determined.

Interestingly, if you set the default value for the `condition` argument to `defaultValue: {includeArchived: false}` then it works as it should, but it should not be necessary to do this additional lookup.

Warrants further investigation. Also, once solved, we should give `condition` a default value of `{}` so these default conditions can apply automatically.

benjie:
This is, indeed, an issue in GraphQL itself (arguably in the spec)

To reproduce:

```
const {
  graphqlSync,
  GraphQLSchema,
  GraphQLObjectType,
  GraphQLInputObjectType,
  GraphQLInt,
} = require("graphql");

const ExampleInputObject = new GraphQLInputObjectType({
  name: "ExampleInputObject",
  fields: {
    number: {
      type: GraphQLInt,
      defaultValue: 3,
    },
  },
});

const Query = new GraphQLObjectType({
  name: "Query",
  fields: {
    example: {
      args: {
        inputObject: {
          type: ExampleInputObject,
          default: {},
        },
      },
      type: GraphQLInt,
      resolve(source, { inputObject }) {
        return inputObject && inputObject.number;
      },
    },
  },
});

const schema = new GraphQLSchema({
  query: Query,
});

// All three of these should be equivalent?
const result1 = graphqlSync(schema, `{example}`);
const result2 = graphqlSync(schema, `{example(inputObject:{})}`);
const result3 = graphqlSync(schema, `{example(inputObject:{number: 3})}`);

console.log(JSON.stringify(result1.data));
[...]

PR #692, opened 2020-11-13
feat: allow for conditions with default values

benjie:
## Description

Changes the `condition` argument so that it has a `defaultValue` of `{}` (but [coerced to a value respecting child defaultValues](https://github.com/graphql/graphql-js/issues/385#issuecomment-726741458)), this means that `defaultValue` inside of a `*Condition` will be respected.

## Performance impact

TBD.

## Security impact

None known.

## Status

On hold; needs more thought. See discussion: https://github.com/graphile/graphile-engine/issues/691


benjie:
This is unlikely to make it into V4 because it's still awaiting fixes to the GraphQL spec. Closing.

PR #709, opened 2021-01-26
Allow more specific `TArgs` types

### LINKED ISSUE #708: Improve `makeWrapResolversPlugin` typing for TS
### Summary

Specifying any args in `makeWrapResolversPlugin` that is more narrow than `any` fails typechecking with

```
Property 'input' is missing in type '{ [argName: string]: any; }' but required in type '{ input: $Typename; }'.
```

### Steps to reproduce

Here's a simple example:

```typescript
const getPointFromGMaps => {
  const getLatitudeAndLongitude = async (
    resolve: any,
    _source: any,
    args: {
      input: CreateDestinationInput; // This is a type generated by `graphql-codegen`
    },
    _context: any,
    _resolveInfo: GraphQLResolveInfo
  ) => {
    // Code omitted for brevity
  };
  return makeWrapResolversPlugin({
    Mutation: {
      updateDestination: getLatitudeAndLongitude,
      createDestination: getLatitudeAndLongitude
    }
  });
};
```

### Expected results

This should typecheck without issue.

### Actual results

The example above fails with 
```
Property 'input' is missing in type '{ [argName: string]: any; }' but required in type '{ input: CreateDestinationInput; }'.
```

### Possible Solution

It seems like the issue is due to the fact that the type parameters of `ResolverWrapperFn` gets instantiated as `any, any, { [argName: string]: any }`. It _seems_ like this can be fixed by forwarding the type parameters from `makeWrapResolversPlugin` (and by setting the type parameter defaults in the same way as in `ResolverWrapperFn` so we don't break existing code) to `ResolverWrapperFn`.

I have a PoC that I can submit a PR for if you want. Thanks in advance!

blakepettersson:
## Description

Allow types more specific than `{ [argName: string]: any }` to be
passed into `makeWrapResolversPlugin`. We now set `TSource`, `TContext`
and `TArgs` as type parameters to `makeWrapResolversPlugin` which
means we can optionally pass in types more narrow than the defaults.

This _should_ be a backwards compatible change since we set the type
defaults to be the same as in `ResolverWrapperFn`. Fixes #708.

## Performance impact

Should be zero.

## Security impact

Should be zero.

## Checklist

### DIFF
--- a/packages/graphile-utils/src/makeWrapResolversPlugin.ts
+++ b/packages/graphile-utils/src/makeWrapResolversPlugin.ts
@@ -143,7 +143,7 @@ export default function makeWrapResolversPlugin<T>(
       if (!resolveWrapperOrSpec) {
         return field;
       }
-      const resolveWrapper: ResolverWrapperFn<unknown, unknown> | undefined =
+      const resolveWrapper: ResolverWrapperFn | undefined =
         typeof resolveWrapperOrSpec === "function"
           ? resolveWrapperOrSpec
           : resolveWrapperOrSpec.resolve;
--- a/packages/graphile-utils/src/makeWrapResolversPlugin.ts
+++ b/packages/graphile-utils/src/makeWrapResolversPlugin.ts
@@ -27,19 +27,33 @@ interface ResolverWrapperRequirements {
   siblingColumns?: Array<{ column: string; alias: string }>;
 }
 
-interface ResolverWrapperRule {
+interface ResolverWrapperRule<
+  TSource extends unknown = any,
+  TContext extends unknown = any,
+  TArgs extends unknown = { [argName: string]: any }
+> {
   requires?: ResolverWrapperRequirements;
-  resolve?: ResolverWrapperFn;
+  resolve?: ResolverWrapperFn<TSource, TContext, TArgs>;
   // subscribe?: ResolverWrapperFn;
 }
 
-interface ResolverWrapperRules {
+interface ResolverWrapperRules<
+  TSource extends unknown = any,
+  TContext extends unknown = any,
+  TArgs extends unknown = { [argName: string]: any }
+> {
   [typeName: string]: {
-    [fieldName: string]: ResolverWrapperRule | ResolverWrapperFn;
+    [fieldName: string]:
+      | ResolverWrapperRule<TSource, TContext, TArgs>
+      | ResolverWrapperFn<TSource, TContext, TArgs>;
   };
 }
 
-type ResolverWrapperRulesGenerator = (options: Options) => ResolverWrapperRules;
+type ResolverWrapperRulesGenerator<
+  TSource extends unknown = any,
+  TContext extends unknown = any,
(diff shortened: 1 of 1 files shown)


packages/graphile-utils/src/makeWrapResolversPlugin.ts
```diff
@@ -27,19 +27,33 @@ interface ResolverWrapperRequirements {
...
+interface ResolverWrapperRules<
+  TSource extends unknown = any,
+  TContext extends unknown = any,
+  TArgs extends unknown = { [argName: string]: any }
+> {
   [typeName: string]: {
-    [fieldName: string]: ResolverWrapperRule | ResolverWrapperFn;
+    [fieldName: string]:
+      | ResolverWrapperRule<TSource, TContext, TArgs>
+      | ResolverWrapperFn<TSource, TContext, TArgs>;
```
benjie: I'm not sure this really makes sense. TContext will be the same for all resolvers, but TSource and TArgs will be field specific, so in this two dimensional map it wouldn't give a way to indicate TSource and TArgs for each individual field. e.g.

```graphql
type User {
  name(short: Boolean = false): String
  age(in: DurationEnum!): Int
}
type Post {
  digest: String
}
```

```js
const rules = {
  User: {
    name: {
      requires: { siblingColumns: [
        {column: 'first_name', alias: '$firstName'},
        {column: 'lastst_name', alias: '$lastName'}
      ]},
      resolver(
        soruce: {$firstName: string, $lastName: string},
        args: {short: boolean},
        context: TContext
      ) { ... }
    }
    
    age: {
      requires: { siblingColumns: [
        {column: 'date_of_birth', alias: '$dob'}
      ]},
      resolver(
        source: {$dob: Date},
        args: {in: DurationEnum},
        context: TContext
      ) { ... }
    },
  },
  Post: { ... }
}
```

Here we can see the args and source types are different for each field.

benjie (changes requested):
Thanks for taking a look into this, but I'm not sure this solution works as currently proposed.

packages/graphile-utils/src/makeWrapResolversPlugin.ts
blakepettersson: Hmm... good point. I pushed some changes; for the inference to properly work it seems like we'd need to do some changes to `ResolverWrapperRules`. If we change `ResolverWrapperRules` to a type instead of an interface we can use `in keyof` to go through all the keys of the `ResolverWrapperRules` instance, and then use conditional types to infer the generic type parameters of `ResolverWrapperFn` (or `ResolverWrapperRule`). 

These changes aren't quite backwards compatible though...

benjie: Yeah; they'll need to wait for V5 I'm afraid. Thanks anyway!

benjie:
PostGraphile V5 is in beta; if you're at all interested in porting your work on this into V5 please reach out; but for now I'm going to close this.

PR #711, opened 2021-01-29
fix(watch): fix pool leak on unexpected connection termination

### LINKED ISSUE #710: Connection pool leak when unexpected tcp reset occurs
https://github.com/graphile/graphile-engine/blob/5cd51b66ce1a9ecd0f2574e0105a0d6cbfecbb75/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js#L1192

This comment is wrong; we are running postgraphile in kubernetes, with istio-sidecar enabled.  By default, this sidecar/proxy will force-close the connection after 1 hour.  After 10 such occurrences, all database activity stops.  In the linked function, the pgClient/releasePgClient are set to null, then a reconnect occurs.  Because they are null, no cleanup can happen(ie, releaseClient is never called, and unwatch never happens.

If I comment out the 2 lines, then there is no pooled connection leak.

I then changed the istio timeout from the default of 1 hour, down to something much quicker, 30 seconds.  I get the reconnect in 2s error, but the daemon keeps functioning after 10 such occurrences.

benjie:
## Description

The `Client is already cleaned up` comment was incorrect; this fixes the release behaviour on error.

- `_handleChange` is now only cancelled in `stop()`; it doesn't seem to be necessary to cancel it in error handler
- When releasing the client, you can pass `false` to prevent issuing the `unlisten` call; this is risky as it may leak event listeners.

Need to be sure that `client.on('error', ...)` only fires when the client connection is closed... I'm not sure if this is the case, may need to dig through the `pg` source code to find out.

Fixes #710 

## Performance impact

None.

## Security impact

Fixes a potential denial of service issue with watch mode (note that watch mode is _still_ not recommended for production).

## Checklist

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
@@ -1204,9 +1204,7 @@ Original error: ${e.message}
 
     _handleClientError: (e: Error) => void;
     _handleClientError(e) {
-      // Client is already cleaned up
-      this.client = null;
-      this._reallyReleaseClient = null;
+      this._releaseClient(false);
       this._reconnect(e);
     }
     async _reconnect(e) {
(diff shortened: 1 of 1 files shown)


benjie:
> Need to be sure that client.on('error', ...) only fires when the client connection is closed... I'm not sure if this is the case, may need to dig through the pg source code to find out.

I think yes, because these are the only `emit('error')`s and they both relate to connection issues.:

- https://github.com/brianc/node-postgres/blob/80c500ffbffff8c2445dce44661e85590dc026e3/packages/pg/lib/client.js#L310-L320
- https://github.com/brianc/node-postgres/blob/80c500ffbffff8c2445dce44661e85590dc026e3/packages/pg/lib/client.js#L307

PR #716, opened 2021-02-08
Add support for SSL in LDS / Live Queries

msand:
## Description

Adds support for SSL in LDS / Live Queries as boolean option liveSSL

## Performance impact

Only if used, normal SSL overhead

## Security impact

Improves security

## Checklist

### DIFF
--- a/packages/lds/src/index.ts
+++ b/packages/lds/src/index.ts
@@ -10,6 +10,7 @@ export interface Options {
   tablePattern?: string;
   sleepDuration?: number;
   temporary?: boolean;
+  liveSSL?: boolean;
 }
 
 const DROP_STALE_SLOTS_INTERVAL = 15 * 60 * 1000;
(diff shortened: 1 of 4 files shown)


msand:
Seems this can be used instead:

```javascript
const querystring = require('querystring'); // https://nodejs.org/api/querystring.html

// See https://github.com/brianc/node-postgres/tree/master/packages/pg-connection-string
const params = {
  ...(CA
    ? {
        ssl: true,
        sslmode: 'verify-full',
        sslrootcert: caFilename,
        sslcert: certFileName,
        sslkey: keyFileName,
      }
    : {}),
};
const connStringParams = querystring.stringify(params);
const connectionString = `${DSN}?${connStringParams}`;
```

benjie:
Yes; I would indicate the SSL parameters via the connection string in the same way that you can for the main PostGraphile connection string. Thank you for providing a demonstration :+1:

PR #720, opened 2021-02-25
Lodash dependency on a diet

miles-po:
## Description

The complete lodash adds quite a bit of heft to node_modules. Many of the transitive dependencies aren't even used by postgraphile & related. By using the modularized lodash dependencies, hoping to slim things down a bit for use in more space-constrained environments like AWS Lambda. Lodash unpacked is 1.4MB. This should drop that number by 80-90%.

## Performance impact

Should have exact feature/code parity. No features added or removed.

## Security impact

yarn.lock was predictably updated with more up-to-date patched versions of dependencies.

## Checklist

### DIFF
--- a/packages/graphile-build-pg/package.json
+++ b/packages/graphile-build-pg/package.json
@@ -42,7 +42,12 @@
     "debug": "^4.1.1",
     "graphile-build": "4.12.0-alpha.0",
     "jsonwebtoken": "^8.5.1",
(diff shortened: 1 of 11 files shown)


benjie:
Vetting this many dependencies is too much maintenance work. When releasing to lambda, use webpack or similar to form a smaller bundle as demonstrated in the pinned lambda example at https://github.com/graphile

PR #722, opened 2021-03-16
feat: export pgSmartTagRulesFromJSON

erichosick:
## Description

We're building out a JSON schema -> Smart Tags File utility. We want to verify that the Smart Tags File fully compiles so we make a call to makeJSONPgSmartTagsPlugin. If that call fails, we want to throw the exception and stop building the Smart Tags File. Currently, there are just warning messages.

Exporting the pgSmartTagRulesFromJSON function gives the ability to do this.

## Performance impact

There should be no impact on performance.

## Security impact

unknown

## Checklist

- [X] My code matches the project's code style and `yarn lint:fix` passes.

### DIFF
--- a/packages/graphile-utils/src/makePgSmartTagsPlugin.ts
+++ b/packages/graphile-utils/src/makePgSmartTagsPlugin.ts
@@ -240,7 +240,7 @@ export type JSONPgSmartTags = {
   };
 };
 
-function pgSmartTagRulesFromJSON(
+export function pgSmartTagRulesFromJSON(
   json: JSONPgSmartTags | null
 ): PgSmartTagRule[] {
   if (!json) {


packages/graphile-utils/src/makePgSmartTagsPlugin.ts
```diff
@@ -240,7 +240,7 @@ export type JSONPgSmartTags = {
   };
 };
 
-function pgSmartTagRulesFromJSON(
+export function pgSmartTagRulesFromJSON(
```
benjie, line 243: You also need to export this from `index.ts`

erichosick: Doh. Does `pgSmartTagRulesFromJSON` get exported when we `export * from "./makePgSmartTagsPlugin";` in [graphile-utils/src/index.ts#L22](https://github.com/graphile/graphile-engine/blob/9940895db5a35f218484620d610845201278611c/packages/graphile-utils/src/index.ts#L22)?

benjie: So it does; I generally have a policy of not using `export *` (because it doesn't work well with TypeScript declaration merging) so wasn't expecting that. :man_shrugging:

benjie (approved):

Issue #740, opened 2021-05-14
"Expected SQL item" bug when using edge field with order by computed column on mutation payload


benjie:
### Summary

The edge field on mutation payloads throws an error when you attempt to use it with an orderBy that includes a computed column (with `@sortable` smart tag).

### Steps to reproduce

See: https://github.com/pszafer/postgraphile_errors/tree/delete_error

### Expected results

No error, correct edge/cursor returned.

### Actual results

```
"Expected SQL item, instead received '({\n        queryBuilder\n      }) => sql.fragment`(${sql.identifier(proc.namespaceName, proc.name)}(${queryBuilder.getTableAlias()}))`'.",
```

### Additional context

PostGraphile v4.11.0.

### Possible Solution

The error lies here:

https://github.com/graphile/graphile-engine/blob/b715f6d1be7dd19e39830b4329fa3c758e5ae91b/packages/graphile-build-pg/src/plugins/PgMutationPayloadEdgePlugin.js#L170-L174

Specifically we've assumed if it's not a string then it's an SQL node, but that's not the case - it can also be a thunk.

Normally when ordering we call `queryBuilder.orderBy` with a value like the above:

https://github.com/graphile/graphile-engine/blob/1bc8cfefdab7a61fd7ad287bcdff66298352e308/packages/graphile-build-pg/src/plugins/PgConnectionArgOrderBy.js#L165

QueryBuilder automatically calls thunks when generating the compiled data:

https://github.com/graphile/graphile-engine/blob/1bc8cfefdab7a61fd7ad287bcdff66298352e308/packages/graphile-build-pg/src/QueryBuilder.js#L870-L875

which we've not done in the edge plugin. This is done because it passes a reference to the context: 

https://github.com/graphile/graphile-engine/blob/1bc8cfefdab7a61fd7ad287bcdff66298352e308/packages/graphile-build-pg/src/QueryBuilder.js#L138-L140

which includes the QueryBuilder so that orders may use things like the query builder's alias whilst building their fragments.

The PgMutationPayloadEdgePlugin does have access to a QueryBuilder so we should be able to fix this by changing the code above to something along the lines of:

```js
const expr = isString(col)
  ? sql.fragment`${queryBuilder.getTableAlias()}.${sql.identifier(col)}`
  : typeof col === 'function'
  ? col({ queryBuilder })
  : col;
```

benjie:
[...]

PR #741, opened 2021-05-14
fix(order): fix order by computed column SQL item bug

pszafer:
## Description

Issue: https://github.com/graphile/graphile-engine/issues/740
## Checklist

Passes with multiple warning from different files.


packages/postgraphile-core/__tests__/fixtures/mutations/mutation-delete.graphql
```diff
@@ -62,3 +63,19 @@ fragment deletePersonPayload2 on DeletePersonPayload {
   }
   query { nodeId }
 }
+
+fragment deletePersonPayloadWithEdge on DeletePersonPayload {
+  clientMutationId
+  deletedPersonId
+  personEdge(orderBy: FIRST_NAME_ASC) {
+    node {
```
benjie, line 72: when requesting an edge we should also request the cursor (otherwise what’s the point?)

```suggestion
    cursor
    node {
```

benjie (changes requested):

benjie:
@pszafer You've added `cursor` but I think you need to update the snapshots (`yarn test -u`) to update the snapshots?

benjie (approved):

pszafer:
sorry I missed your message. I see it's merged already. Thanks!

PR #748, opened 2021-06-14
feat(pg): use table enums in functions via domain

ben-pr-p:
This is still a draft, still figuring out how to run and use the tests, but this is working via local testing.

Your domain needs to be named either `my_enum_table_name_enum_domain` or you need a smart comment saying `@enum my_enum_table_name`.

If your enum table is on a unique constraint, this will work all the same, it just needs to be `my_enum_table_constraint_name_enum_domain` or `@enum my_enum_table_constraint_name`.

## Description

See https://github.com/graphile/postgraphile/issues/1500.

## Performance impact

Negligible, a single if statement. 

## Security impact

Unknown

## Checklist

- [ ] I've added tests for the new feature, and `yarn test` passes. (TODO)
- [ ] I have detailed the new feature in the relevant documentation. (TODO)
- [ ] I have added this feature to 'Pending' in the `RELEASE_NOTES.md` file (if one exists). (TODO)

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
@@ -754,6 +754,9 @@ export default (function PgTypesPlugin(
             ] = gqlInputType;
           }
         }
+
+        let shouldSkipAddType;
+
         // Enums
         if (
           !gqlTypeByTypeIdAndModifier[type.id][typeModifierKey] &&
@@ -997,31 +1000,58 @@ end`;
           type.type === "d" &&
           type.domainBaseTypeId
         ) {
-          const baseType = getGqlTypeByTypeIdAndModifier(
-            type.domainBaseTypeId,
-            typeModifier
-          );
-          const baseInputType = getGqlInputTypeByTypeIdAndModifier(
-            type.domainBaseTypeId,
-            typeModifier
-          );
-          // Hack stolen from: https://github.com/graphile/postgraphile/blob/ade728ed8f8e3ecdc5fdad7d770c67aa573578eb/src/graphql/schema/type/aliasGqlType.ts#L16
-          gqlTypeByTypeIdAndModifier[type.id][typeModifierKey] = Object.assign(
-            Object.create(baseType),
-            {
-              name: inflection.domainType(type),
-              description: type.description,
-            }
-          );
-          if (baseInputType && baseInputType !== baseType) {
+          // might be used as an enum alias: #1500
+          if (
+            type.name.endsWith("_enum_domain") ||
+            (type.tags && type.tags.enum)
+          ) {
+            const underlyingEnumTypeName = type.name.endsWith("_enum_domain")
+              ? type.name.replace("_enum_domain", "")
+              : type.tags.enum;
+
+            const baseTypeId = `FAKE_ENUM_${type.namespaceName}_${underlyingEnumTypeName}`;
+
+            const baseType = getGqlTypeByTypeIdAndModifier(
+              baseTypeId,
+              typeModifierKey
+            );
+
+            const baseInputType = getGqlInputTypeByTypeIdAndModifier(
+              baseTypeId,
+              typeModifierKey
+            );
+
+            gqlTypeByTypeIdAndModifier[type.id][typeModifierKey] = baseType;
             gqlInputTypeByTypeIdAndModifier[type.id][
               typeModifierKey
-            ] = Object.assign(Object.create(baseInputType), {
-              name: inflection.inputType(
-                gqlTypeByTypeIdAndModifier[type.id][typeModifierKey]
-              ),
+            ] = baseInputType;
+            shouldSkipAddType = true;
+          } else {
+            const baseType = getGqlTypeByTypeIdAndModifier(
+              type.domainBaseTypeId,
+              typeModifier
+            );
+            const baseInputType = getGqlInputTypeByTypeIdAndModifier(
+              type.domainBaseTypeId,
+              typeModifier
+            );
+            // Hack stolen from: https://github.com/graphile/postgraphile/blob/ade728ed8f8e3ecdc5fdad7d770c67aa573578eb/src/graphql/schema/type/aliasGqlType.ts#L16
+            gqlTypeByTypeIdAndModifier[type.id][
+              typeModifierKey
+            ] = Object.assign(Object.create(baseType), {
+              name: inflection.domainType(type),
               description: type.description,
             });
+            if (baseInputType && baseInputType !== baseType) {
+              gqlInputTypeByTypeIdAndModifier[type.id][
+                typeModifierKey
+              ] = Object.assign(Object.create(baseInputType), {
+                name: inflection.inputType(
+                  gqlTypeByTypeIdAndModifier[type.id][typeModifierKey]
+                ),
+                description: type.description,
+              });
+            }
           }
         }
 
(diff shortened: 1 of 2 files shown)


packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
```diff
@@ -997,31 +1000,58 @@ end`;
...
+
+            gqlTypeByTypeIdAndModifier[type.id][typeModifierKey] = baseType;
             gqlInputTypeByTypeIdAndModifier[type.id][
               typeModifierKey
-            ] = Object.assign(Object.create(baseInputType), {
-              name: inflection.inputType(
-                gqlTypeByTypeIdAndModifier[type.id][typeModifierKey]
-              ),
+            ] = baseInputType;
+            shouldSkipAddType = true;
```
benjie, line 1032: This feels messy but I'm not sure how to solve it:

1. use `shouldSkipAddType` (messy)
2. return here directly (unexpected given rest of function does not)
3. have `addType` not throw in this circumstance (might hide programming error, though not sure what...?)

None of these are the clear winner but I think we should go with option 3 - let's make it so that `addType` does not throw if it receives the exact identical type again `(type1 === type2)`. Should be able to achieve this with a `Set` which is supported in Node 8 so we can use it: https://node.green/#ES2015-built-ins-Set

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
```diff
@@ -997,31 +1000,58 @@ end`;
...
+          // might be used as an enum alias: #1500
+          if (
+            type.name.endsWith("_enum_domain") ||
+            (type.tags && type.tags.enum)
+          ) {
+            const underlyingEnumTypeName = type.name.endsWith("_enum_domain")
+              ? type.name.replace("_enum_domain", "")
+              : type.tags.enum;
+
+            const baseTypeId = `FAKE_ENUM_${type.namespaceName}_${underlyingEnumTypeName}`;
```
benjie: This feels very brittle; can we instead use getGqlTypeByPgTypeAndModifier (or similar) passing the type ID for the referenced enum table? This allows me to change the enum identifier without keeping particular lines of code in sync.

benjie (changes requested):
Looking good! 👍

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
ben-pr-p: I did 1 because I did not want to do 2 (the function is build as setting up a bunch of stuff, and then returning 1 thing, and I didn't want to violate that).

I considered 3, but decided not to do it because I recognized it as an error I found useful many times in development (if I tried to create a table `string` for example, I'd get a (good) error here about how String was already a registered type.

Is there a way to make `addType` not throw that doesn't also prevent the good things it catches?

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
ben-pr-p: Based on my investigation, the ID for the referenced enum table wasn't actually available in the introspection results. Looks like it's substituted here: https://github.com/graphile/graphile-engine/blob/v4/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js#L580 and here https://github.com/graphile/graphile-engine/blob/v4/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js#L557

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
benjie: I think you may have misinterpreted my suggestion: addType would still throw on a name conflict, it just won’t throw if the exact type is already known; i.e. it becomes idempotent (`addType(addType(type))` would become allowed)

benjie: Or maybe I misinterpreted your reply and you’re referring to the fallback String… in which case I can see how that would be useful. I’ve never faced this issue myself.

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
benjie: You might be right; I’ll check later in the week 👍

benjie: Yeah you're right; let's stick with using the enum fake name. With that in mind, please could you extract the generation of this value into a function that can be called in both places; something like:

```js
const fakeEnumIdentifier = (klass, constraint, list = false) => {
  const constraintIdent = constraint.type === "p" ? "" : `_${constraint.name}`;
  const baseEnumIdentifier = `FAKE_ENUM_${klass.namespaceName}_${klass.name}${constraintIdent}`;
  if (list) {
    return `${baseEnumIdentifier}_list`;
  } else {
    return baseEnumIdentifier;
  }
}
```

This'll make it less brittle. Please put this helper on Build so external plugins can use it too. I'm kinda tempted to put something random into it to stop people using the true value in their plugins, but I think that ship has already sailed.

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
benjie: Let's make `addType(addType(type))` idempotent but have it throw if you try and add any of the builtin scalars (`String`, `Int`, `Float`, `Boolean`, `ID`) or if you pass a `non-null` or `list` variant.

ben-pr-p: I'm not exactly sure how to check if that exact type exists for the idempotency check (how to decide if this *exact type* is already known).

I'm seeing that each type has some combination of `name`, `description`, `astNode`, resolvers, etc.

Would a deep equality check on the `astNode` be the thing to do?

I'm still worried that we'd potentially be registering different resolvers for the same type sharing the same AST node, nd that could lead to unpredictable behavior.

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
ben-pr-p: Done in https://github.com/graphile/graphile-engine/pull/748/commits/5f1f5693d33fc034123a19691a6bd5318458222c

ben-pr-p:
What type of testing should we have for this? Just snapshot testing of an example?

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
benjie: It turns out `addType` is already idempotent in this way?

https://github.com/graphile/graphile-engine/blob/726c210ae6f559c132f859fed18dc031a2a2cb1c/packages/graphile-build/src/makeNewBuild.js#L307

So this `shouldSkipAddType` shouldn't be necessary?

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
benjie: Excellent thanks!

benjie:
> What type of testing should we have for this? Just snapshot testing of an example?

Yeah some examples added to https://github.com/graphile/graphile-engine/blob/v4/packages/postgraphile-core/__tests__/kitchen-sink-schema.sql and queries/mutations added to https://github.com/graphile/graphile-engine/tree/v4/packages/postgraphile-core/__tests__/fixtures should cover it :+1:

MarkASwanson:
Is this still being worked on?  It would be nice to be able to return an ENUM type from a function.

ben-pr-p:
I don't have time to take it over the finish line with tests, etc. at the moment (and likely for at least another month). @MarkASwanson if you want to clone and take over [this branch](https://github.com/graphile/graphile-engine/pull/748/commits/5f1f5693d33fc034123a19691a6bd5318458222c), that'd be great! The code works, we just need test coverage.

benjie:
@tsnobip has taken over this PR via new PR https://github.com/graphile/graphile-engine/pull/832 which adds tests.

benjie:
This was merged in https://github.com/graphile/graphile-engine/pull/832

PR #749, opened 2021-07-01
CockroachDB compatibility

benjie:
## Description

CockroachDB is not based on PostgreSQL but it does aim to be wire-compatible with PostgreSQL. This PR aims to add support for CockroachDB to PostGraphile.

## Performance impact

This should have negligible impact on PostGraphile running on PostgreSQL. The performance metrics of PostGraphile running against CockroachDB are not yet known.

## Security impact

This should have no impact on PostGraphile running on PostgreSQL. The security impact of PostGraphile running against CockroachDB is not yet known.

## Checklist

- [x] ~~I have added this feature to 'Pending' in the `RELEASE_NOTES.md` file (if one exists).~~
- [x] ~~If this is a breaking change I've explained why.~~


benjie:
For anyone following this issue, I just ran V5 against Cockroach and introspection worked and basic queries and mutations work too. There's noise in the console due to Cockroach not supporting certain PostgreSQL settings or the `LISTEN` command, but that's just noise.

I don't know if we _fully_ support Cockroach, but it's a big step up from what we had in V4. Reach out if this is important to you :)

PR #751, opened 2021-07-21
feat(graphile-utils): for makeChangeNullabilityPlugin, able to define array nullability, and Mutations and Queries args

GemN:
## Description

In the **makeChangeNullabilityPlugin**
Allow to use define nullability for Arrays.
Allow manipulation of Queries and Mutation arguments

## Performance impact

unknown

## Security impact

unknown

## Checklist

### DIFF
--- a/packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
+++ b/packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
@@ -26,7 +26,7 @@ interface ChangeNullabilityRules {
 }
 
 export default function makeChangeNullabilityPlugin(
-    rules: ChangeNullabilityRules
+  rules: ChangeNullabilityRules
 ): Plugin {
   return (builder: SchemaBuilder, _options: Options) => {
     function applyNullability<Type extends GraphQLType | GraphQLInputType>(type: Type, allNullabilities: boolean[]): Type {
--- a/packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
+++ b/packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
@@ -1,30 +1,61 @@
 import type {
-  SchemaBuilder,
   Build,
   Context,
-  Plugin,
   Options,
+  Plugin,
+  SchemaBuilder,
 } from "graphile-build";
-import type { GraphQLInputFieldConfig, GraphQLFieldConfig } from "graphql";
+import {
+  GraphQLFieldConfig,
+  GraphQLInputFieldConfig,
+  GraphQLInputType,
+  GraphQLNonNull,  GraphQLType } from "graphql";
+import { GraphQLList } from "graphql/type/definition";
+
+interface NullabilityOpts {
+  type?: boolean | boolean[]
+  args?: {
+    [argName: string]: boolean | boolean[]
+  }
+}
 
 interface ChangeNullabilityRules {
   [typeName: string]: {
-    [fieldName: string]: boolean;
-  };
+    [fieldName: string]: boolean | boolean[] | NullabilityOpts
+  },
 }
 
 export default function makeChangeNullabilityPlugin(
-  rules: ChangeNullabilityRules
+    rules: ChangeNullabilityRules
 ): Plugin {
   return (builder: SchemaBuilder, _options: Options) => {
+    function applyNullability<Type extends GraphQLType | GraphQLInputType>(type: Type, allNullabilities: boolean[]): Type {
+      const [nullability, ...rest] = allNullabilities;
+      const nullableType = type instanceof GraphQLNonNull ? type.ofType : type;
+      let inner = null;
+      if (nullableType instanceof GraphQLList) {
+        inner = new GraphQLList(applyNullability(nullableType.ofType, rest))
+      } else {
+        if (rest.length > 0) {
+          throw new Error(`Too many booleans passed`)
+        }
+        inner = nullableType
+      }
+      return nullability ? inner : (inner === type ? new GraphQLNonNull(inner) : type) // Optimisation if it's already non-null*/;
+    }
+    function applyTypeNullability<Type extends GraphQLType | GraphQLInputType>(type: Type, shouldBeNullable: boolean | boolean[]): Type {
+      const allNullabilities: boolean[] = shouldBeNullable instanceof Array ? shouldBeNullable : [shouldBeNullable]
+      return allNullabilities.length > 0 ? applyNullability(type, allNullabilities) : type
+    }
     function changeNullability<
-      Field extends GraphQLInputFieldConfig | GraphQLFieldConfig<any, any>
-    >(field: Field, build: Build, context: Context<Field>): typeof field {
+        Field extends GraphQLInputFieldConfig | GraphQLFieldConfig<any, any>
+        >(field: Field, build: Build, context: Context<Field>): typeof field {
       const {
         Self,
         scope: { fieldName },
       } = context;
-      const typeRules = rules[Self.name];
+      const typeName = Self.name
+      const typeRules = rules[typeName];
       if (!typeRules) {
         return field;
       }
@@ -32,18 +63,29 @@ export default function makeChangeNullabilityPlugin(
       if (shouldBeNullable == null) {
         return field;
       }
-      const {
-        graphql: { getNullableType, GraphQLNonNull },
-      } = build;
-      const nullableType = getNullableType(field.type);
-      return {
-        ...field,
-        type: shouldBeNullable
-          ? nullableType
-          : nullableType === field.type
-          ? new GraphQLNonNull(field.type)
-          : field.type, // Optimisation if it's already non-null
-      };
+      const shouldApplyNullabilities: NullabilityOpts = {}
+      if (typeof shouldBeNullable === "object" && !(shouldBeNullable instanceof Array)) {
+        shouldApplyNullabilities.type = shouldBeNullable.type;
+        shouldApplyNullabilities.args = shouldBeNullable.args;
+      } else {
+        shouldApplyNullabilities.type = shouldBeNullable;
+      }
+      const updatedField = { ...field };
+      try {
+        if (shouldApplyNullabilities.type) {
+          updatedField.type = applyTypeNullability(field.type, shouldApplyNullabilities.type);
+        }
+        if (shouldApplyNullabilities.args) {
+          Object.entries(shouldApplyNullabilities.args).forEach(([argName, argShouldBeNullable]) => {
+            if (!("args" in field) || !field.args?.[argName]) return ;
+            const argType = field.args[argName].type;
+            field.args[argName].type = applyTypeNullability(argType, argShouldBeNullable)
+          })
+        }
+      } catch (err) {
+        throw new Error(`makeChangeNullabilityPlugin. For ${typeName} > ${fieldName}: ${err.message}`);
+      }
+      return updatedField;
     }
     builder.hook("GraphQLInputObjectType:fields:field", changeNullability);
     builder.hook("GraphQLObjectType:fields:field", changeNullability);
--- a/packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
+++ b/packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
@@ -80,20 +80,13 @@ export default function makeChangeNullabilityPlugin(
       if (shouldBeNullable == null) {
         return field;
       }
-      const shouldApplyNullabilities: NullabilityOpts = {};
-      if (
-        typeof shouldBeNullable === "object" &&
-        !Array.isArray(shouldBeNullable)
-      ) {
-        shouldApplyNullabilities.type = shouldBeNullable.type;
-        shouldApplyNullabilities.args = shouldBeNullable.args;
-      } else {
-        shouldApplyNullabilities.type = shouldBeNullable;
-      }
-      const updatedField = field;
+      const shouldApplyNullabilities: NullabilityOpts =
+        typeof shouldBeNullable === "boolean" || Array.isArray(shouldBeNullable)
+          ? { type: shouldBeNullable }
+          : shouldBeNullable;
       try {
         if (shouldApplyNullabilities.type) {
-          updatedField.type = applyTypeNullability(
+          field.type = applyTypeNullability(
             build,
             field.type,
             shouldApplyNullabilities.type
@@ -102,16 +95,16 @@ export default function makeChangeNullabilityPlugin(
         if (shouldApplyNullabilities.args) {
           Object.entries(shouldApplyNullabilities.args).forEach(
             ([argName, argShouldBeNullable]) => {
-              if (!("args" in field) || !field["args"]?.[argName]) {
+              if (!field.args || !field.args[argName]) {
                 console.warn(
                   `warning: makeChangeNullabilityPlugin. For ${typeName} > ${fieldName}: can't apply nullability rule for non-existing arg ${argName}`
                 );
                 return;
               }
-              const argType = field["args"][argName].type;
-              field["args"][argName].type = applyTypeNullability(
+              const arg = field.args[argName];
+              arg.type = applyTypeNullability(
                 build,
-                argType,
+                arg.type,
                 argShouldBeNullable
               );
             }
@@ -122,7 +115,7 @@ export default function makeChangeNullabilityPlugin(
           `makeChangeNullabilityPlugin. For ${typeName} > ${fieldName}: ${err.message}`
         );
       }
-      return updatedField;
+      return field;
     }
     builder.hook("GraphQLInputObjectType:fields:field", changeNullability);
     builder.hook("GraphQLObjectType:fields:field", changeNullability);
--- a/packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
+++ b/packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
@@ -5,22 +5,21 @@ import type {
   Plugin,
   SchemaBuilder,
 } from "graphile-build";
-import {
-  GraphQLFieldConfig,
-  GraphQLInputFieldConfig,
-  GraphQLInputType,
-  GraphQLNonNull,
-  GraphQLType,
-} from "graphql";
-import { GraphQLList } from "graphql/type/definition";
+import type { GraphQLFieldConfig, GraphQLInputFieldConfig } from "graphql";
 
-interface NullabilityOpts {
+export interface NullabilityOpts {
+  /**
+   * The nullability specifications for the return type of the field.
+   */
   type?: boolean | boolean[];
+
+  /**
+   * The nullability specifications for the types of the named arguments of the field.
+   */
   args?: {
     [argName: string]: boolean | boolean[];
   };
 }
-
 interface ChangeNullabilityRules {
   [typeName: string]: {
     [fieldName: string]: boolean | boolean[] | NullabilityOpts;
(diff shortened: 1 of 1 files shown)


packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
+  GraphQLType,
+} from "graphql";
+import { GraphQLList } from "graphql/type/definition";
+
+interface NullabilityOpts {
+  type?: boolean | boolean[];
+  args?: {
+    [argName: string]: boolean | boolean[];
+  };
+}
```
benjie:
```suggestion
export interface NullabilityOpts {
  /**
   * The nullability specifications for the return type of the field.
   */
  type?: boolean | boolean[];
  
  /**
   * The nullability specifications for the types of the named arguments of the field.
   */
  args?: {
    [argName: string]: boolean | boolean[];
  };
}
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
-    [fieldName: string]: boolean;
+    [fieldName: string]: boolean | boolean[] | NullabilityOpts;
   };
 }
 
 export default function makeChangeNullabilityPlugin(
   rules: ChangeNullabilityRules
 ): Plugin {
   return (builder: SchemaBuilder, _options: Options) => {
+    function applyNullability<Type extends GraphQLType | GraphQLInputType>(
```
benjie: [`GraphQLType` covers input types too](https://github.com/graphql/graphql-js/blob/6a5f51f3bd07f3914c7f1ec01908e9e325b77d58/src/type/definition.ts#L52-L68)

```suggestion
    function applyNullability<Type extends GraphQLType>(
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
+      type: Type,
+      allNullabilities: boolean[]
+    ): Type {
```
benjie: This is technically not true. E.g. we might feed in `GraphQLList<GraphQLObjectType>` but output `GraphQLList<GraphQLNonNull<GraphQLObjectType>>`.

I think you should remove the generic from this function.

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
+        if (rest.length > 0) {
+          throw new Error(`Too many booleans passed`);
+        }
+        inner = nullableType;
+      }
+      return nullability
+        ? inner
+        : inner === type
+        ? new GraphQLNonNull(inner)
+        : type; // Optimisation if it's already non-null*/;
```
benjie: Let's not be too clever here:

```suggestion
        : new GraphQLNonNull(inner);
```

(I'm concerned that if `nullableType` is a `GraphQLList` then `inner` will differ from `type` and thus `type` will be used, whereas it should use `GraphQLNonNull(type)` in this case.)

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
+    }
+    function applyTypeNullability<Type extends GraphQLType | GraphQLInputType>(
+      type: Type,
+      shouldBeNullable: boolean | boolean[]
+    ): Type {
+      const allNullabilities: boolean[] =
+        shouldBeNullable instanceof Array
```
benjie: Please use `Array.isArray` instead:

```suggestion
        Array.isArray(shouldBeNullable)
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
 } from "graphile-build";
-import type { GraphQLInputFieldConfig, GraphQLFieldConfig } from "graphql";
+import {
+  GraphQLFieldConfig,
+  GraphQLInputFieldConfig,
+  GraphQLInputType,
+  GraphQLNonNull,
+  GraphQLType,
+} from "graphql";
+import { GraphQLList } from "graphql/type/definition";
```
benjie: You must only import *types* here, the underlying `GraphQLList` / `GraphQLNonNull` must come from `build.graphql.GraphQLList`/etc.

Note I've showing `import type` here but that syntax might not be supported in the current version of TypeScript - either upgrade TypeScript or just know to only import types here.

```suggestion
import type {
  GraphQLFieldConfig,
  GraphQLInputFieldConfig,
} from "graphql";
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
+      const shouldApplyNullabilities: NullabilityOpts = {};
+      if (
+        typeof shouldBeNullable === "object" &&
+        !(shouldBeNullable instanceof Array)
+      ) {
+        shouldApplyNullabilities.type = shouldBeNullable.type;
+        shouldApplyNullabilities.args = shouldBeNullable.args;
+      } else {
+        shouldApplyNullabilities.type = shouldBeNullable;
+      }
```
benjie: Do you need to handle boolean here? Also by doing the `Array.isArray` check first the code is easier to follow:

```suggestion
      if (Array.isArray(shouldBeNullable)) {
        shouldApplyNullabilities.type = shouldBeNullable;
      } else if (typeof shouldBeNullable === 'boolean') {
        shouldApplyNullabilities.type = [shouldBeNullable];
      } else {
        shouldApplyNullabilities.type = shouldBeNullable.type;
        shouldApplyNullabilities.args = shouldBeNullable.args;
      }
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
+      const updatedField = { ...field };
```
benjie: No need to clone; mutating it directly is fine (and more performant).
```suggestion
      const updatedField = field;
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
+        if (shouldApplyNullabilities.type) {
+          updatedField.type = applyTypeNullability(
+            field.type,
+            shouldApplyNullabilities.type
+          );
+        }
+        if (shouldApplyNullabilities.args) {
+          Object.entries(shouldApplyNullabilities.args).forEach(
+            ([argName, argShouldBeNullable]) => {
+              if (!("args" in field) || !field.args?.[argName]) return;
```
benjie: This line should probably throw, or at least log a warning.

benjie (changes requested):
Looking good; here's some early feedback :+1:

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
benjie: (And the other functions.)

GemN:
Thanks for the review, gonna do the changes this week when I have time!

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
GemN: ✅

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
GemN: ✅

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
GemN: ✅

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
GemN: ✅

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
GemN: ✅

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
GemN: ✅ (Actually I copy pasted the previous code that already had the optimisation, that's why)

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
GemN: ✅

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
GemN: ✅

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
GemN: ❌ I can't add 
```ts
  } else if (typeof shouldBeNullable === 'boolean') {
    shouldApplyNullabilities.type = [shouldBeNullable];
  }
```
Because the function `applyTypeNullability` is used to apply nullability on typeName.fieldName.args.argName as well.
It the same function used to cover
- typeName.fieldName: boolean | boolean[] (when we're sure it's not nullability opts)
- typeName.fieldName.type: boolean | boolean[]
- typeName.fieldName.args.argName: boolean | boolean[]

benjie: Sorry, I don't see the problem. The arg type could be a list/etc too, so converting a boolean to a boolean array seems fine?

GemN: If I move the condition there, It won't be applied for typeName.fieldName.args, because the condition that's change a boolean to a boolean array is applied when passing through the function applyTypeNullability

GemN: @benjie

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,128 @@
...
+      const shouldApplyNullabilities: NullabilityOpts = {};
+      if (
+        typeof shouldBeNullable === "object" &&
+        !Array.isArray(shouldBeNullable)
+      ) {
+        shouldApplyNullabilities.type = shouldBeNullable.type;
+        shouldApplyNullabilities.args = shouldBeNullable.args;
+      } else {
+        shouldApplyNullabilities.type = shouldBeNullable;
+      }
```
benjie: I prefer to avoid mutation where it can be avoided without performance penalty.
```suggestion
      const shouldApplyNullabilities: NullabilityOpts =
        typeof shouldBeNullable === "boolean" || Array.isArray(shouldBeNullable)
          ? { type: shouldBeNullable }
          : shouldBeNullable;
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,128 @@
...
+      const updatedField = field;
```
benjie: The `updatedField` alias is misleading because we're simply mutating the underlying `field`. Let's make that more explicit.

```suggestion
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,128 @@
...
+      try {
+        if (shouldApplyNullabilities.type) {
+          updatedField.type = applyTypeNullability(
```
benjie:
```suggestion
          field.type = applyTypeNullability(
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,128 @@
...
+              );
+            }
+          );
+        }
+      } catch (err) {
+        throw new Error(
+          `makeChangeNullabilityPlugin. For ${typeName} > ${fieldName}: ${err.message}`
+        );
+      }
+      return updatedField;
```
benjie:
```suggestion
      return field;
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,128 @@
...
+          updatedField.type = applyTypeNullability(
+            build,
+            field.type,
+            shouldApplyNullabilities.type
+          );
+        }
+        if (shouldApplyNullabilities.args) {
+          Object.entries(shouldApplyNullabilities.args).forEach(
+            ([argName, argShouldBeNullable]) => {
+              if (!("args" in field) || !field["args"]?.[argName]) {
```
benjie: I think we deliberately don't support this syntax. The reasoning is lost in the mists of time (something to do with GRAPHILE_TURBO?)

```suggestion
              if (!field.args || !field.args[argName]) {
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,128 @@
...
+                console.warn(
+                  `warning: makeChangeNullabilityPlugin. For ${typeName} > ${fieldName}: can't apply nullability rule for non-existing arg ${argName}`
+                );
+                return;
+              }
+              const argType = field["args"][argName].type;
```
benjie: If the `[""]` is required for TypeScript reasons we should solve that elsewhere.

```suggestion
              const arg = field.args[argName];
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,128 @@
...
+              field["args"][argName].type = applyTypeNullability(
+                build,
+                argType,
+                argShouldBeNullable
+              );
```
benjie: Let's make this more similar to the type handling code above

```suggestion
              arg.type = applyTypeNullability(
                build,
                arg.type,
                argShouldBeNullable
              );
```

benjie (changes requested):
This is looking good! I've simplified things in a few places.

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
+        if (shouldApplyNullabilities.args) {
+          Object.entries(shouldApplyNullabilities.args).forEach(
+            ([argName, argShouldBeNullable]) => {
+              if (!field.args || !field.args[argName]) {
+                console.warn(
+                  `warning: makeChangeNullabilityPlugin. For ${typeName} > ${fieldName}: can't apply nullability rule for non-existing arg ${argName}`
+                );
+                return;
+              }
+              const arg = field.args[argName];
```
benjie:
```suggestion
              const arg = field.args ? field.args[argName] : null;
              if (!arg) {
                console.warn(
                  `Warning: makeChangeNullabilityPlugin cannot apply rule - field ${typeName}.${fieldName} has no argument '${argName}'`
                );
                return;
              }
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
+                build,
+                arg.type,
+                argShouldBeNullable
+              );
+            }
+          );
+        }
+      } catch (err) {
+        throw new Error(
+          `makeChangeNullabilityPlugin. For ${typeName} > ${fieldName}: ${err.message}`
```
benjie:
```suggestion
          `Error occurred whilst processing makeChangeNullabilityPlugin rule for ${typeName}.${fieldName}: ${err.message}`
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
   };
 }
 
 export default function makeChangeNullabilityPlugin(
   rules: ChangeNullabilityRules
 ): Plugin {
   return (builder: SchemaBuilder, _options: Options) => {
+    function applyNullability(
+      build: Build,
+      type: any,
```
benjie:
```suggestion
      type: GraphQLType,
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
+        if (rest.length > 0) {
+          throw new Error(`Too many booleans passed`);
+        }
+        inner = nullableType;
+      }
+      return nullability ? inner : new GraphQLNonNull(inner);
+    }
+    function applyTypeNullability(
+      build: Build,
+      type: any,
```
benjie:
```suggestion
      type: GraphQLType,
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
-  SchemaBuilder,
   Build,
   Context,
-  Plugin,
   Options,
+  Plugin,
+  SchemaBuilder,
 } from "graphile-build";
-import type { GraphQLInputFieldConfig, GraphQLFieldConfig } from "graphql";
+import type { GraphQLFieldConfig, GraphQLInputFieldConfig } from "graphql";
```
benjie:
```suggestion
import type { GraphQLFieldConfig, GraphQLInputFieldConfig, GraphQLType } from "graphql";
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
+      if (nullableType instanceof GraphQLList) {
+        inner = new GraphQLList(
+          applyNullability(build, nullableType.ofType, rest)
+        );
+      } else {
+        if (rest.length > 0) {
+          throw new Error(`Too many booleans passed`);
+        }
+        inner = nullableType;
+      }
```
benjie:
```suggestion
      if (rest.length > 0 && !(nullableType instanceof GraphQLList)) {
        throw new Error(`There were too many booleans for the number of lists in the type`);
      }
      const inner = nullableType instanceof GraphQLList
        ? new GraphQLList(applyNullability(build, nullableType.ofType, rest))
        : nullableType;
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
+      return nullability ? inner : new GraphQLNonNull(inner);
+    }
+    function applyTypeNullability(
+      build: Build,
+      type: any,
+      shouldBeNullable: boolean | boolean[]
+    ): any {
```
benjie:
```suggestion
    ): GraphQLType {
```

packages/graphile-utils/src/makeChangeNullabilityPlugin.ts
```diff
@@ -1,49 +1,121 @@
...
 
 export default function makeChangeNullabilityPlugin(
   rules: ChangeNullabilityRules
 ): Plugin {
   return (builder: SchemaBuilder, _options: Options) => {
+    function applyNullability(
+      build: Build,
+      type: any,
+      allNullabilities: boolean[]
+    ): any {
```
benjie:
```suggestion
    ): GraphQLType {
```

benjie (changes requested):
Some more refactoring

benjie (changes requested):
I expect we're going to need to do some TypeScript assertions (`as GraphQLOutputType` / `as GraphQLInputType`) due to the polymorphism of the code, but this is looking quite good.

The main thing we need now (other than prettier/typescript fixes) is some tests.

benjie:
In V5 I went with a different solution whereby you specify strings like `[[!]!]!` to indicate the nullability. https://postgraphile.org/postgraphile/next/make-change-nullability-plugin

benjie:
To avoid breaking changes during the V4->V5 transition, this plugin would need to be refactored to use the new string-based solution to be merged into V4, and will need tests. Since it's been awaiting tests for a couple years, I'm going to close it; but I appreciate the work that went into it and it definitely inspired the V5 implementation - thanks! :pray:

PR #756, opened 2021-09-10
Flag to stop LDS loop while no timer is active

### LINKED ISSUE #755: Race condition when closing LDS
### Summary

In our integration test suite, we do teardown the LDS Live Source by calling `schema.__pgLdsSource?.close()`. This does however not work consistently, and from time to time (apparently when the server - that also runs the db in a docker container - is under higher load) it fails to stop the LDS loop and the test job (github action) is only stopped by timing out, failing the build step.

In the meantime, it does keep logging
> Error during LDS loop: Pool has been closed

I believe I tracked this down to a bug in https://github.com/graphile/graphile-engine/blob/6692c2a65f406c55c3d1edfff51c8f7cc2296956/packages/lds/src/index.ts#L96-L101
```ts
  async function loop() {
    try {
      const rows = await client.getChanges(null, 500);
      …
    } catch (e) {
      console.error("Error during LDS loop:", e.message);
      // Recovery time...
      loopTimeout = setTimeout(loop, sleepDuration * 10);
      return;
    }
    loopTimeout = setTimeout(loop, sleepDuration);
  }
  loop();
```

While the `client.getChanges` query is `await`ed, no timeout is scheduled and the `clearTimeout` will be ineffective, so the loop will start over even after `.close()` has been called.

### Steps to reproduce

By setting the `sleepDuration` to a minimum value, the loop spends much more time in the query so a `close()` will hit the sleep much less likely.

### Possible Solution

I can see a few approaches:
* just set a `closed` flag and check this before calling `setTimeout` again
* `.unref()` each timer? I think this might have worked in my case, but probably not a generic solution
* check whether the `client.pool` is closed before calling `getChanges()`, and breaking out of the loop
* check whether the `client.pool` is closed in the `catch` handler of the loop, breaking out from there
* a specific "client closed" exception that is handled differently in the `catch` of the loop, breaking out from there

ab-pm:
Fixes #755.

I hope using the duration variable for the flag is not "too clever", but I wanted to get around writing `if (!closed)` twice.

## Checklist

- [ ] My code matches the project's code style and `yarn lint:fix` passes.

### DIFF
--- a/packages/lds/src/index.ts
+++ b/packages/lds/src/index.ts
@@ -92,16 +92,19 @@ export default async function subscribeToLogicalDecoding(
   }
 
   let loopTimeout: NodeJS.Timer;
+  let nextSleepDuration = 0;
 
   const ldSubscription = {
     close: async () => {
       clearTimeout(loopTimeout);
+      nextSleepDuration = 0;
       await client.close();
     },
   };
 
   let nextStaleCheck = Date.now() + DROP_STALE_SLOTS_INTERVAL;
   async function loop() {
+    nextSleepDuration = sleepDuration;
     try {
       const rows = await client.getChanges(null, 500);
       if (rows.length) {
@@ -161,10 +164,11 @@ export default async function subscribeToLogicalDecoding(
     } catch (e) {
       console.error("Error during LDS loop:", e.message);
       // Recovery time...
-      loopTimeout = setTimeout(loop, sleepDuration * 10);
-      return;
+      nextSleepDuration *= 10;
     }
-    loopTimeout = setTimeout(loop, sleepDuration);
+    if (nextSleepDuration) { // else loop has been stopped
+      loopTimeout = setTimeout(loop, nextSleepDuration);
+    }
   }
   loop();
   return ldSubscription;
--- a/packages/lds/src/index.ts
+++ b/packages/lds/src/index.ts
@@ -92,18 +92,20 @@ export default async function subscribeToLogicalDecoding(
   }
 
   let loopTimeout: NodeJS.Timer;
-  let nextSleepDuration = 0;
+  /** A value of `-1` means to stop looping. */
+  let nextSleepDuration = -1;
 
   const ldSubscription = {
     close: async () => {
       clearTimeout(loopTimeout);
-      nextSleepDuration = 0;
+      nextSleepDuration = -1;
       await client.close();
     },
   };
 
   let nextStaleCheck = Date.now() + DROP_STALE_SLOTS_INTERVAL;
   async function loop() {
+    // Reset the sleep duration (useful after an error where sleep is 10x larger)
     nextSleepDuration = sleepDuration;
     try {
       const rows = await client.getChanges(null, 500);
@@ -164,10 +166,15 @@ export default async function subscribeToLogicalDecoding(
     } catch (e) {
       console.error("Error during LDS loop:", e.message);
       // Recovery time...
-      nextSleepDuration *= 10;
+      if (nextSleepDuration >= 0) {
+        nextSleepDuration = sleepDuration * 10;
+      }
     }
-    if (nextSleepDuration) { // else loop has been stopped
+    if (nextSleepDuration >= 0) {
       loopTimeout = setTimeout(loop, nextSleepDuration);
+    } else {
+      // loop() was stopped; abort
+      return;
     }
   }
   loop();
--- a/packages/lds/src/index.ts
+++ b/packages/lds/src/index.ts
@@ -92,23 +92,26 @@ export default async function subscribeToLogicalDecoding(
   }
 
   let loopTimeout: NodeJS.Timer;
-  /** A value of `-1` means to stop looping. */
-  let nextSleepDuration = -1;
+  /** Set true to halt the loop. */
+  let isClosing = false;
 
   const ldSubscription = {
     close: async () => {
       clearTimeout(loopTimeout);
-      nextSleepDuration = -1;
+      // If the `loop` function is currently in progress, tell it not to queue again.
+      isClosing = true;
       await client.close();
     },
   };
 
   let nextStaleCheck = Date.now() + DROP_STALE_SLOTS_INTERVAL;
   async function loop() {
-    // Reset the sleep duration (useful after an error where sleep is 10x larger)
-    nextSleepDuration = sleepDuration;
     try {
       const rows = await client.getChanges(null, 500);
+      if (isClosing) {
+        // Skip processing this data and exit.
+        return;
+      }
       if (rows.length) {
         for (const row of rows) {
           const {
@@ -166,15 +169,13 @@ export default async function subscribeToLogicalDecoding(
     } catch (e) {
       console.error("Error during LDS loop:", e.message);
       // Recovery time...
-      if (nextSleepDuration >= 0) {
-        nextSleepDuration = sleepDuration * 10;
+      if (!isClosing) {
+        loopTimeout = setTimeout(loop, sleepDuration * 10);
       }
-    }
-    if (nextSleepDuration >= 0) {
-      loopTimeout = setTimeout(loop, nextSleepDuration);
-    } else {
-      // loop() was stopped; abort
       return;
+    }
+    if (!isClosing) {
+      loopTimeout = setTimeout(loop, sleepDuration);
     }
   }
   loop();


packages/lds/src/index.ts
```diff
@@ -92,16 +92,19 @@ export default async function subscribeToLogicalDecoding(
   }
 
   let loopTimeout: NodeJS.Timer;
+  let nextSleepDuration = 0;
```
benjie:
```suggestion
  /** A value of `-1` means to stop looping. */
  let nextSleepDuration = -1;
```

packages/lds/src/index.ts
```diff
@@ -92,16 +92,19 @@ export default async function subscribeToLogicalDecoding(
...
 
   const ldSubscription = {
     close: async () => {
       clearTimeout(loopTimeout);
+      nextSleepDuration = 0;
```
benjie:
```suggestion
      nextSleepDuration = -1;
```

packages/lds/src/index.ts
```diff
@@ -92,16 +92,19 @@ export default async function subscribeToLogicalDecoding(
...
       await client.close();
     },
   };
 
   let nextStaleCheck = Date.now() + DROP_STALE_SLOTS_INTERVAL;
   async function loop() {
+    nextSleepDuration = sleepDuration;
```
benjie:
```suggestion
    // Reset the sleep duration (useful after an error where sleep is 10x larger)
    nextSleepDuration = sleepDuration;
```

packages/lds/src/index.ts
```diff
@@ -161,10 +164,11 @@ export default async function subscribeToLogicalDecoding(
     } catch (e) {
       console.error("Error during LDS loop:", e.message);
       // Recovery time...
-      loopTimeout = setTimeout(loop, sleepDuration * 10);
-      return;
+      nextSleepDuration *= 10;
```
benjie: Make this more explicit to avoid refactoring breaking it later

```suggestion
      if (nextSleepDuration >= 0) {
        nextSleepDuration = sleepDuration * 10;
      }
```

packages/lds/src/index.ts
```diff
@@ -161,10 +164,11 @@ export default async function subscribeToLogicalDecoding(
...
+    }
+    if (nextSleepDuration) { // else loop has been stopped
```
benjie:
```suggestion
    if (nextSleepDuration >= 0) {
```

packages/lds/src/index.ts
```diff
@@ -161,10 +164,11 @@ export default async function subscribeToLogicalDecoding(
...
+      loopTimeout = setTimeout(loop, nextSleepDuration);
     }
```
benjie, line 179:
```suggestion
    } else {
      // loop() was stopped; abort
      return;
    }
```

benjie (changes requested):
Lets use `-1` rather than `0` as the special meaning value for sleep duration but otherwise looks okay from a quick review :+1:

packages/lds/src/index.ts
```diff
@@ -92,16 +92,21 @@ export default async function subscribeToLogicalDecoding(
   }
 
   let loopTimeout: NodeJS.Timer;
+  /** A value of `-1` means to stop looping. */
+  let nextSleepDuration = -1;
 
   const ldSubscription = {
     close: async () => {
       clearTimeout(loopTimeout);
+      nextSleepDuration = -1;
```
benjie:
```suggestion
      // If the `loop` function is currently in progress, tell it not to queue again.
      isClosing = true;
```

packages/lds/src/index.ts
```diff
@@ -92,16 +92,21 @@ export default async function subscribeToLogicalDecoding(
   }
 
   let loopTimeout: NodeJS.Timer;
+  /** A value of `-1` means to stop looping. */
+  let nextSleepDuration = -1;
```
benjie:
```suggestion
  /** Set true to halt the loop. */
  let isClosing = false;
```

packages/lds/src/index.ts
```diff
@@ -92,16 +92,21 @@ export default async function subscribeToLogicalDecoding(
...
       clearTimeout(loopTimeout);
+      nextSleepDuration = -1;
       await client.close();
     },
   };
 
   let nextStaleCheck = Date.now() + DROP_STALE_SLOTS_INTERVAL;
   async function loop() {
+    // Reset the sleep duration (useful after an error where sleep is 10x larger)
+    nextSleepDuration = sleepDuration;
```
benjie:
```suggestion
```

packages/lds/src/index.ts
```diff
@@ -92,16 +92,21 @@ export default async function subscribeToLogicalDecoding(
...
     try {
       const rows = await client.getChanges(null, 500);
```
benjie, line 110:
```suggestion
      const rows = await client.getChanges(null, 500);
      if (isClosing) {
        // Skip processing this data and exit.
        return;
      }
```

packages/lds/src/index.ts
```diff
@@ -161,10 +166,16 @@ export default async function subscribeToLogicalDecoding(
...
       // Recovery time...
-      loopTimeout = setTimeout(loop, sleepDuration * 10);
+      if (nextSleepDuration >= 0) {
+        nextSleepDuration = sleepDuration * 10;
+      }
+    }
+    if (nextSleepDuration >= 0) {
+      loopTimeout = setTimeout(loop, nextSleepDuration);
+    } else {
+      // loop() was stopped; abort
```
benjie:
```suggestion
      if (!isClosing) {
        loopTimeout = setTimeout(loop, sleepDuration * 10);
      }
```

packages/lds/src/index.ts
```diff
@@ -161,10 +166,16 @@ export default async function subscribeToLogicalDecoding(
...
       return;
     }
-    loopTimeout = setTimeout(loop, sleepDuration);
   }
```
benjie, line 180:
```suggestion
    if (!isClosing) {
      loopTimeout = setTimeout(loop, sleepDuration);
    }
  }
```

benjie (changes requested):
I decided overloading `nextSleepDuration` was too clever in the end - namely that you have to track to see if the `* 10` was applied multiple times/etc. Much more straightforward if the mutable value is just a boolean.

I've also added an early exit after receiving but before processing the data.

packages/lds/src/index.ts
```diff
@@ -104,6 +108,10 @@ export default async function subscribeToLogicalDecoding(
   async function loop() {
     try {
       const rows = await client.getChanges(null, 500);
+      if (isClosing) {
+        // Skip processing this data and exit.
+        return;
+      }
```
benjie, line 114: This is the main part I need to think about if it's acceptable or not - it's effectively throwing away LDS data.

ab-pm: I wouldn't throw it away. I don't think there's anything wrong with firing a few other events, even if they don't reach anyone. Better than discarding data that would have been needed.

OTOH, teardown is only available in test environments, so it might not matter at all… Or is `ldSubscription.close()` also used for graceful exit in prod?

benjie: Yeah, I'm not happy with it... but also calling stuff _after_ release resolves definitely feels wrong. I'm thinking we should make the release method await the loop exiting, but that's a bit more code than I can write via the GitHub interface :laughing: Fancy taking it on? I was thinking wrapping `loop` in a promise we can `await` might work. Or maybe using a deferred: https://github.com/graphile/worker/blob/main/src/deferred.ts

ab-pm: Having a `let onClose: (() => void) | null` instead of the `isClosing` flag, with `return new Promise(resolve => { onClose = resolve; })` might work? I'm only using Github here either, that's why I failed the prettier check…

benjie: Yeah one of us should probably check this out and actually write it in an editor. And... you know... see if it works :wink:

benjie: I've just spent some time thinking this over and I feel the above 4 lines should be removed and a mechanism to wait for the current `loop()` call to exit before `ldSubscription.close()` calls `await client.close()` should be added.

benjie:
It seems no-one was sufficiently motivated to check this out locally and do the required editing over the past couple years, so I'm going to close it.

PR #759, opened 2021-10-13
feat(inflection): use builtin for NATURAL and PRIMARY_KEY_ASC/DESC

Eldow:
## Description
I am trying to turn all my enums in snake_case and I couldn't get those three -`NATURAL`, `PRIMARY_KEY_ASC`, `PRIMARY_KEY_DESC`- to follow the global enum inflector. This PR wraps those with `inflection.builtin`, following @benjie advice on discord, allowing to customize them through `makeAddInflectorsPlugin`, eg. :

```js
export const NaturalCasePlugin = makeAddInflectorsPlugin((inflectors) => {
  const { enumName: oldEnumName } = inflectors
  return {
    enumName(value: string) {
      return oldEnumName.call(this, snakeCase(value))
    },
    builtin(value: string) {
      if (value === 'NATURAL') {
        return this.enumName(value)
      }
      return value
    }
  }
}
```

## Performance impact

unknown

## Security impact

unknown

## Checklist

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgConnectionArgOrderByDefaultValue.js
+++ b/packages/graphile-build-pg/src/plugins/PgConnectionArgOrderByDefaultValue.js
@@ -37,7 +37,8 @@ export default (function PgConnectionArgOrderByDefaultValue(builder) {
 
       const primaryKeyAsc = inflection.builtin("PRIMARY_KEY_ASC");
       const defaultValueEnum =
-        TableOrderByType.getValues().find(v => v.name === primaryKeyAsc) || TableOrderByType.getValues()[0];
+        TableOrderByType.getValues().find(v => v.name === primaryKeyAsc) ||
+        TableOrderByType.getValues()[0];
 
       return extend(args, {
         orderBy: extend(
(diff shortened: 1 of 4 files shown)


packages/graphile-build-pg/src/plugins/PgConnectionArgOrderByDefaultValue.js
```diff
@@ -36,8 +36,9 @@ export default (function PgConnectionArgOrderByDefaultValue(builder) {
       }
 
       const defaultValueEnum =
-        TableOrderByType.getValues().find(v => v.name === "PRIMARY_KEY_ASC") ||
-        TableOrderByType.getValues()[0];
+        TableOrderByType.getValues().find(
+          v => v.name === inflection.builtin("PRIMARY_KEY_ASC")
```
benjie: Please extract this function call to just above the `defaultValueEnum` declaration

```js
const primaryKeyAsc = inflection.builtin("PRIMARY_KEY_ASC");
const defautlValueEnum = ...
```

and make this change in the other similar places too :+1:

benjie (changes requested):
Looks good; one minor tweak and then it's good to merge :+1:

packages/graphile-build-pg/src/plugins/PgConnectionArgOrderByDefaultValue.js
Eldow: Done !

benjie (approved):
This looks entirely reasonable :+1: 

Thanks :raised_hands:

Eldow:
Thank you !

PR #770, opened 2021-12-09
polish: improve error message when enum creation fails

tsnobip:
Let me know if you think this could be better rephrased.

## Description

Improve error message when creating an enum fails, to explain that it could be caused by a security policy that hides the rows.

## Performance impact

None, just rephrasing an error message.

## Security impact

None, just rephrasing an error message.

## Checklist

### DIFF
--- a/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
+++ b/packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
@@ -546,7 +546,7 @@ function enumTables(introspectionResults) {
       const data = allData.filter(row => row[col.name] != null);
       if (data.length < 1) {
         throw new Error(
-          `Enum table "${klass.namespaceName}"."${klass.name}" contains no entries for enum constraint '${constraint.name}'.`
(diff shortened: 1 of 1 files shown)


packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -546,7 +546,7 @@ function enumTables(introspectionResults) {
       const data = allData.filter(row => row[col.name] != null);
       if (data.length < 1) {
         throw new Error(
-          `Enum table "${klass.namespaceName}"."${klass.name}" contains no entries for enum constraint '${constraint.name}'.`
+          `Enum table "${klass.namespaceName}"."${klass.name}" contains no visible entries for enum constraint '${constraint.name}'. Check that the table contains rows and that no security policies hide them.`
```
benjie:
```suggestion
          `Enum table "${klass.namespaceName}"."${klass.name}" contains no visible entries for enum constraint '${constraint.name}'. Check that the table contains at least one row and that the rows are not hidden by row-level security policies.`
```

benjie (changes requested):

benjie (approved):
🙌

Issue #742, opened 2021-05-20
pgIntrospection pgFieldIntrospection for GraphQLInputObjectType:fields:field mutation input hook


id3nom:
### Feature description

Add pgIntrospection and pgFieldIntrospection in context.scope for mutation input hook.

### Motivating example

Create a file upload plugin that work with postgreSQL function.

### Breaking changes

None that I know of.

### Supporting development

I [tick all that apply]:

- [ ] am interested in building this feature myself
- [ ] am interested in collaborating on building this feature
- [x] am willing to help testing this feature before it's released
- [ ] am willing to write a test-driven test suite for this feature (before it exists)
- [ ] am a [Graphile sponsor](https://www.graphile.org/sponsor/) ❤️
- [ ] have an active [support or consultancy contract](https://www.graphile.org/support/) with Graphile

benjie:
I think you're asking to add introspection data here:

https://github.com/graphile/graphile-engine/blob/b715f6d1be7dd19e39830b4329fa3c758e5ae91b/packages/graphile-build-pg/src/plugins/makeProcField.js#L635

I'd love a PR for that :+1:

id3nom:
Yes that's what I mean, I successfully have expanded the 'payloadTypeScope' but the 'fieldScope' expansion doesn't work, maybe you can give me a hint on that.

If I find a way to do it I will post a PR with the changes.

benjie:
Probably just adding 

```js
            ...payloadTypeScope
```

after the above line would suffice? If that was raised as a PR I'd also want the variable renamed since it's not just used in the payload any more.

>  the 'fieldScope' expansion doesn't work

I don't really follow. What do you mean by "doesn't work"?

id3nom:
What I mean is that expanding 'payloadTypeScope' give introspection data on the PostgreSQL function but I'm unable to get introspection data on the arguments.

benjie:
Arguments don't have separate introspection; they're part of the function from `pg_proc`: https://github.com/graphile/graphile-engine/blob/726c210ae6f559c132f859fed18dc031a2a2cb1c/packages/graphile-build-pg/src/plugins/introspectionQuery.js#L75-L79

PR #777, opened 2022-01-30
Add Introspection for mutation input hook.

id3nom:
## Description

Add pgIntrospection and pgFieldIntrospection in context.scope for mutation input hook.

https://github.com/graphile/graphile-engine/issues/742

## Performance impact

unknown

## Security impact

unknown

## Checklist

### DIFF
--- a/packages/graphile-build-pg/src/plugins/makeProcField.js
+++ b/packages/graphile-build-pg/src/plugins/makeProcField.js
@@ -629,6 +629,7 @@ export default function makeProcField(
               }
             )}`,
             isMutationInput: true,
+            ...payloadTypeScope,
           }
         );
         args = {


benjie:
What's the motivation for this?

packages/graphile-build-pg/src/plugins/makeProcField.js
```diff
@@ -629,6 +629,7 @@ export default function makeProcField(
               }
             )}`,
             isMutationInput: true,
+            ...payloadTypeScope,
```
benjie, line 632: I do not like reusing the other scope here; please create a new scope for the input type, even if it ends up containing the same things.

benjie (changes requested):

PR #786, opened 2022-04-14
Add support for enum values to @pgQuery

### LINKED ISSUE #1601: pgQuery for `field` returning `enum` fails
### Summary

I _think_ this is a bug. When we use `makeExtendSchema` to extend a type, and attempt to return an enum, we get the error:  `Error: @pgQuery(...) directive called with invalid arguments - for a table value, call it with 'source' for a scalar with 'fragment'!`

I'm having a hard time parsing this message, it's not very clear and could use some improvement. I also can't tell how it relates to my declaration below, or what the resolution is. Have we discovered a new bug?

### Steps to reproduce

We've got a TS enum defined as:

```ts
enum Severity {
  HIGH = 'HIGH',
  LOW = 'LOW',
  MEDIUM = 'MEDIUM',
  NEUTRAL = 'NEUTRAL',
  OK = 'OK'
}
```

And the same declared in GraphQL:

```graphql
    enum Severity {
      OK
      NEUTRAL
      LOW
      MEDIUM
      HIGH
    }
```
 
With a field declared as:

```
      treeShakeableLevel: Severity! @pgQuery(
        fragment: ${embed(
          (queryBuilder: any) =>
            pgSql.fragment`(
              select
              (case when (select
                (
                  coalesce(_extra->'module', _extra->'jsnext:main') is not null
                    or
                  (_extra->'type' is not null and (_extra->'type')::text = 'module')
                )
                from
                  collectors.npm_versions v
                where
                  v.npm_package_id = ${queryBuilder.getTableAlias()}.id
                and
                  v.version = ${queryBuilder.getTableAlias()}.latest_version
                and
                  v.is_deleted = false
              ) = true then ${pgSql.literal(Severity.OK)} else ${pgSql.literal(Severity.NEUTRAL )} end)
            )`
        )}
      )
```

### Expected results

Server starts and creates schema successfully

### Actual results

`Error: @pgQuery(...) directive called with invalid arguments - for a table value, call it with 'source' for a scalar with 'fragment'!`

### Additional context

Of note, if we change `Severity!` to `String!` the server starts without issue.
Please let me know if you need any additional information or schema, resolver code, etc.

### Possible Solution

n/a

angelyan:
## Description

Changes the `isScalarType` check in `makeExtendSchemaPlugin` to `isLeafType` to support enum values with `@pgQuery`.

Fixes https://github.com/graphile/postgraphile/issues/1601.

## Performance impact

unknown

## Security impact

unknown

## Checklist

### DIFF
--- a/packages/graphile-utils/__tests__/ExtendSchemaPlugin-pg.test.js
+++ b/packages/graphile-utils/__tests__/ExtendSchemaPlugin-pg.test.js
@@ -836,3 +836,49 @@ it("allows to retrieve array scalar values", async () => {
     pgClient.release();
   }
 });
+
+it("allows to retrieve enum values", async () => {
+  const schema = await createPostGraphileSchema(pgPool, ["graphile_utils"], {
+    disableDefaultMutations: true,
+    appendPlugins: [
+      makeExtendSchemaPlugin(build => {
+        const { pgSql: sql } = build;
+        return {
+          typeDefs: gql`
+            enum Enum {
+              ENUM
+            }
+            extend type User {
+              myCustomEnum: Enum! @pgQuery(
+                fragment: ${embed(sql.fragment`'ENUM'::text`)}
+              )
+            }
+          `,
+        };
+      }),
+    ],
+  });
+  expect(schema).toMatchSnapshot();
+  const pgClient = await pgPool.connect();
+  try {
+    const { data, errors } = await graphql(
+      schema,
+      `
+        query {
+          user: userById(id: 1) {
+            myCustomEnum
+          }
+        }
+      `,
+      null,
+      { pgClient },
+      {}
+    );
+    expect(errors).toBeFalsy();
+    expect(data).toBeTruthy();
+    expect(data.user).toBeTruthy();
+    expect(data.user.myCustomEnum).toEqual("ENUM");
+  } finally {
+    pgClient.release();
+  }
+});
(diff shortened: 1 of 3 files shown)


benjie:
Thank you for your work on this! Before this can be fully reviewed, it will need tests for enum edge cases. In GraphQL the internal values for enums can be basically anything, and we have various different types of enum in PostGraphile, so please add tests for:

- a PostgreSQL enum value that has to be heavily transformed by the inflection system (e.g. lowercase, contains special symbols, etc)
- an "enum table" enum value that has to be heavily transformed by the inflection system (https://www.graphile.org/postgraphile/enums/#with-enum-tables)
- an enum that uses exotic internal values, for example an `OrderBy` enum value. (This almost certainly won't make sense, but it should fail in a graceful way rather than trying to send a tuple containing SQL fragments/etc into the DB.)

shellscape:
@benjie its pretty cool that @angelyan did the work to get this started, but I'd wager based on the lack of response that they don't have the expertise to satisfy your test requirements. given that you do, please do consider adding the tests you're seeking. (it would be great if this wasn't punted to v5)

benjie:
Reviewing this again, I don't think that this will work for the edge cases that I laid out above. Since V5 is now available to use (and won't suffer from the underlying issue due to a different architecture) I'm going to close it.

PR #791, opened 2022-05-18
fix(pubsub): listen for PG client error to prevent crashes

enisdenjo:
## Description

The pubsub client doesn't have an `error` listener, causing the running process to crash on emitted errors.

## Performance impact

None.

## Security impact

None.

## Checklist

### DIFF
--- a/packages/pg-pubsub/src/index.ts
+++ b/packages/pg-pubsub/src/index.ts
@@ -161,7 +161,7 @@ const plugin: PostGraphilePlugin = {
       let client: pg.PoolClient;
       try {
         client = await pgPool.connect();
-        client.once("error", async e => {
+        client.once("error", e => {
           // eslint-disable-next-line no-console
           console.error("Error occurred in subscriptions client; retrying...");
           // eslint-disable-next-line no-console
--- a/packages/pg-pubsub/src/index.ts
+++ b/packages/pg-pubsub/src/index.ts
@@ -23,6 +23,8 @@ declare module "graphile-build" {
   }
 }
 
+const noop = () => {};
+
 const sleep = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
 
 const debugSubscriptions = createDebugger("postgraphile:subscriptions");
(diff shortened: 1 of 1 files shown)


benjie:
Ah; yes this is required because although `pg` adds an `error` handler to all clients in the pool, it then _removes_ the handler when you check the client out from the pool:

https://github.com/brianc/node-postgres/blob/28ac2a17bce287cfa458153dcabe3ca06ca0e28f/packages/pg-pool/index.js#L288

We probably have this issue in a few more places :sweat_smile:

packages/pg-pubsub/src/index.ts
```diff
@@ -161,6 +161,16 @@ const plugin: PostGraphilePlugin = {
...
+        client.once("error", async e => {
+          // eslint-disable-next-line no-console
+          console.error(`Error occurred in subscriptions client; retrying...`);
+          // eslint-disable-next-line no-console
+          console.error(e);
+          releaseClient(client);
+          if (!pgPool.ending) {
+            setupClient();
+          }
+        });
```
benjie: This needs to be defined as a function and then `client.on('error', fn)`; then we also need a `client.removeListener('error', fn)` when the client is released or when an error occurs. Otherwise if the client is successfully and cleanly returned to the pool, this error handler will stick around, and it will cause issues.

benjie (changes requested):
Thanks for finding this!

packages/pg-pubsub/src/index.ts
enisdenjo: But I use the `once` method - as soon as the event emits, the listener unsubscribes.

enisdenjo:
Converting to a draft, this needs heavier refactoring to be more resilient with error handling.

packages/pg-pubsub/src/index.ts
```diff
@@ -161,6 +161,7 @@ const plugin: PostGraphilePlugin = {
       let client: pg.PoolClient;
       try {
         client = await pgPool.connect();
+        client.on("error", () => {});
```
benjie: We need to remove this listener in `cleanClient`:

https://github.com/graphile/graphile-engine/blob/b7db0ba692ab8885a5356098dc59bb90dd5b9088/packages/pg-pubsub/src/index.ts#L88

Create a `function noop(){}` somewhere so that we can add it by reference:

```suggestion
        client.on("error", noop);
```

benjie (changes requested):

packages/pg-pubsub/src/index.ts
enisdenjo: https://github.com/graphile/graphile-engine/pull/791/commits/f25b47c50ba189eedab0708242d7d2415a45db97

benjie (approved):

PR #792, opened 2022-05-24
Fix: Maximum call stack size exceeded on PG Array parameter

roytan883:
## Description

when PG function arg is Array, has more then 1000 items, will fail on `Maximum call stack size exceeded`.

## Performance impact

null

## Security impact

null


benjie:
Confirmed via experimentation:

```
$ node                                                                                                                                                                                                                                                                                                                                                                                                                                        
Welcome to Node.js v16.13.2.                                                                                                                                                                                                                                                                                                                                                                                                                                                               
Type ".help" for more information.                                                                                                                                                                                                                                                                                                                                                                                                                                                         
> const test = (n) => {const array = new Array(n).fill('a'); [].push(...array); }                                                                                                                                                                                                                                                                                                                                                                                                          
undefined
> test(123169);
undefined
> test(123170);
Uncaught RangeError: Maximum call stack size exceeded
    at Array.push (<anonymous>)
    at test (REPL1:1:63)
```

The actual threshold is dependent on the context in which you call it (because it's a stack size issue).

Please replace the `forEach` calls with traditional `for(...){...}` loops and we can get this merged - thanks :raised_hands: 

(`forEach` has the overhead of calling a function for every loop instance, and when we're talking about exceeding a 100,000 limit that's a lot of calls! `for` is more efficient.)

roytan883:
@benjie add a new commit, use `for of` push item to array

benjie (approved):
Thanks! :rocket:

benjie (changes requested):
Actually, you've raised this PR against `master` which is abandoned. Please rebase on the `v4` branch - thanks :)

PR #793, opened 2022-05-24
fix(pg-sql2): use loop not spread, fixes call stack size exceeded

roytan883:
## Description

when PG function arg is Array, has more then 1000 items, will fail on Maximum call stack size exceeded.

### DIFF
--- a/packages/pg-sql2/src/index.ts
+++ b/packages/pg-sql2/src/index.ts
@@ -231,7 +231,9 @@ export function query(
       const val = values[i];
       if (Array.isArray(val)) {
         const nodes: SQLQuery = val.map(enforceValidNode);
-        items.push(...nodes);
+        for (let item of nodes) {
+          items.push(item);
+        }
       } else {
         const node: SQLNode = enforceValidNode(val);
         items.push(node);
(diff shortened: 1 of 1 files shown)


roytan883:
v4 branche

benjie (approved):

packages/pg-sql2/src/index.ts
```diff
@@ -231,7 +231,9 @@ export function query(
       const val = values[i];
       if (Array.isArray(val)) {
         const nodes: SQLQuery = val.map(enforceValidNode);
-        items.push(...nodes);
+        for (let item of nodes) {
```
benjie:
```suggestion
        for (const item of nodes) {
```

packages/pg-sql2/src/index.ts
```diff
@@ -313,7 +315,10 @@ export function join(items: Array<SQL>, rawSeparator = ""): SQLQuery {
     if (i === 0 || !separator) {
       currentItems.push(...itemsToAppend);
     } else {
-      currentItems.push(sepNode, ...itemsToAppend);
+      currentItems.push(sepNode);
+      for (let item of itemsToAppend) {
```
benjie:
```suggestion
      for (const item of itemsToAppend) {
```

benjie (changes requested):

benjie (approved):

PR #794, opened 2022-05-24
Backport V5 test system to V4

benjie:
## Description

V5's test system is much more pleasant than V4's; but I'd like to bring the V4 tests across to V5 because they test a _lot_ of edge cases. Thus, I'm backporting the testing system from V5 to V4 so that we can later copy the documents and results back to V5 and check for any regressions.

## Performance impact

None, only affects tests.

## Security impact

None, only affects tests.

## Checklist


benjie:
CI is failing on PG9.4 because it does not support the workaround for too many function arguments. I think I previously had a flag that prevented that test running on old Postgres, will need to add that to the config... maybe something like `requiresPg: 100000` (for 10.0.0)?

jemgillam:
> `requiresPg: 100000` (for 10.0.0)?

Works like a charm 👍🏻

PR #801, opened 2022-07-11
feat(pg): Adds support for using partitioned table parents

msjonker:
## Description

This adds the ability to toggle which tables are exposed for partitioned tables.
By default, partitions are exposed and parents are hidden, as they were previously, to prevent a breaking change.
To expose parents and hide the partitions, opt in with:
```js
const options = {
  graphileBuildOptions: {
    pgUsePartitionParents: true,
  },
};
```
See https://github.com/graphile/postgraphile/issues/1160

I'm not sure the process for submitting a PR to the postgraphile repo to add the option exposed in this PR, so that PR has not yet been created.

## Performance impact

The is ultimately just a change to the introspection query, so I don't see there being an impact on performance.

## Security impact

Unknown

## Checklist


packages/graphile-build-pg/src/plugins/introspectionQuery.js
```diff
@@ -187,7 +192,11 @@ with
       rel.relpersistence in ('p') and
       -- We don't want classes that will clash with GraphQL (treat them as private)
       rel.relname not like E'\\\\_\\\\_%' and
-      rel.relkind in ('r', 'v', 'm', 'c', 'f') and
+      ${
+        pgUsePartitionParents
+          ? "rel.relkind in ('r', 'v', 'm', 'c', 'f', 'p') and not rel.relispartition"
```
benjie, line 197: Since this is opt in, we don't need to concern ourselves with whether it supports PG9.6 (which it won't). Nice :+1:

benjie (commented):
From a quick scan this looks good - thanks! (Will review more thoroughly soon.)

benjie:
(Sorry; this is on my radar, but I'm completely ran off my feet until September.)

blahbleepew:
this feature would be very useful, I hope it gets looked at soon!

robertcnix:
👋 Any chance this can be revived?

packages/postgraphile-core/__tests__/partitioned-schema.sql
```diff
@@ -0,0 +1,11 @@
...
+
+create schema partitioned;
+
+create table partitioned.temperature_log (
+    temperature int,
+    log_date date not null
+) partition by range (log_date);
+
+create table partitioned.temperature_log_y2022m07 partition OF partitioned.temperature_log
+    FOR VALUES FROM ('2022-07-01') TO ('2022-07-31');
```
benjie: Please add at least one more partition.

benjie (changes requested):
Excellent; really simple addition - nice and clean. Please add another partition or two to ensure no conflicts occur (they won't with the code you have, but a future version might work in a different way) and then write a query and mutation test against it and then we're good to merge! 🙌 

To write a query test:

1. Create a file like this:
https://github.com/graphile/graphile-engine/blob/29fa78a91df5d9865420fee429918cf3f7010c1e/packages/postgraphile-core/__tests__/queries/base/geometry.queries.test.graphql#L4
2. Edit the metadata at the top to pass the relevant options, and edit the query
3. Run the tests - snapshots should be generated automatically.

Use the prefix you created to prevent the tests running pre PG11. If you need any advice, hit up the Discord https://discord.gg/graphile

And sorry this fell off my radar, it got buried under a large number of notifications and the significant effort I'm putting into PostGraphile V5.

msjonker:
Sounds good! I'll try to get to this in the next couple of weeks.

robertcnix:
FYI, we've run into a small problem with this patch. It seems like the introspection query is not able to determine which table to use as foreign key references.

That is, if there are 2 partitioned tables (t1 and t2) with 2 partitions each, and there is a foreign key between them - there will be 5 records for the foreign keys in the pg_constraint table:

- t1 -> t2
- t1 -> t2_1
- t1 -> t2_2
- t1_1 -> t2
- t1_2 -> t2

The query needs to return just the `t1 -> t2` constraint.

benjie:
Please ensure that the query test you write encompasses:
1. querying a partitioned table directly
2. querying from a partitioned table to a non-partitioned table
3. querying from a non-partitioned table to a partitioned table
4. querying from one partitioned table to another

msjonker:
> FYI, we've run into a small problem with this patch. It seems like the introspection query is not able to determine which table to use as foreign key references.
> 
> That is, if there are 2 partitioned tables (t1 and t2) with 2 partitions each, and there is a foreign key between them - there will be 5 records for the foreign keys in the pg_constraint table:
> 
> * t1 -> t2
> * t1 -> t2_1
> * t1 -> t2_2
> * t1_1 -> t2
> * t1_2 -> t2
> 
> The query needs to return just the `t1 -> t2` constraint.

I'm not sure I understand how this is occurring. Yes, this is how the raw pg_constraint catalog is, but it would seem that the querying takes care of this. The constraints query inner joins and filters on the results of class query, which either includes children tables or the parent table, but not both.

benjie:
You've missed the schema snapshots; I tried to add them myself but you haven't granted me permission to push to this branch.

```
$ git commit -m"Add missing schema snapshots"
[partitioned-tables 3e91f237] Add missing schema snapshots
 2 files changed, 3903 insertions(+)
 create mode 100644 packages/postgraphile-core/__tests__/integration/schema/partitioned.1.graphql
 create mode 100644 packages/postgraphile-core/__tests__/integration/schema/partitioned.2.graphql
$ git push -u
ERROR: Permission to ECO-Parking-Technologies/graphile-engine.git denied to benjie.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

packages/postgraphile-core/scripts/test
```diff
@@ -47,15 +51,21 @@ echo "Database reset successfully ✅"
 # Now run the tests
 if [ $SERVER_VERSION_NUM -ge 110000 ]; then
   yarn jest --forceExit -i $@
+elif [ $SERVER_VERSION_NUM -lt 100000 ] && [ $SERVER_VERSION_NUM -ge 90500 ]; then
+  yarn jest --forceExit -i -t "^((?!partitioned).)*$" $@
```
benjie: This is saying 9.5 <= PG < 10; so it should not run the `pg11` tests?

```suggestion
  yarn jest --forceExit -i -t "^((?!pg11|partitioned).)*$" $@
```

benjie: I think you meant to enable partition tests on PG10+?

```suggestion
elif [ $SERVER_VERSION_NUM -ge 100000 ]; then
  yarn jest --forceExit -i -t "^((?!pg11).)*$" $@
```

benjie (changes requested):
Looking really good! Couple minor issues including the missing files mentioned above. Very close now!

**Please do not edit history now**

(Note to self: fully reviewed including data comparisons in tests. Also the graphileBuildOptions correctly uses `pg` prefix even though the PostGraphile option does not - good 👍 )

packages/postgraphile-core/scripts/test
benjie: See below comment instead.

msjonker:
Sorry for the late reply; I was out on vacation.

I have checked in the partitioned schema snapshots and committed your change. Thanks!

benjie (approved):
Excellent; thanks!

benjie:
Finally released in v4.14.0 - sorry for the delay, I've been a bit overwhelmed with work as of late :sweat_smile: (Plus I wanted to tie up a number of V4 loose ends before this release.)

PR #805, opened 2022-08-26
Add distinctOn to QueryBuilder

stwiname:
## Description

Adds the ability to provide `DISTINCT ON` to Query Builder.

## Performance impact

unknown

## Security impact

unknown

## Checklist


packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -744,7 +753,14 @@ ${sql.join(
         : this.buildSelectFields();
 
     let fragment = sql.fragment`\
-select ${useAsterisk ? sql.fragment`${this.getTableAlias()}.*` : fields}
+select ${
+      this.compiledData.distinctOn.length &&
+      sql.fragment`distinct on (${sql.join(
+        this.compiledData.distinctOn,
+        ", "
+      )})`
```
benjie: Please use ternary rather than passing `0` into the `sql` statement. We do happen to handle that currently because we have a truthy test, but that will go away in the next major version - it was an accident and it allows mistakes to pass through unchallenged.

```suggestion
      this.compiledData.distinctOn.length
      ? sql.fragment`distinct on (${sql.join(
          this.compiledData.distinctOn,
          ", "
        )})`
      : sql.blank
```

benjie (changes requested):
Please can you add a test that uses this functionality?

~~It's probably easiest to do so using our new testing framework (the one on `v4` branch right now is a nightmare); if you branch off of the last good commit in #794 (currently `18b184d38d88f2b9cd7dbeb8bc47c358ec8cfee3`) and~~ 

Add a test similar to the "named_query_builder" test:

https://github.com/graphile/graphile-engine/blob/18b184d38d88f2b9cd7dbeb8bc47c358ec8cfee3/packages/postgraphile-core/__tests__/queries/base/named_query_builder.toys.test.graphql

The line in the config

```
#> ToyCategoriesPlugin: true
```

is used here:

https://github.com/graphile/graphile-engine/blob/18b184d38d88f2b9cd7dbeb8bc47c358ec8cfee3/packages/postgraphile-core/__tests__/helpers-v5.js#L470

to add the plugin that uses it. That's where you'd add a plugin that uses the `distinctOn`. Then it's just a case of updating the assertions at the top of the test file and running the tests, doing so should auto-generate the SQL and JSON5 snapshots (if it doesn't, run it again with `UPDATE_SNAPSHOTS=1` environmental variable set).

benjie:
No need to base off that specific commit any more, I've merged #794

stwiname:
I've found a bug with order by I will improve tests further and fix.

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -192,6 +196,7 @@ class QueryBuilder {
       offset: null,
       first: null,
       last: null,
+      distinctOn: [],
```
benjie, line 198: Please also add to the `beforeLock` section

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -868,11 +886,16 @@ order by (row_number() over (partition by 1)) desc`; /* We don't need to factor
...
     } else if (type === "orderBy") {
-      this.compiledData[type] = this.data[type].map(([a, b, c]) => [
-        callIfNecessary(a, context),
-        b,
-        c,
-      ]);
+      // Add distinct to existing order by to avoid `SELECT DISTINCT ON expressions must match initial ORDER BY expressions`
+      // Convert distinct fields to order by applying ascending and using existing order by for matching fields
+      const distinctAsOrderBy = this.data["distinctOn"].map(
+        d => this.data[type].find(([a]) => isEqual(a, d)) ?? [d, true, null]
```
benjie: Avoid deep comparison, it has performance overhead. Here's a function you can use instead:

```ts
export function arraysMatch<T>(
  array1: ReadonlyArray<T>,
  array2: ReadonlyArray<T>,
  comparator: (val1: T, val2: T) => boolean = (v1, v2) => v1 === v2,
): boolean {
  const l = array1.length;
  if (l !== array2.length) {
    return false;
  }
  for (let i = 0; i < l; i++) {
    if (!comparator(array1[i], array2[i])) {
      return false;
    }
  }
  return true;
}
```

Use it something like this:

```suggestion
        d => this.data[type].find(([a]) => a === d || (Array.isArray(a) && Array.isArray(d) && arraysMatch(a, d)) ?? [d, true, null]
```

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -868,11 +886,16 @@ order by (row_number() over (partition by 1)) desc`; /* We don't need to factor
...
+      );
+
+      // Create a union of the distinct and order by fields
+      this.compiledData[type] = [
+        ...new Set([...distinctAsOrderBy, ...this.data[type]]),
+      ].map(([a, b, c]) => [callIfNecessary(a, context), b, c]);
```
benjie: I don't want to incur the performance costs of the `distinctOn` logic when distinctOn is not used (which will be in the _vast_ majority of schemas). Therefore, please keep the original code, and wrap it with `if (this.data.distinctOn.length === 0) { /* old code */ } else { /* new code */ }`

packages/graphile-build-pg/src/QueryBuilder.js
```diff
@@ -925,6 +950,7 @@ order by (row_number() over (partition by 1)) desc`; /* We don't need to factor
     this.lock("limit");
     this.lock("first");
     this.lock("last");
+    this.lock("distinctOn");
```
benjie, line 989: I'm concerned that because `distinctOn` changes the order, it may break connections (specifically cursors) in some cases. This is one of the reasons that we lock orderBy when we do. E.g. if you order by `[USERNAME_ASC, NAME_DESC, ID_ASC]` and then `distinctOn: [NAME]` then the order would be changed to `NAME_DESC, USERNAME_ASC, ID_ASC` and pagination would work differently.

If you don't want to dig into this, maybe the best bet is to forbid combining distinctOn with connections - e.g. before setting distinctOn you lock cursorComparator/selectCursor/etc and then you enforce that they aren't used (otherwise you throw).

packages/postgraphile-core/__tests__/integration/DistinctOnPlugin.js
```diff
@@ -0,0 +1,45 @@
...
+        return args;
+      }
+
+      addArgDataGenerator(({ distinct }) => {
+        return {
+          pgQuery: queryBuilder => {
+            distinct?.map(field => {
+              const id = sql.fragment`${queryBuilder.getTableAlias()}.${sql.identifier(
+                field
+              )}`;
```
benjie, line 29: Just a warning in case people think copying things out of the tests is a good idea.

```suggestion
              // THIS IS AN EXAMPLE FOR THE TESTS. DO NOT USE THIS IN REAL PRODUCTION CODE!
              // Instead you should use an enum to indicate the allowed identifiers; otherwise
              // you risk interacting with the system columns and maybe worse.
              const id = sql.fragment`${queryBuilder.getTableAlias()}.${sql.identifier(
                field
              )}`;
```

packages/postgraphile-core/__tests__/queries/base/distinct_query_builder.toys.test.graphql
```diff
@@ -0,0 +1,22 @@
+## expect(errors).toBeFalsy();
+#> schema: ["distinct_query_builder"]
+#> subscriptions: true
+#> DistinctOnPlugin: true
+{
+  t1: allToys(distinct: ["color"]) {
+    nodes {
+      name
+      color
```
benjie, line 10:
```suggestion
      id
      name
      color
```

packages/postgraphile-core/__tests__/queries/base/distinct_query_builder.toys.test.graphql
```diff
@@ -0,0 +1,22 @@
...
+    }
+  }
+  t2: allToys(distinct: ["id", "name"]) {
+    nodes {
+      name
```
benjie, line 16:
```suggestion
      id
      name
      color
```

packages/postgraphile-core/__tests__/queries/base/distinct_query_builder.toys.test.graphql
```diff
@@ -0,0 +1,22 @@
...
+    }
+  }
+  t3: allToys {
+    nodes {
+      name
```
benjie, line 62:
```suggestion
      id
      name
      color
```

benjie (changes requested):
Excellent work on filling this out!

My main concerns right now are:

1. connections - are they affected by distinctOn? Does changing the order break them? Will cursors work in all cases? Does totalCount still work as expected?
2. relations - if we select a relation inside the selection set, will Postgres execute this efficiently for only the first match in the distinct on, or will it actually execute it for every match and then throw results away (which could be a significant performance cost)? Even if this is a cost, that can just be a cost related to distinctOn and we can document it, but it's important to understand it.

packages/graphile-build-pg/src/QueryBuilder.js
stwiname: I have added the appropriate locks as I don't have the time to dig into it. I hope I got the locks correct.

From the little research I did cursors should work correctly with `DISTINCT ON`

stwiname:
> Excellent work on filling this out!
> 
> My main concerns right now are:
> 
>     1. connections - are they affected by distinctOn? Does changing the order break them? Will cursors work in all cases? Does totalCount still work as expected?
> 
>     2. relations - if we select a relation inside the selection set, will Postgres execute this efficiently for only the first match in the distinct on, or will it actually execute it for every match and then throw results away (which could be a significant performance cost)? Even if this is a cost, that can just be a cost related to distinctOn and we can document it, but it's important to understand it.

I'm struggling to answer these.  But I have added a test that includes `totalCount` and the result doesn't take into consideration of distinct.

benjie:
Would using distinctOn only in lists satisfy your needs? Then you don’t need to worry about aggregates, cursors, etc. (See docs on —simple-collections.)

stwiname:
> Would using distinctOn only in lists satisfy your needs? Then you don’t need to worry about aggregates, cursors, etc. (See docs on —simple-collections.)

We need both cursors and aggregates.

benjie:
I spent some time tidying this up, but unfortunately it breaks very easily. I was expecting it to break when you combine `last` with `distinctOn` because that reverses the ordering; but the moment you start trying to do cursor pagination it fails because cursor pagination uses conditions and those conditions change what the result of each group is. For example:

| foo | bar |
|---|---|
| 1 | a |
| 1 | b |
| 1 | c |
| 2 | a |
| 2 | b |
| 2 | c |

If you do `distinctOn:"foo"` of this, then you'd expect results like:

```json5
[
  {foo: 1, bar: a}, // cursor = C1
  {foo: 2, bar: a}  // cursor = C2
]
```

But as soon as you do cursor pagination, this falls down. If you ask for the first record after cursor C1 then you don't get C2, instead what you actually get is the result `{foo: 1, bar: b}` because cursor pagination effectively says `foo > 1 OR (foo = 1 AND bar > 'a')`.

To make `distinctOn` work with cursor pagination would involve significantly more work; and since that's a requirement of this ("We need both cursors and aggregates.") I'm going to close this PR for now because it does not address the need safely. I think the best bet to add distinctOn safely would be to do it in a subquery and then select the results from that, but that would be somewhat expensive in many cases.

PR #812, opened 2022-10-07
feat(pg-pubsub): add option to immediately get a result on listen

ol-teuto:
## Description

Motivation:

Pretty much the same as https://github.com/graphile/graphile-engine/pull/612

```
subscription MySubscription {
  listen(topic: "user:2", immediate: true) {
    relatedNodeId
    query {
      user(id: "2") {
        email
      }
    }
  }
}
```

This allows to get a live view of the user email with id 2. Without the option to immediately trigger the query, this would depend on something triggering a `NOTIFY` first.

This is a non breaking change that adds a feature, it's off by default.

It seems like the tests only check the generated schema, so I only updated that test and didn't add a new one

Regarding documentation, should I open a PR on https://github.com/graphile/graphile.github.io before this is merged?

## Performance impact

This is optional and disabled by default, thus there is no performance impact

## Security impact

I can't think of a scenario where this impacts security

## Checklist

### DIFF
--- a/packages/pg-pubsub/src/PgGenericSubscriptionPlugin.ts
+++ b/packages/pg-pubsub/src/PgGenericSubscriptionPlugin.ts
@@ -174,11 +174,11 @@ const PgGenericSubscriptionPlugin: Plugin = function (
               topic: {
                 type: new GraphQLNonNull(GraphQLString),
               },
-              immediate: {
+              initialEvent: {
                 defaultValue: false,
-                type: GraphQLBoolean,
+                type: new GraphQLNonNull(GraphQLBoolean),
                 description:
-                  "If true, this subscription will immediately yield data",
+                  "If true, this subscription will trigger an event as soon as it initiates.",
               },
             },
             subscribe: async (_parent, args, _context, _resolveInfo) => {
@@ -226,7 +226,7 @@ const PgGenericSubscriptionPlugin: Plugin = function (
                   }
                 });
               }
-              if (args.immediate) {
+              if (args.initialEvent) {
                 return withInitialValue(null, asyncIterator);
               } else {
                 return asyncIterator;
--- a/packages/pg-pubsub/src/PgGenericSubscriptionPlugin.ts
+++ b/packages/pg-pubsub/src/PgGenericSubscriptionPlugin.ts
@@ -1,6 +1,6 @@
 import debugFactory from "debug";
 import { Plugin } from "graphile-build";
-import { GraphQLFieldConfig } from "graphql";
+import { GraphQLBoolean, GraphQLFieldConfig } from "graphql";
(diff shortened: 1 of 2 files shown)


packages/pg-pubsub/src/PgGenericSubscriptionPlugin.ts
```diff
@@ -142,6 +174,12 @@ const PgGenericSubscriptionPlugin: Plugin = function (
               topic: {
                 type: new GraphQLNonNull(GraphQLString),
               },
+              immediate: {
+                defaultValue: false,
+                type: GraphQLBoolean,
```
benjie: Let's turn `immediate: Boolean = false` to `immediate: Boolean! = false`.

```suggestion
                type: new GraphQLNonNull(GraphQLBoolean),
```

packages/pg-pubsub/src/PgGenericSubscriptionPlugin.ts
```diff
@@ -142,6 +174,12 @@ const PgGenericSubscriptionPlugin: Plugin = function (
...
+                description:
+                  "If true, this subscription will immediately yield data",
```
benjie:
```suggestion
                  "If true, this subscription will trigger an event as soon as it initiates.",
```

packages/pg-pubsub/src/PgGenericSubscriptionPlugin.ts
```diff
@@ -142,6 +174,12 @@ const PgGenericSubscriptionPlugin: Plugin = function (
               topic: {
                 type: new GraphQLNonNull(GraphQLString),
               },
+              immediate: {
```
benjie: The word `immediate` here is a little confusing - without reading the docs (which don't appear in Explorer, for example), it feels like you should set it because you want to immediately listen ("Why wouldn't I want to listen immediately?!").

We need a name for this that more intuitively encompasses what it does. I'm not super happy with `initialEvent` because it feels like rather than being a boolean it should be an input object to specify the initial event (I do not want to do this!). But `includeInitialEvent`/`triggerInitialEvent`/`triggerEventOnStart`/`withStartEvent`/`withInitialEvent`/`eventOnInit`/etc feel too wordy. And `fireImmediately` feels a bit unfriendly.

Having spent a good 5 minutes pacing the living room trying to think of a better name, right now `initialEvent` is my preferred name for this. But I'm sure there's a better one out there.

Naming things :man_facepalming: 

```suggestion
              initialEvent: {
```

benjie (changes requested):
Seems like a reasonable feature to add :+1:

benjie (approved):
Oh, GitHub didn't notify me you'd pushed another commit! Sorry about that :grimacing:

PR #814, opened 2022-10-26
chore: add graphql 16 as peer dependency for graphql-parse-resolve-info

idoshamun:
I propose to add graphql 16 as a supported version to the peer dependency of graphql-parse-resolve-info.
I'm currently not able to use both graphql 16 and graphql-parse-resolve-info without opting for legacy peer dependencies with npm. Once I do that, it works perfectly in runtime.

### DIFF
--- a/packages/graphql-parse-resolve-info/package.json
+++ b/packages/graphql-parse-resolve-info/package.json
@@ -29,7 +29,7 @@
   },
   "homepage": "https://github.com/graphile/graphile-engine/tree/master/packages/graphql-parse-resolve-info",
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


benjie:
Does it definitely work with variable values/etc? Did you use TypeScript when testing?

idoshamun:
I didn't do full-blown testing just my use case. But I do use TS and all functionality works properly

benjie (approved):

idoshamun:
Thanks for merging!

benjie:
🙌

benjie:
I think this went out in 4.13.0; let me know if not.

PR #830, opened 2023-04-19
Fix add extension types

### LINKED ISSUE #368: Hooks filtering out interface child types from `makeExtendSchema`
### Summary

I am trying to add pure-GraphQL types using the `makeExtendSchema` plugin that includes an interface type and a few types that implement it. I can correctly specify these types and the `__resolveType` handler; however, if I only link the interface into the Graph (i.e. as a return result from a query), the child types are filtered from the final schema.

### Steps to reproduce

Using the schema below:

```graphql
interface MyInterface {
  id: String!
}

type MyType implements MyInterface {
  id: String!
}

extend type Query {
  myQuery: MyInterface
}
```

By providing the typedefs above using the `makeExtendSchema` plugin, `MyInterface` will be included in the resulting schema, but `MyType` will not.

If I instead explicitly tie `MyType` into the graph:

```graphql
extend type Query {
  myQuery: MyInterface
  myOtherQuery: MyType
}
```

then `MyType` is correctly added the the final schema.

### Expected results

All interface types and their implementing types are added to the final schema, regardless of if they are explicitly tied to the graph.

### Actual results

Only explicit implementing types are added to the graph.

### Possible Solution

I currently have a hack in place that adds an explicity query type `___hack: MyType` and then I remove this during the `finalize` hook.

While trying to diagnose, I do find that `MyType` is available during the `GraphQLObject` hook, so I know that the `makeExtendSchema` is correctly providing them back, but some other hook is filtering it out by the time `GraphQLObject:Interfaces` is hit.

mgagliardo91:
## Description

Resolves https://github.com/graphile/graphile.github.io/issues/368

This PR updates the `makeExtendSchemaPlugin` to add all new types to the schema during the `GraphQLSchema` hook so that they are available in the final schema, regardless of explicit usage. Up to this point, types were only added to the schema if they were explicitly (direct or recursively) mapped from a Query/Mutation. 

## Performance impact

Minimum - runs during the `GraphQLSchema` hook.

## Security impact

N/A

## Checklist

### DIFF
--- a/packages/graphile-utils/src/makeExtendSchemaPlugin.ts
+++ b/packages/graphile-utils/src/makeExtendSchemaPlugin.ts
@@ -479,6 +479,7 @@ export default function makeExtendSchemaPlugin(
     });
 
     builder.hook("GraphQLSchema", (schema, build, _context) => {
+      const { inflection } = build;
       const {
         [`ExtendSchemaPlugin_${uniqueId}_typeExtensions`]: typeExtensions,
         [`ExtendSchemaPlugin_${uniqueId}_newTypes`]: newTypeDefinitions,
@@ -497,6 +498,11 @@ export default function makeExtendSchemaPlugin(
         types: [
           ...(schema.types || []),
           ...typeExtensions.GraphQLSchema.types,
+          ...[
+            build.getTypeByName(inflection.builtin("Query")),
+            build.getTypeByName(inflection.builtin("Mutation")),
+            build.getTypeByName(inflection.builtin("Subscription")),
+          ].filter(_ => _),
           ...newTypes,
         ],
       };
--- a/packages/graphile-utils/src/makeExtendSchemaPlugin.ts
+++ b/packages/graphile-utils/src/makeExtendSchemaPlugin.ts
@@ -497,12 +497,12 @@ export default function makeExtendSchemaPlugin(
         ],
         types: [
           ...(schema.types || []),
-          ...typeExtensions.GraphQLSchema.types,
           ...[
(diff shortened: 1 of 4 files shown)


packages/graphile-utils/src/makeExtendSchemaPlugin.ts
```diff
@@ -480,14 +481,24 @@ export default function makeExtendSchemaPlugin(
...
         ...schema,
         directives: [
           ...(schema.directives || build.graphql.specifiedDirectives || []),
           ...typeExtensions.GraphQLSchema.directives,
         ],
-        types: [...(schema.types || []), ...typeExtensions.GraphQLSchema.types],
+        types: [
+          ...(schema.types || []),
+          ...typeExtensions.GraphQLSchema.types,
+          ...newTypes,
```
benjie: This is great and very simple, but the problem is that it significantly changes the generated schema order - putting the types added by makeExtendSchemaPlugin above everything else.

I _think_ that `graphql` walks the types in order, so we may be able to solve this fairly simply by just ensuring that the Query, Mutation and Subscription types are processed first. That might look something like this:

```suggestion
          ...([
            build.getTypeByName(inflection.builtin("Query")),
            build.getTypeByName(inflection.builtin("Mutation")),
            build.getTypeByName(inflection.builtin("Subscription")),
          ].filter(_ => _)),
          ...(schema.types || []),
          ...typeExtensions.GraphQLSchema.types,
          ...newTypes,
```

benjie (changes requested):

benjie (approved):

PR #832, opened 2023-05-25
feat(pg): use table enums in functions via domain

tsnobip:
supersedes #748. I just added tests.

## Description

See https://github.com/graphile/postgraphile/issues/1500.

Basically, this allows to return enums in functions:
```sql 
create table stage_options (
  type text primary key
);
comment on table stage_options is E'@enum';
insert into stage_options (type) values ('pending'), ('round 1'), ('round 2'), ('rejected'), ('hired');

-- either follow the convention of [enum_name]_enum_domain
create domain stage_options_enum_domain as text;
-- or use any name  for the domain and add a smart comment
-- create domain stage as text;
-- comment on domain stage is E'@enum stage_options';

--- this function will add a `nextStage` field to applicant with type `StageOptions`
create or replace function applicants_next_stage(
  a applicants
) returns stage_options_enum_domain
as $$
  select (case when a.stage = 'round 2' then 'hired' 
    else 'rejected' end)::stage_options_enum_domain;
$$ language sql stable;
```

For constraint based enums, you would do define it this way:
```sql
create table my_enums (
  transportation text not null constraint transportation_mean unique
);

comment on constraint transportation_mean on my_enums is E'@enum';
insert into my_enums (transportation) values ('CAR'), ('BIKE'), ('SUBWAY');

-- either follow the convention of [enum_table_name]_[constraint_name]_enum_domain
create domain my_enums_transportation_mean_enum_domain as text;
-- or use any name  for the domain and add a smart comment 
-- referencing the enum this way [enum_table_name]_[constraint_name]
create domain transportation as text;
comment on domain transportation is E'@enum my_enums_transportation_mean';
```

## Checklist

- [x] I have detailed the new feature in the relevant documentation. Here: https://github.com/graphile/graphile.github.io/pull/374


packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -209,6 +209,25 @@ export type PgEntity =
...
+  name: string,
+  constraint?: PgConstraint,
+  list = false
+) => {
+  const constraintIdent = constraint
+    ? constraint.type === "p"
+      ? ""
+      : `_${constraint.name}`
+    : "";
+  const baseEnumIdentifier = `FAKE_ENUM_${namespaceName}_${name}${constraintIdent}`;
```
benjie: `constraint` should either be required (in which case you have to pass the primary key constraint if you don't know), or it should just be a string (in which case you can pass "" with a comment indicating it's referencing the primary key)

```suggestion
  identifierExtra = "",
  list = false
) => {
  const baseEnumIdentifier = `FAKE_ENUM_${namespaceName}_${name}${identifierExtra}`;
```

packages/graphile-build-pg/src/plugins/PgIntrospectionPlugin.js
```diff
@@ -1330,6 +1360,7 @@ Original error: ${e.message}
       }
       return build.extend(build, {
         pgIntrospectionResultsByKind: introspectionResultsByKind,
+        pgMakeFakeEnumIdentifier: fakeEnumIdentifier,
```
benjie: I'm okay with this name, but I don't usually use `Make` for scalars. I'd probably have called it `getPgFakeEnumIdentifier` or similar. (No action necessary.)

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
```diff
@@ -994,30 +998,58 @@ end`;
...
-          );
-          if (baseInputType && baseInputType !== baseType) {
+          // might be used as an enum alias: #1500
+          if (
+            type.name.endsWith("_enum_domain") ||
+            (type.tags && type.tags.enum)
+          ) {
+            const underlyingEnumTypeName = type.name.endsWith("_enum_domain")
+              ? type.name.replace("_enum_domain", "")
+              : type.tags.enum;
```
benjie: Tag is more explicit, so it should "win".

`.replace` isn't safe in case `_enum_domain` occurs more than once.

```suggestion
          const tagEnumName =
            type.tags && typeof type.tags.enum === "string" ? type.tags : null;
          // TODO: move this to the top of the file
          const ENUM_DOMAIN_SUFFIX = "_enum_domain";
          const truncatedTypeName =
            !tagEnumName && type.name.endsWith(ENUM_DOMAIN_SUFFIX)
              ? type.name.slice(0, type.name.length - ENUM_DOMAIN_SUFFIX.length)
              : null;
          const underlyingEnumTypeName = tagEnumName || truncatedTypeName;
          if (underlyingEnumTypeName !== null) {
```

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
```diff
@@ -994,30 +998,58 @@ end`;
...
+            if (baseInputType && baseInputType !== baseType) {
+              gqlInputTypeByTypeIdAndModifier[type.id][typeModifierKey] =
+                Object.assign(Object.create(baseInputType), {
+                  name: inflection.inputType(
+                    gqlTypeByTypeIdAndModifier[type.id][typeModifierKey]
+                  ),
+                  description: type.description,
+                });
+            }
           }
```
benjie, line 1070: This PR doesn't have to _implement_ this, but I'd like to know how we'd handle it: if you use constraint-based enums rather than primary key enum tables, how would you reference the constraint in both of these cases (auto-named and via smart tags)

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
```diff
@@ -994,30 +998,58 @@ end`;
...
+
+            const baseType = getGqlTypeByTypeIdAndModifier(
+              baseTypeId,
+              typeModifierKey
+            );
+
+            const baseInputType = getGqlInputTypeByTypeIdAndModifier(
+              baseTypeId,
+              typeModifierKey
+            );
```
benjie, line 1028: If these types aren't both found, we should throw?

packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
```diff
@@ -1238,3 +1239,67 @@ create table geometry.geom (
...
+  id int,
+  first_name text,
+  last_name text,
+  stage text references function_returning_enum.stage_options (type)
+);
+
+--- follow the convention of [enum_name]_enum_domain
+create domain function_returning_enum.stage_options_enum_domain as text;
+
+create or replace function function_returning_enum.applicants_next_stage(
```
benjie: No need for `or replace` here, and it requires the reader to validate the function hasn't already been defined.

```suggestion
create function function_returning_enum.applicants_next_stage(
```

packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
```diff
@@ -1238,3 +1239,67 @@ create table geometry.geom (
...
+    else 'ok' end)::function_returning_enum.length;
+$$ language sql stable;
+
+create or replace function function_returning_enum.applicants_name_length(
+  a function_returning_enum.applicants
+) returns function_returning_enum.length
+as $$
+  select function_returning_enum.text_length(a.last_name, 4)::function_returning_enum.length;
+$$ language sql stable;
+comment on function function_returning_enum.applicants_name_length is E'@filterable';
```
benjie: These are all great, but we also need to see usage where the enum domain is used as an argument to the function (rather than the return type as above).

yarn.lock
```diff
@@ -3380,9 +3380,9 @@ camelcase@^5.0.0, camelcase@^5.3.1:
   integrity sha512-L28STB170nwWS63UjtlEOE3dldQApaJXZkOI1uMFfzf3rRuPegHaHesyee+YxQ+W6SvRDQV6UrdOdRiR153wJg==
 
 caniuse-lite@^1.0.30001265:
-  version "1.0.30001270"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001270.tgz#cc9c37a4ec5c1a8d616fc7bace902bb053b0cdea"
-  integrity sha512-TcIC7AyNWXhcOmv2KftOl1ShFAaHQYcB/EPL/hEyMrcS7ZX0/DvV1aoy6BzV0+16wTpoAyTMGDNAJfSqS/rz7A==
+  version "1.0.30001489"
+  resolved "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001489.tgz"
+  integrity sha512-x1mgZEXK8jHIfAxm+xgdpHpk50IN3z3q3zP261/WS+uvePxW8izXuCu6AHz0lkuYTlATDehiZ/tNyYBdSQsOUQ==
```
benjie: If this change isn't essential, please revert it.

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
```diff
@@ -753,6 +754,9 @@ export default (function PgTypesPlugin(
               gqlInputType;
           }
         }
+
+        let shouldSkipAddType;
```
benjie:
```suggestion
        let shouldSkipAddType = false;
```

benjie (changes requested):
This is really good! Main feedback:

1. Need tests for function arguments (in addition to the return types already tested)
2. What would happen for _constraint_ enums rather than table enums?
3. Make sure we're handling user error well, with helpful messages. ("X has defined '@enum Y', but no type named 'Y' could be found" or similar.)

benjie:
Also, please reference this thread: https://github.com/graphile/graphile-engine/pull/748#discussion_r650828353 - `shouldSkipAddType` shouldn't be needed?

tsnobip:
Thanks for the review @benjie, I'll address your comments during the week.

yarn.lock
tsnobip: probably not essential, I had warnings about it and since something else was raising errors, I decided to do the upgrade to get rid of the noise of the warning, but I can definitely drop the commit.

tsnobip:
a few comments are still remaining but I can't make them work so far, I'll get back to it next week.

tsnobip:
@benjie I think I addressed all your comments including tests with constraint enums, are we good now?

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
```diff
@@ -994,30 +1001,72 @@ end`;
...
-            type.domainBaseTypeId,
-            typeModifier
-          );
-          // Hack stolen from: https://github.com/graphile/postgraphile/blob/ade728ed8f8e3ecdc5fdad7d770c67aa573578eb/src/graphql/schema/type/aliasGqlType.ts#L16
-          gqlTypeByTypeIdAndModifier[type.id][typeModifierKey] = Object.assign(
-            Object.create(baseType),
-            {
-              name: inflection.domainType(type),
-              description: type.description,
+          // might be used as an enum alias: #1500
```
benjie: Wrong repo for a short link, alas.

```suggestion
          // might be used as an enum alias: https://github.com/graphile/postgraphile/issues/1500
```

packages/graphile-build-pg/src/plugins/PgTypesPlugin.js
```diff
@@ -994,30 +1001,72 @@ end`;
...
+          const truncatedTypeName =
+            !tagEnumName && type.name.endsWith(ENUM_DOMAIN_SUFFIX)
+              ? type.name.slice(0, type.name.length - ENUM_DOMAIN_SUFFIX.length)
+              : null;
+          const underlyingEnumTypeName = tagEnumName || truncatedTypeName;
+          if (underlyingEnumTypeName !== null) {
+            const baseTypeId = getPgFakeEnumIdentifier(
+              type.namespaceName,
+              underlyingEnumTypeName
+            );
```
benjie, line 1018: So this works for constraints by assuming the format of `...${name}${identifierExtra}`? Seems a reasonable approach; I do wonder if this means we should simplify `fakeEnumIdentifier` down to just accepting `name` (which should include `identifierExtra`) to make this more explicit... But meh 🤷‍♂️

packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
```diff
@@ -1238,3 +1239,125 @@ create table geometry.geom (
...
+comment on function function_returning_enum.applicants_pet_food is E'@filterable';
+
+create domain function_returning_enum.transportation as text;
+comment on domain function_returning_enum.transportation is E'@enum enum_table_transportation_enum';
+
+create function function_returning_enum.applicants_by_transportation(
+  transportation function_returning_enum.transportation
+) returns setof function_returning_enum.applicants
+as $$ 
+  select * from function_returning_enum.applicants a where a.transportation = transportation;
```
benjie: Took me a while to figure out why `applicantsByTransportation` returned all three records!

```suggestion
  select * from function_returning_enum.applicants a where a.transportation = applicants_by_transportation.transportation;
```

benjie (changes requested):
Looking good!

packages/postgraphile-core/__tests__/queries/base/function-returning-enum.json5
```diff
@@ -0,0 +1,115 @@
...
+        lastName: "Bowie",
+        stage: "ROUND_2",
+        nextStage: "HIRED",
+        favoritePet: "DOG",
+        petFood: "CAT",
+        nameLength: "OK",
+        transportation: "SUBWAY",
+        favoritePetTransportation: "BIKE",
+      },
+      {
```
benjie:
```suggestion
      {
```

benjie (approved):
If this passes CI, it's good to merge 👍

packages/postgraphile-core/__tests__/kitchen-sink-schema.sql
tsnobip: arf I get bitten by this so often, damn ambiguous names in postgres!

benjie:
(I still don't think the `shouldSkipAddType` is needed: https://github.com/graphile/graphile-engine/pull/748#discussion_r669667007 )

tsnobip:
> (I still don't think the `shouldSkipAddType` is needed: [#748 (comment)](https://github.com/graphile/graphile-engine/pull/748#discussion_r669667007) )

you're the boss! Should we change it?

benjie:
Meh; we can remove it in a follow-up PR if we really want to. Was just trying to keep the code simple.

PR #856, opened 2024-03-13
chore: add node specifier to assert import

Hebilicious:
## Description

Hello there,
In order to use this library in some runtimes (ie cloudflare workers), I have to patch this library with an import specifier.  I believe this change could benefit other users.

Additionally, I haven't looked at the rest of the codebase, so maybe you would want to add node specifiers everywhere in one single PR ?

## Checklist

- [ ] My code matches the project's code style and `yarn lint:fix` passes.

### DIFF
--- a/packages/graphql-parse-resolve-info/src/index.ts
+++ b/packages/graphql-parse-resolve-info/src/index.ts
@@ -1,4 +1,4 @@
-import * as assert from "assert";
+import * as assert from "node:assert";
 import {
   getNamedType,
   isCompositeType,


benjie:
If you’re doing it, it needs doing via a lint rule to enforce it in future imports too. However I have a vague recollection that adding a specifier like this actually breaks it on some runtimes (bun?), have a look through the history to see if specifiers were removed at one point.

mderrick:
I tried patching this myself for Cloudflare workers but am met with `Uncaught Error: Dynamic require of "node:assert" is not supported` and had to remove `assert` entirely and replace it with a manual throw. It would be great if this package could support CF workers.

PR #858, opened 2024-03-25
feat(graphql-parse-resolve-info): add isResolveTree type guard

### LINKED ISSUE #849: Provide a type guard to distinguish between ResolveTree and FieldsByTypeName
### Feature description

`parseResolveInfo` return type is `ResolveTree | FieldsByTypeName | null | undefined`. You can get rid of `null | undefined` with an `if` condition but you're left with `ResolveTree | FieldsByTypeName`. The problem is that `simplifyParsedResolveInfoFragmentWithType` expects a `ResolveTree` as its first argument. We need a way to tell them apart. Right now I'm using this:

```ts
function isResolveTree(
  value: ResolveTree | FieldsByTypeName,
): value is ResolveTree {
  return typeof value.name === "string" && Boolean(value.fieldsByTypeName);
}
```

### Motivating example

It's hard to use `simplifyParsedResolveInfoFragmentWithType` when using TypeScript without this type guard, unless I'm doing something wrong here.

### Supporting development

I [tick all that apply]:

- [x] am interested in building this feature myself
- [x] am interested in collaborating on building this feature
- [x] am willing to help testing this feature before it's released
- [x] am willing to write a test-driven test suite for this feature (before it exists)
- [ ] am a [Graphile sponsor](https://www.graphile.org/sponsor/) ❤️
- [ ] have an active [support or consultancy contract](https://www.graphile.org/support/) with Graphile

Nargonath:
Fix #849

## Description

This PR adds a TypeScript type guard to determine whether a value is a `ResolveTree` or not.

## Performance impact

None (or minimal), it's just a two property read and boolean conditions.

## Security impact

none

## Checklist

### DIFF
--- a/packages/graphql-parse-resolve-info/README.md
+++ b/packages/graphql-parse-resolve-info/README.md
@@ -372,6 +372,26 @@ single-level object containing only the fields compatible with
            ...as before...
 ```
 
+### `isResolveTree(value)`
+
+Determines whether the argument is a value of type `ResolveTree`. This is useful only to TypeScript users. It allows to differentiate between `ResolveTree` and `FieldsByTypeName` objects to pass to `simplifyParsedResolveInfoFragmentWithType `.
+
+Example:
+
+```ts
+import { isResolveTree } from "graphql-parse-resolve-info";
+
+const { parsedResolveInfoFragment, simplifiedFragment } = await graphql(
+  Schema,
+  query
+);
+
+isResolveTree(parsedResolveInfoFragment); // returns true
+isResolveTree(simplifiedFragment); // returns true
+isResolveTree(parsedResolveInfoFragment.fieldsByTypeName); // returns false
+isResolveTree(simplifiedFragment.fieldsByTypeName); // returns false
+```
+
 ## Thanks
 
 This project was originally based on https://github.com/tjmehta/graphql-parse-fields, but has evolved a lot since then.
--- a/packages/graphql-parse-resolve-info/src/index.ts
+++ b/packages/graphql-parse-resolve-info/src/index.ts
@@ -37,6 +37,12 @@ export interface ResolveTree {
   fieldsByTypeName: FieldsByTypeName;
 }
 
+export function isResolveTree(
+  value: ResolveTree | FieldsByTypeName | null | undefined
+): value is ResolveTree {
+  return typeof value?.name === "string" && Boolean(value.fieldsByTypeName);
+}
+
 const debug = debugFactory("graphql-parse-resolve-info");
 
 const DEBUG_ENABLED = debug.enabled;
--- a/packages/graphql-parse-resolve-info/__tests__/test.test.js
+++ b/packages/graphql-parse-resolve-info/__tests__/test.test.js
@@ -1,6 +1,7 @@
 const {
   parseResolveInfo,
   simplifyParsedResolveInfoFragmentWithType,
+  isResolveTree,
 } = require("../src");
 const {
   graphql,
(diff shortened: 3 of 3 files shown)


packages/graphql-parse-resolve-info/README.md
```diff
@@ -372,6 +372,26 @@ single-level object containing only the fields compatible with
...
+
+Example:
+
+```ts
+import { isResolveTree } from "graphql-parse-resolve-info";
+
+const { parsedResolveInfoFragment, simplifiedFragment } = await graphql(
+  Schema,
+  query
+);
```
benjie, line 387: I think you might have mis-typed this example, we wouldn't get these results from calling `graphql()`?

Maybe something like this would be more helpful?

```suggestion
const myResolver: GraphQLFieldResolver<any, any> = (source, args, context, resolveInfo) => {
  const parsedResolveInfoFragment = parseResolveInfo(resolveInfo);
  if (!isResolveTree(parsedResolveInfoFragment)) {
    throw new Error(`Expected ResolveTree`);
  }
  const simplifiedFragment = simplifyParsedResolveInfoFragmentWithType(
    parsedResolveInfoFragment,
    resolveInfo.returnType
  );
  // ...
};
```

packages/graphql-parse-resolve-info/src/index.ts
```diff
@@ -37,6 +37,12 @@ export interface ResolveTree {
   fieldsByTypeName: FieldsByTypeName;
 }
 
+export function isResolveTree(
+  value: ResolveTree | FieldsByTypeName | null | undefined
+): value is ResolveTree {
+  return typeof value?.name === "string" && Boolean(value.fieldsByTypeName);
```
benjie, line 43: I'm not sure this is safe; depending on how it's used it could be tricked by a query like:

```
{
  field {
    name: id # `ID` type is always a string
    fieldsByTypeName: __typename
  }
}
```

benjie (changes requested):
A safe way to know if the result is a ResolveTree or FieldsByTypeName is to know whether you called the function with `keepRoot` set or not. Maybe it would be better to add a function overload to `parseResolveInfo`, e.g.:

```ts
export function parseResolveInfo(
  resolveInfo: GraphQLResolveInfo,
  options?: Omit<ParseOptions, "keepRoot"> & { keepRoot?: false }
): ResolveTree | null | undefined;
export function parseResolveInfo(
  resolveInfo: GraphQLResolveInfo,
  options: Omit<ParseOptions, "keepRoot"> & { keepRoot: true }
): FieldsByTypeName;
export function parseResolveInfo(
  resolveInfo: GraphQLResolveInfo,
  options: ParseOptions
): ResolveTree | FieldsByTypeName | null | undefined;
export function parseResolveInfo(
  resolveInfo: GraphQLResolveInfo,
  options: ParseOptions = {}
): ResolveTree | FieldsByTypeName | null | undefined {
  // ...
}
```

This way you wouldn't even need the type guard, it would be typed for you automatically? If this seems like a feasible approach, please consider updating the PR to take this approach instead.

Nargonath:
Thanks for the suggestion @benjie. I wasn't aware of the `keepRoot` option. Does it already exist or is it something you suggest to create as part of this PR?

benjie:
It already exists, IIRC it’s what controls the type that is returned. The changes I have suggested are types only.

benjie:
_[semi-automated message]_ Hi, there has been no activity in this issue for a while so I'm closing it to keep the issues/pull requests manageable. If this is still an issue, please re-open with additional details.

PR #872, opened 2025-05-13
Fix return type of simplifyParsedResolveInfoFragmentWithType

dhritzkiv:
## Description

Specify type as `ResolveTree` for returned `fields` in `simplifyParsedResolveInfoFragmentWithType`. Otherwise, the return type is simply `{}` which is not indexable in TS.

## Checklist

### DIFF
--- a/packages/graphql-parse-resolve-info/src/index.ts
+++ b/packages/graphql-parse-resolve-info/src/index.ts
@@ -352,7 +352,7 @@ export function simplifyParsedResolveInfoFragmentWithType(
   type: GraphQLType
 ) {
   const { fieldsByTypeName } = parsedResolveInfoFragment;
-  const fields = {};
+  const fields: ResolveTree = {};
   const strippedType = getNamedType(type);
   if (isCompositeType(strippedType)) {
     Object.assign(fields, fieldsByTypeName[strippedType.name]);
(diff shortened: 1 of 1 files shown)


benjie (dismissed):
Thanks!

packages/graphql-parse-resolve-info/src/index.ts
```diff
@@ -352,7 +352,7 @@ export function simplifyParsedResolveInfoFragmentWithType(
   type: GraphQLType
 ) {
   const { fieldsByTypeName } = parsedResolveInfoFragment;
-  const fields = {};
+  const fields: ResolveTree = {};
```
benjie:
```
src/index.ts:355:9 - error TS2739: Type '{}' is missing the following properties from type 'ResolveTree': name, alias, args, fieldsByTypeName

355   const fields: ResolveTree = {};
```

dhritzkiv:
Sorry about that – I didn't see that tests fail when I initially submitted this PR.

I've now changed the syntax to allow for treating `fields` as `ResolveTree`, when initializing it.

benjie:
`fields` isn't actually a ResolveTree; it's a `Record<string, ResolveTree>`. Whilst reworking this I spotted a bug in the merging, so I've raised an alternative fix in #873.

PR #879, opened 2026-03-10
lds: parse wal2json string values according to their type

ab-pm:
## Description

We're using @graphile/lds in a LiveSource (similar to subscriptions-lds) that is not only used for monitoring live queries, but also publishes the actual row data. We discovered a bug in our application where the announcement `.data` that is passed to the `subscribeToLogicalDecoding` callback did not match the expected value - we assumed it would basically look like the results from `pg.query('SELECT * FROM schema.table')`. However we found that some of the properties contained plain strings!

The culprit appears to be that wal2json outputs most record values as string, basically as `value::text` (I believe?) and not `to_json(value)` as one might expect. The changes emitted by `PgLogicalDecoding` are then simply transformed from key/value lists to a dictionary, but no type-specific parsing is performed (the `columntypes` are available but ignored).

## Performance impact

unknown

## Correctness impact

As far as I can tell, most live queries listen for changes to records by the record identifier (primary key), and those typically consist of integers and texts where this is not an issue. I fail to understand how the `liveCollection` predicates work and what they check for.

In our case, we were looking at the published data and tried to access some properties that should have been `Date`s, `JSON` values or `Array`s (`timestamptz`, `jsonb` and `int[]` respectively). The issue surely occurs with other rarely used datatypes, like enum(array)s, ranges, custom composite types, etc.

## Checklist

- [x] I have not added this feature to `RELEASE_NOTES.md` as I did not find any

## Further steps

This PR does report the issue and serves to start the discussion. If you're interested in merging the proposed fix, I'd like to continue with some refactorings; in particular I think it should be possible to inject custom type parsers. For that I would suggest to make `parse`, `changeToRecord` and `changeToPk` methods of the `PgLogicalDecoding` class, and use `this.types` instead of `pg.types` where `.types` should be a `pg.CustomTypesConfig`-shaped object that can be configured via the constructor `options` and falls back to `pg.types` if not passed.

### DIFF
--- a/.github/workflows/ci/docker-entrypoint-initdb.d/020-wal2json.sh
+++ b/.github/workflows/ci/docker-entrypoint-initdb.d/020-wal2json.sh
@@ -5,7 +5,7 @@ set -e
 # https://github.com/debezium/docker-images/blob/master/postgres/11/Dockerfile
 
 PROTOC_VERSION=1.3
-WAL2JSON_COMMIT_ID=c54d89649c3fe5e0aa79c3a87493935232e962a7
+WAL2JSON_COMMIT_ID=wal2json_2_6
 USE_PGXS=1
 export PGUSER=postgres
 export PGPASSWORD=postgres
@@ -13,13 +13,14 @@ createuser --superuser root
 
 apt-get update
 apt-get install -f -y --no-install-recommends \
-        software-properties-common \
+        apt-transport-https \
+        ca-certificates \
         build-essential \
         pkg-config \
         git \
         postgresql-server-dev-$PG_MAJOR
-add-apt-repository "deb http://ftp.debian.org/debian testing main contrib"
-apt-get update
+# add-apt-repository "deb http://ftp.debian.org/debian testing main contrib"
+# apt-get update
 rm -rf /var/lib/apt/lists/*
 
 cd /
--- a/packages/lds/src/pg-logical-decoding.ts
+++ b/packages/lds/src/pg-logical-decoding.ts
@@ -60,27 +60,7 @@ export interface DeleteChange extends Change {
   oldkeys: Keys;
 }
 
-const parse = (value: any, typeOid: number) => {
-  if (value === null) return null;
-  // wal2json always outputs `bool`s as boolean, not as string, irregardless of `numeric-data-types-as-string`.
-  if (typeOid === pg.types.builtins.BOOL) return value;
-  // FIXME: this should use `client.getTypeParser` or have some other non-global option to configure type parsing
-  const parser = pg.types.getTypeParser(typeOid, "text");
-  return parser(value);
-};
-
-export const changeToRecord = (change: InsertChange | UpdateChange) => {
-  const { columnnames, columnvalues, columntypeoids } = change;
-  return columnnames.reduce<Record<string, any>>((memo, name, i) => {
-    memo[name] = parse(columnvalues[i], columntypeoids[i]);
-    return memo;
-  }, {});
-};
-
-export const changeToPk = (change: UpdateChange | DeleteChange) => {
-  const { keyvalues, keytypeoids } = change.oldkeys;
-  return keyvalues.map((value, i) => parse(value, keytypeoids[i]));
-};
+const id = <T>(value: T): T => value;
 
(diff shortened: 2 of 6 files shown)


benjie:
Let’s make this opt in so people who don’t need it don’t incur the additional cost; add something like `types?: typeof pg.types` to Options.

ab-pm:
Done.

I notice the CI appears to be running some older postgres versions with some older wal2json only; apparently [`numeric-data-types-as-string`](https://github.com/eulerto/wal2json/pull/255) was introduced only in [release 2.6](https://github.com/eulerto/wal2json/releases/tag/wal2json_2_6) 2y ago. I guess this should also become configurable?

packages/lds/src/pg-logical-decoding.ts
```diff
@@ -81,37 +72,67 @@ const toLsnData = ([lsn, data]: [string, string]): Payload => ({
...
     this.temporary = temporary;
+    const parametersSql = Object.entries({
+      "add-tables": tablePattern != "*.*" ? tablePattern : null,
+      "include-types": "f", // type names are unnecessary
+      "include-type-oids": types ? "t" : null,
+      "numeric-data-types-as-string": types ? "t" : null,
+      ...params,
+    })
+      .flatMap(entry => (typeof entry[1] == "string" ? entry : []))
+      .map(pg.Client.prototype.escapeLiteral)
```
ab-pm, line 115: Not pretty. Unfortunately the top-level export `pg.escapeLiteral` is only available since https://github.com/brianc/node-postgres/pull/2954, and I didn't want to require pg@^8.11.0…

.github/workflows/ci/docker-entrypoint-initdb.d/020-wal2json.sh
```diff
@@ -5,21 +5,22 @@ set -e
...
+        apt-transport-https \
+        ca-certificates \
         build-essential \
         pkg-config \
         git \
         postgresql-server-dev-$PG_MAJOR
-add-apt-repository "deb http://ftp.debian.org/debian testing main contrib"
-apt-get update
+# add-apt-repository "deb http://ftp.debian.org/debian testing main contrib"
+# apt-get update
```
ab-pm, line 23: You may want to not merge this. As the commit message indicates, I have no idea what I was doing; but apparently installing `ca-certificates` solved the TLS problem with git-cloning the wal2json repository?

Not sure what the debian `testing` repository is good for. If it's still needed, would simply echoing this strings into [some `/etc/apt/sources.list.d/???.list` file](https://manpages.debian.org/trixie/apt/sources.list.5.en.html) work?

packages/lds/src/pg-logical-decoding.ts
```diff
@@ -81,37 +72,67 @@ const toLsnData = ([lsn, data]: [string, string]): Payload => ({
...
+export interface LdsOptions {
+  /** The 'add-tables' wal2json parameter. Defaults to `*.*`. */
   tablePattern?: string;
+  /** The [replication slot](https://www.postgresql.org/docs/current/logicaldecoding-explanation.html#LOGICALDECODING-REPLICATION-SLOTS) identifier. Defaults to `postgraphile`. */
   slotName?: string;
+  /** Whether `.createSlot()` should create a temporary replication slot which will be limited to the `client` session and gets cleaned up automatically. Defaults to `false`. */
   temporary?: boolean;
+  /** (Custom) [type parsers](https://node-postgres.com/features/queries#types) to deserialise the wal2json column string values. Pass `pg.types` to get the default type parsing. Defaults to `undefined`, that is raw values will get emitted. */
+  types?: pg.CustomTypesConfig;
+  /** Extra [parameters to be passed to wal2json](https://github.com/eulerto/wal2json?tab=readme-ov-file#parameters). Use e.g. `{'numeric-data-types-as-string', 't'}` to make the type parsers apply to numeric values. */
```
ab-pm, line 84: A better example might be
```suggestion
  /** Extra [parameters to be passed to wal2json](https://github.com/eulerto/wal2json?tab=readme-ov-file#parameters). Use e.g. `{'filter-tables', 'app_public.irrelevant'}` to ignore a specific table. */
```

